PR #6, opened 2019-12-31
IDEMPIERE-2924: Adding XLSX Export support

dpansheriya:
Before building, make sure that local copy of  maven cache for .m2\repository\p2\osgi\bundle\org.apache.poi.poi-ooxml-schemas\4.1.0 is deleted. I noticed that eclipse set target always load fresh copy but mvn verify try to reuse from local cache.


globalqss:
Tested export to xlsx a file with 70.002 records

PR #10, opened 2020-01-08
IDEMPIERE-4146 void method on Post Depreciation Entry

ocurieles:


globalqss:
moving patches to ticket - closing as no provided ways to test it

PR #13, opened 2020-01-17
IDEMPIERE-4065: Making report configurable to reverse value

dpansheriya:
@globalqss this pull request has corrected Updated SQL and tested with postgresql and oracle


globalqss:

PR #14, opened 2020-01-21
IDEMPIERE-4152 : Payment Allocation : notify user when date changes

nmicoud:


globalqss:
Integrated manually in 7.1

PR #19, opened 2020-04-30
IDEMPIERE-4249 Platform Refresh for 7.1z

hengsin:
Upgrade to Tycho 1.7.0, Eclipse 2020-03, Zk 9.0.1 and Jetty 9.4.28


hieplq:
i guess change to file .classpath org.eclipse.jdt.core.prefs is accidently when you import by eclipse?

if i use fresh idempiere source and use eclipse 2020-03 do import "Existing maven project" then eclipse will set back .classpath and org.eclipse.jdt.core.prefs to prev version
so maybe you use import "Exists projects into workspace"

what's difference between two version of that file?
version of this commit stick to jdk-11, so when i setup new jdk (example jdk-13) then setting still stick to jdk-11.
so when build by eclipse (command don't get any issue, because this one is eclipse special) i get issue with org.adempiere.server and org.adempiere.install
(actually only change on org.eclipse.jdt.core.prefs is make issue change on .classpath is not effect)

hieplq:
use jdk-13
jdkCompilerJdk13
user source level jdk-13
jdkCompilerLevel
compiler error
jdkCompiler

hengsin:
hi @hieplq , I don't remember that now. Anyway, please create a new ticket and pull request for the issue above.

PR #21, opened 2020-04-30
IDEMPIERE-1035 hard to find exceptions when using restrictive rights …

d-ruiz:
This commit improves the error message when a role has no access workflows and the toolbar button is clicked. 

The current message is:
1035-BeforeError

The message after the commit is:
1035-AfterError

To test this you can create a new user with a role that only has access to Sales Order. Log in as that user with that role -> open a Sales Order window -> click on the workflow button.


d-ruiz:
Requested changes added to the pull request

PR #27, opened 2020-05-03
IDEMPIERE-4168 Move Webstore out of core

hengsin:
Change EntityType of webstore table, column, field and menu from 'D' to
'WSTORE'


hengsin:
Push to later.

PR #30, opened 2020-05-04
IDEMPIERE-4157: Refactoring code and Adding translation support

dpansheriya:
Carlos, as per you code review comments, all tool tips are added translation support. Also refactored to move code to maintain state of quick from open session status from  SessionManager class to AbstractADWindowContent.

### DIFF
--- a/migration/i7.1z/oracle/202005041900_IDEMPIERE-4157.sql
+++ b/migration/i7.1z/oracle/202005041900_IDEMPIERE-4157.sql
@@ -1,0 +1,42 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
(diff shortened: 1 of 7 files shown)


hengsin:
I don't understand why the need to keep the list of open quick form tab as session attribute. It is local to the ad window instance and it should be maintain there as instance state and I don't see why there's the need there to have all this static method and session stuff. Please highlight here if I've miss anything.

dpansheriya:
Hengsin, Approach suggested by you may work. The purpose of this code is to avoid reopen of same tab twice and if two tab having same window open, hope it may not cause any issue. Let me test this scenario and then will update here about findings.

PR #34, opened 2020-05-06
IDEMPIERE-4268 : Web Services : Read miss cross-tenant check

nmicoud:
Check that the record is on same tenant as the session's


globalqss:
Closing for deeper review

PR #41, opened 2020-05-09
IDEMPIERE-3333: Respecting ReportView order by clause when there is n…

dpansheriya:
…o orderby field on print format configured.

For quick testing I suggest below steps.

1. Open Unallocated Invoice report. It is already sorted by document no
2. Click customize button and remove sort order
3. Now this report is not sorted by DocumentNo
4. Go to report view and look on record named RV_C_Invoice NotAllocated.
5.  Add DocumentNo in Sql ORDER By field
6. Cache reset and refresh Unallocated invoice report.  This time it should be sorted by document number.


hengsin:
please use the model class directly instead of using mtable.get(...).getpo(...)

globalqss:
Indeed, I think as this piece of code is going to be visited very frequently it would be better to implement the class MReportView with cache in MReportView.get

PR #44, opened 2020-05-09
IDEMPIERE-1146 + IDEMPIERE-3234  / Centralized ID not working with na…

globalqss:
Hi guys, I did some tests on those two tickets and noticed the code was distributed in a bad way (part on DB and part on MSequence), and also it was not good the way it was written.

So, I refactored the method MSequence.getNextID:
- to take into account first the priorities to get the ID from http
- to error in case somebody wants to get an official ID locally
- to get properly native sequence IDs on excluded tables on official
- and get the IDs from AD_Sequence at the end if not http/native

Found a performance issue calling MTable and MTable.getColumns, so I decided to implement a Cache of tables with EntityType column, that made the class to behave fast.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MSequence.java
+++ b/org.adempiere.base/src/org/compiere/model/MSequence.java
@@ -54,10 +54,7 @@ public class MSequence extends X_AD_Sequence
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -2215317912000946608L;
-
-	/** Log Level for Next ID Call					*/
-	private static final Level LOGLEVEL = Level.ALL;
+	private static final long serialVersionUID = -750059366190164777L;
 
 	private static final int QUERY_TIME_OUT = 30;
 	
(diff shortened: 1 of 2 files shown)


hengsin:
hi carlos, you need to move the @deprecated marking from msequence.getnextid to db.getnextid since you have reserved their fate.

globalqss:
@hengsin - I think we still the preferred way must be to call DB.getNextID() just in case of any future wrongdoing.

Maybe we can remove the deprecate from the MSequence.getNextID() - however I think is better to keep a centralized way to call via DB.getNextID - as it was before - and calling directly MSequence works but is not recommended.

hengsin:
hi @globalqss, it would be super confusing to have the actual codes doing the work in a @deprecated method (since @deprecated means you shouldn't continue to use it and it might be removed in future release). so at the minimum, you need to remove the @deprecated tag.

globalqss:
Ready @hengsin - deprecation removed

PR #47, opened 2020-05-09
Idempiere 3333 - Respecting Report view Order by clause in print format

dpansheriya:
I made changes as per peer review comment.


org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -723,7 +725,15 @@ else if (index == -1)
 				finalSQL.append(by);
 			}
 		}	//	order by
-		
+		else if (format.getAD_ReportView_ID() > 0)
+		{
+			X_AD_ReportView reportView = MReportView.get(Env.getCtx(),format.getAD_ReportView_ID());
```
globalqss, line 730: I thinks this must be:

MReportView reportView = MReportView.get(Env.getCtx(),format.getAD_ReportView_ID());

And you don't need to import of X_AD_ReportView

PR #51, opened 2020-05-12
IDEMPIERE-4286 : add MOrg.get method with transaction name

nmicoud:
So it can be used by AFTER_NEW event

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MOrg.java
+++ b/org.adempiere.base/src/org/compiere/model/MOrg.java
@@ -35,7 +35,7 @@ public class MOrg extends X_AD_Org
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -5604686137606338725L;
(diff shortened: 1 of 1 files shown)


hengsin:
we should not cache PO with trx name

nmicoud:
so what should I do in this particular case ?
replace :
MOrg.get(ctx, po.getAD_Org_ID(), trxName)
by a direct sql ?

hengsin:
you can retrieve it using the normal PO method but you shouldn't put that directly into s_cache. The problem is after your trx/trxname ends, you now have a PO with an invalid trxName in the global cache. The next operation that get that PO from s_cache can cause havoc with it.

nmicoud:
Your solution (new MOrg(ctx, po.getAD_Org_ID(), trxName)) is working, thanks for the tip.
Closing the branch

PR #60, opened 2020-05-14
Idempiere 3333

dpansheriya:
Carlos,
I made changes of replacing variable type X_AD_ReportView by MRerportView.


globalqss:
@dpansheriya - didn't notice the pull request was made against release-7.1 instead of master - we have changed the workflow and now everything needs to be integrated to master and periodically we collect the bug fixes and integrate to release-7.1

PR #65, opened 2020-05-18
IDEMPIERE-4294 Cannot install wildcard certificate #resolve IDEMPIERE…

globalqss:
…-4294

Solve java.lang.IllegalStateException: KeyStores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory.


globalqss:
It's already tested locally - and also installed a wildcard certificate in test-postgresql.idempiere.org, test-oracle.idempiere.org,

PR #69, opened 2020-05-19
Idempiere-4298:support update jdk (jdk 12 13 14)

hieplq:
update org.apache.aries.spifly.dynamic.bundle to 1.2.4 to support jdk-12, 13, 14

also upudate asm to latest 7.3.1

### DIFF
--- a/org.adempiere.base.callout/META-INF/MANIFEST.MF
+++ b/org.adempiere.base.callout/META-INF/MANIFEST.MF
@@ -3,7 +3,6 @@ Bundle-ManifestVersion: 2
 Bundle-Name: iDempiere Callout
 Bundle-SymbolicName: org.adempiere.base.callout
 Bundle-Version: 7.1.0.qualifier
-Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version>=11))"
 Require-Bundle: org.adempiere.base;bundle-version="0.0.0"
 Eclipse-RegisterBuddy: org.adempiere.base
--- a/org.adempiere.base.process/META-INF/MANIFEST.MF
+++ b/org.adempiere.base.process/META-INF/MANIFEST.MF
@@ -3,7 +3,6 @@ Bundle-ManifestVersion: 2
 Bundle-Name: iDempiere Processes
 Bundle-SymbolicName: org.adempiere.base.process
 Bundle-Version: 7.1.0.qualifier
-Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version>=11))"
 Require-Bundle: org.adempiere.base;bundle-version="0.0.0"
 Export-Package: org.adempiere.process,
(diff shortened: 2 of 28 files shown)


globalqss:
@hieplq - I don't see the need to remove the setting in pom.xml for:
`
<executionEnvironmentDefault>JavaSE-11</executionEnvironmentDefault>
<executionEnvironment>JavaSE-11</executionEnvironment>
`
Is that necessary?  Because from what I understand we support java 11 - and indeed we recommend java 11 - and we probably won't recommend another version until java 17 (LTS).

Is good that we allow to run iDempiere with java 12/13/14, but I think we don't want to enforce any change beyond java 11, any line of code added must be supported by java 11 - this is, we cannot accept code changes that just exists in java 12/13/14.


On the same line of thinking, the removal of:
`Bundle-RequiredExecutionEnvironment: JavaSE-11`
doesn't seem necessary either, from what I understand, that is an indication of the **minimum** required environment.


But I'm not sure about that - what does @hengsin think about?

Regards,

Carlos Ruiz

hieplq:
hi @globalqss has some point need to clear

1. we don't remove support jdk-11. this improve still run well with jdk-11 and jdk-11 is default
2. we don't remove <executionEnvironmentDefault> and <executionEnvironment> on POM
just remove Bundle-RequiredExecutionEnvironment: JavaSE-11 on manifest.MF
3. Bundle-RequiredExecutionEnvironment is old way to define minimum required environment like you said. before eclipse 2000-03 remove this will make dialog warning when launch server but with eclipse 2000-03 it's fine
because eclipse 2000-03 support new method Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version>=11))"

globalqss:
Hi @hieplq

That explains the Bundle-RequiredExecutionEnvironment - agree about that, thanks for the clarification.

And also from what I see you made jdk.version a variable in pom.xml but still points to 11.

From my side it looks good (haven't conducted tests).  But because of the impact of the change would be good to have also @hengsin approval.

Regards,

Carlos Ruiz

hengsin:
Hi @hieplq ,

Why the need to remove Bundle-RequiredExecutionEnvironment ? With Eclipse 2020-03, if I remove Bundle-RequiredExecutionEnvironment, although it is not a fatal error, it still raise a plugin warning of "No required execution environment has been set". Also, is this really deprecated ? I can't find deprecation info for that either.

Regards,
Low

hieplq:
> Why the need to remove Bundle-RequiredExecutionEnvironment ? 
i run this one
`export JAVA_HOME=path to jdk-13
mvn verify
`
and get bellow error
`Resolution errors:
Bundle org.adempiere.base - Missing Constraint: Bundle-RequiredExecutionEnvironment: JavaSE-11`

seem tycho not respect "minimum required environment" seem need to set correct value or multi value (JavaSE-11 JavaSE-12,..)
when JavaSE-15 release we need to modify again
not sure with multi value can build with jdk-11 or not
so i remove it to remove dependency to that value

seem Bundle-RequiredExecutionEnvironment is eclipse special (but add to osgi standard)
i saw bnd don't use this header
i remember when i do update jdk-10, i saw on spec of osgi 6 or osgi 7 recommend to use Require-Capability over Bundle-RequiredExecutionEnvironment
but still not yet find it again

PR #75, opened 2020-05-19
IDEMPIERE-2523 improve method to add comment to log script

globalqss:
#resolve IDEMPIERE-2523

### DIFF
--- a/migration/i7.1z/oracle/202005192336_IDEMPIERE-2523.sql
+++ b/migration/i7.1z/oracle/202005192336_IDEMPIERE-2523.sql
@@ -1,0 +1,31 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-2523 improve method to add comment to log script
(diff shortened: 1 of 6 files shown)


hieplq:
why need to re-open preference to show that field?

it's a bit small input

image

in case you agree i will modify to visiable input click "log migrate script" and make that field on new line help it bigger now

globalqss:
> why need to re-open preference to show that field?

Because is based on the DisplayLogic and the context is only set after the flag is saved

> it's a bit small input

Not a big deal, I didn't want to make it multi-line to avoid people adding newlines in the comment message and possibly messing up the migration script (didn't test, but I think it would be broken)

> in case you agree i will modify to visiable input click "log migrate script" and 
> make that field on new line help it bigger now

I didn't want to change the Dialog, just did it with the tools in dictionary, if you find an easy way to do it, please go ahead, but I think it doesn't worth a big change - is a feature that are used just by a few people and whoever using this must know what we are doing.

PR #80, opened 2020-05-22
IDEMPIERE-4304: Fixing duplicate saved Filter created if existing fil…

dpansheriya:
…ter is shared.


d-ruiz:
Hi Deepak, 

As a user with AD_User_ID = 0 exists in the system and it is a valid user (System). 

Setting AD_User_ID = 0 for global queries should not be used, System can have its own saved queries for the system windows. That's why it was designed as, global queries are those with AD_USER_ID = NULL, a design that this commit would break.

Therefore, I am closing this pull request, additional comments will be added to the ticket.

Also, I think with the current workflow for pull requests. All PR must be made into master

PR #85, opened 2020-05-28
IDEMPIERE-4313 : WebServices : Yes-No output values not consistent be…

nmicoud:
…tween Read and Query

### DIFF
--- a/org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
+++ b/org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
@@ -1390,8 +1390,13 @@ public class ModelADServiceImpl extends AbstractService implements ModelADServic
 					if (po.get_Value(i) != null){						
 						if(po.get_Value(i) instanceof byte[]){
 							dfid.setVal(new String(Base64.encodeBase64((byte[]) po.get_Value(i))));
-						}else						
-						    dfid.setVal(po.get_Value(i).toString());
+						}
+						else if(po.get_Value(i) instanceof Boolean) {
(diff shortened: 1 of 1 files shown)


org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
```diff
@@ -1390,8 +1390,13 @@ public WindowTabDataDocument readData(ModelCRUDRequestDocument req) {
 					if (po.get_Value(i) != null){						
 						if(po.get_Value(i) instanceof byte[]){
 							dfid.setVal(new String(Base64.encodeBase64((byte[]) po.get_Value(i))));
-						}else						
-						    dfid.setVal(po.get_Value(i).toString());
+						}
+						else if(po.get_Value(i) instanceof Boolean) {
+							dfid.setVal(po.get_Value(i).toString().equals("true") ? "Y" : "N");
```
globalqss, line 1395: Hi @nmicoud - I think link 1395 would be better expressed as
dfid.setVal((Boolean)po.get_Value(i) ? "Y" : "N");

On the other hand - reading the method - is it possible that we also have mismatches between readData and queryData in the formatting of numeric and dates/times/hours values - did you check that?

nmicoud: Hi Carlos,

I've updated the PR with your suggestions.

I didn't check number or dates as we haven't used them.

But, made a quick test and :
 - "Created" column values were exported with same result : "2020-04-30 07:33:10.661"
 - "Amt" column values were exported with same result : "1234"

nb: this was not heavily tested, so I may have miss something

PR #86, opened 2020-05-28
IDEMPIERE-4310

mschnbeck:
Tested reactivation against warehouse orders, POS-orders and on credit orders without entry in SysConfig and with REACTIVATE_WH_ORDER_REVERSES_INVOICE set to Y and N


org.adempiere.base/src/org/compiere/model/MOrder.java
```diff
@@ -75,6 +75,8 @@
 	 */
 	private static final long serialVersionUID = -7784588474522162502L;
 
+	private static final boolean isReactWHOrderReverseInvoice = MSysConfig.getBooleanValue(MSysConfig.REACTIVATE_WH_ORDER_REVERSES_INVOICE, false, Env.getAD_Client_ID(Env.getCtx()));
```
globalqss, line 78: @mschnbeck - just for showing best practice - but please don't fix the pull request as we're going to discuss it better in the ticket:
- MSysConfig.get as a static is not a good idea as it can be different per every client, or it can be changed and cache reset.

mschnbeck: Ok. I got the idea from another usage of MSysConfig but maybe of course that was installation specific.

PR #90, opened 2020-05-30
IDEMPIERE-4314:eclipse project setting independent JDK

hieplq:
ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4314
detail explain: https://github.com/idempiere/idempiere/pull/19#issuecomment-636077414


hengsin:
Perhaps we should consider removing all this files from the repository in future. It is a common practise for maven base projects.

hieplq:
> Perhaps we should consider removing all this files from the repository in future. It is a common practise for maven base projects.

agree, eclipse is change behavior often, i just see what change on your commit is by eclipse
when i setup release-7.1, ever i user import exists maven, eclipse change exact like your commit

PR #95, opened 2020-06-02
IDEMPIERE-4319 Cascade workflow activities (completing chained docume…

globalqss:
…nts) exception problem


hengsin:
don't see any changes here, just //throw new AdempiereException(e); ?

globalqss:
> don't see any changes here, just //throw new AdempiereException(e); ?

Fixed - commenting more in the ticket

PR #97, opened 2020-06-02
IDEMPIERE-3551: allow use environment variable to setting system conf…

hieplq:
this change fixed what @hengsin comment. it also improve to add ability override configuration on database by environment variable for every configuration at system level

> commit a1f1b609af8c42830110b5cf8f294447f0dec705 move "MSysConfig.getValue(MSysConfig.AUTOMATIC_PACKIN_FOLDERS);" outside of the if block. This effectively break the checking to avoid db access before db connection is up.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MSysConfig.java
+++ b/org.adempiere.base/src/org/compiere/model/MSysConfig.java
@@ -234,7 +234,15 @@ public class MSysConfig extends X_AD_SysConfig
 	 */
 	public static String getValue(String Name, String defaultValue)
 	{
-		return getValue(Name, defaultValue, 0, 0);
+        // IDEMPIERE-3551: allow setting system configuration (system level only, not client level) by environment variable over database value
+        // note: change value on database and reset cache don't effect configuration setting by this way
+        String envValue = System.getProperty(Name);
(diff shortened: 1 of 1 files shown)


globalqss:
@hieplq - the commit is done over a JIRA ticket that is closed, and it makes a wider change.  Please create a new ticket to discuss the impact of this.

It sounds like an interesting idea - that would mean the system variable will take precedence over the SysConfig, would like to know opinion from @hengsin about this too.

hengsin:
Not good to change the behaviour of existing public API that's working as intented (different story if it is bug fixing). If needed, better add new api for it, for e.g getValue(key, default, boolean checkSystemProperty) or the more wordy version of getSystemPropertyOrConfigValue(key, defaultValue)

globalqss:
But anyways, I still don't get completely the value added of this as the solution to any problem there is simply to change the SysConfig record.

hieplq:
it's for conventional.

i don't need to start idempiere then go configuration to change value and restart (for case AUTOMATIC_PACKIN_FOLDERS) or reset cache

your method is less step than when you use sql to modify value before run idempiere but still not conventional

PR #98, opened 2020-06-02
IDEMPIERE-4237:timeout page continue do comet request and log a lot m…

hieplq:
it's workaround bu working (hot fix) @ocurieles can you test


ocurieles:
Thanks @hieplq - I will test right now..

ocurieles:
@hieplq tested and working fine :) 

 Thanks

hengsin:
hi @hieplq ,
Wouldn't this break the handling of request timeout ?
Perhaps we should make changes at the server side so that serverpush.js can identified it is session instead of request timeout.
Regards,
Low

hieplq:
we can implement logic at ZkAtmosphereHandler.onRequest, recognizes session timeout and set some special flag to response and use that value for script validate

but for work-around, i want make it simple as possible

about your concern it's confuse with request timeout (or other timeout). we can add more condition to check

if (me._req.status = 0 && me._req.state()=="rejected" && me._req.status = 0 && me._req.statusText = "timeout"){
     ;//ignore by session timeout 
}else{
    me._schedule();
}

i guess other case ever it same statusText then status code will difference, more special like 408 for request timeout

hengsin:
Hi @hieplq ,
understood that you want to keep it simple but at the same time, it is important to avoid to build a workaround that will create another issue.

hengsin:
working on a solution as I suggested above.

PR #99, opened 2020-06-03
Idempiere 4237

hieplq:
this commit make check session timeout condition more safer


globalqss:
solved in pull request #100

PR #100, opened 2020-06-03
IDEMPIERE-4237 timeout page continue do comet request and log a lot m…

hengsin:
…essage

Revert SC_NO_Content for error bad to SC_BAD_REQUEST so that the failure
count at client side will work.
Add session timeout/notfound checking at client side.


hieplq:
go to "response.setStatus(HttpServletResponse.SC_BAD_REQUEST, error);" has some case
+ "Could not find desktop id"
+ "Could not find desktop"
+ "Could not find session"

so i think need to filter for only case "Could not find session"

hengsin:
Hi @hieplq , this are all bad request and need to add to the client side's failure count. The previous SC_NO_CONTENT is a mistake as that will not trigger the failure count logic at client side and can cause infinite bad request from client side.

PR #102, opened 2020-06-04
Fix NPE in Allocation Acct

ocurieles:
https://idempiere.atlassian.net/browse/IDEMPIERE-4324


globalqss:
Please check
https://wiki.idempiere.org/en/Fork_and_Branch_Git_Workflow#Create_feature_branch

PR #103, opened 2020-06-04
IDEMPIERE-4324

ocurieles:
Allocation Acct NPE When the payment is ZERO


globalqss:
@ocurieles - the patch gets short to consider all the potential NPE in fl (two lines above) and in payment, can you please test the pull request 104 instead
https://github.com/idempiere/idempiere/pull/104

PR #112, opened 2020-06-11
IDEMPIERE-4321 (2pack) after packin role, it reload and call to warni…

globalqss:
…ng on IDEMPIERE-2767

### DIFF
--- a/org.adempiere.pipo/src/org/adempiere/pipo2/PackInHandler.java
+++ b/org.adempiere.pipo/src/org/adempiere/pipo2/PackInHandler.java
@@ -526,6 +526,7 @@ public class PackInHandler extends DefaultHandler {
 			.setOrderBy("AD_Client_ID, Name")
 			.list();
     	for (MRole role : roles) {
+    		role.setAD_User_ID(Env.getAD_User_ID(m_ctx.ctx));
         	role.updateAccessRecords(false);
     	}
 	}


hieplq:
seem AD_User_ID just for load access record and load Substituted
i saw we need to re-design code relate AD_User_ID
it's wrong when MRole is share object (it on cache) but it has properties AD_User_ID set per each user login
at moment it don't make error because it use on synchronized function only

globalqss:
> seem AD_User_ID just for load access record and load Substituted
> i saw we need to re-design code relate AD_User_ID
> it's wrong when MRole is share object (it on cache) but it has properties AD_User_ID set per each user login
> at moment it don't make error because it use on synchronized function only

@hieplq - the cache key for MRole is:
String key = AD_Role_ID + "_" + AD_User_ID;
So, different users are not sharing the same object.

Regards,

Carlos Ruiz

hengsin:
it works with the roleid+userid key but that's one of the significant performance bottleneck in the system.

globalqss:
> it works with the roleid+userid key but that's one of the significant performance bottleneck in the system.

Sorry, I don't understand the statement - the cache is to avoid a high cost re-reading of role children (inherited, substitute, recursive, etc) - how is that a performance bottleneck?

BTW - this commit is solving the issue reported in IDEMPIERE-4321 about an unwanted severe log, not anymore.  If the idea is to refactor MRole I think is better to open a different ticket.

hengsin:
the key is roleid+userid so the role access cache is not share by user that's using the same role. this is a significant draw back and it is especially bad when you add in included roles and asp as well.

globalqss:
> the key is roleid+userid so the role access cache is not share by user that's using the same role. this is a significant draw back and it is especially bad when you add in included roles and asp as well.

But that makes sense, because the role behaves different depending on the user logged in (substitute)

PR #125, opened 2020-06-18
Idempiere 4339

nmicoud:
Hi Carlos,
Please find 2 commits :
 #1 : automatically insert records on the Parameters table
 #2 : some improvements to the WS security window layout

=> I'm using them for some weeks and I think it should be useful for everyone.

Regards,

Nicolas


migration/i7.1/oracle/202006181620_IDEMPIERE-4339.sql
```diff
@@ -0,0 +1,98 @@
+SET SQLBLANKLINES ON
```
globalqss, line 1: Hi @nmicoud - can you please move the scripts to i7.1z?

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
```diff
@@ -286,5 +286,45 @@ public X_WS_WebServiceFieldInput getFieldInput(String colName){
 		
 		return m_inputFieldMap.get(colName);
 	}
+
+	public static void insertParameters (Properties ctx, MWebServiceType ws, String trxName) {
+
+		if (ws.getWS_WebService().getValue().equals("ModelADService")) {
```
globalqss, line 292: java best practice:
instead of variable.equals(constant) is better to write
constant.equals(variable)

this is about safety in case the variable is null - the first version will throw an NPE the second will just return false

that applies to this line and also 295, 298, 305..309

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
```diff
@@ -286,5 +286,45 @@ public X_WS_WebServiceFieldInput getFieldInput(String colName){
...
+			}
+			else if (method.equals("runProcess")) {
+				addWsParameter(ctx, ws, "AD_Process_ID", X_WS_WebService_Para.PARAMETERTYPE_Constant, "", trxName); // can't fill it as the process is unknown
+				addWsParameter(ctx, ws, "AD_Menu_ID", X_WS_WebService_Para.PARAMETERTYPE_Constant, "0", trxName);
+				addWsParameter(ctx, ws, "AD_Record_ID", X_WS_WebService_Para.PARAMETERTYPE_Free, "", trxName);
+			}
+			else {
+				String value = "";
+				if (method.equals("createData")) value = "Create";
+				if (method.equals("deleteData")) value = "Delete";
```
globalqss, line 306: here and the next three lines is better to use an else if - that avoids checking for a condition that you already know is false

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
```diff
@@ -286,5 +286,45 @@ public X_WS_WebServiceFieldInput getFieldInput(String colName){
...
+				addWsParameter(ctx, ws, "Action", X_WS_WebService_Para.PARAMETERTYPE_Constant, value, trxName);
+				addWsParameter(ctx, ws, "RecordID", X_WS_WebService_Para.PARAMETERTYPE_Free, "", trxName);
+			}
+		}
+	}
 	
+	private static void addWsParameter(Properties ctx, MWebServiceType ws, String name, String type, String value, String trxName) {
+		
+		if (DB.getSQLValueEx(trxName, "SELECT 1 FROM WS_WebService_Para WHERE WS_WebServiceType_ID = ? AND ParameterName = ?", ws.getWS_WebServiceType_ID(), name) != 1) {
+			X_WS_WebService_Para wsp = new X_WS_WebService_Para(ctx, 0, trxName);
```
globalqss, line 321: There is a MWebServicePara - is always preferred to use the M class instead of the X_ class, using the X_ class skips the validations added in beforeSave.

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
```diff
@@ -286,5 +286,45 @@ public X_WS_WebServiceFieldInput getFieldInput(String colName){
 		
 		return m_inputFieldMap.get(colName);
 	}
+
+	public static void insertParameters (Properties ctx, MWebServiceType ws, String trxName) {
```
globalqss, line 290: it's usually preferable to make this methods not static, so, you can simply call
ws.insertParameters() - no need to pass the ctx and trx as they are part of the object.

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
nmicoud: ok, updated

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
nmicoud: done

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
nmicoud: Correct, here's the reason : I've developped the code on a plugin based on org.adempiere.base (and MWebServicePara is not reachable from there)

org.idempiere.webservices/WEB-INF/src/org/compiere/model/MWebServiceType.java
nmicoud: Thanks for the tip

migration/i7.1z/postgresql/202006181620_IDEMPIERE-4339.sql
```diff
@@ -0,0 +1,95 @@
...
+-- Jun 18, 2020, 4:15:52 PM CEST
+UPDATE AD_Column SET AD_Val_Rule_ID=200139, IsToolbarButton='N',Updated=TO_TIMESTAMP('2020-06-18 16:15:52','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=0 WHERE AD_Column_ID=56757
+;
+
+-- Jun 18, 2020, 4:15:52 PM CEST
+UPDATE AD_Tab SET IsSingleRow='N',Updated=TO_TIMESTAMP('2020-06-18 16:14:36','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=0 WHERE AD_Tab_ID=53191
+;
+
+-- Jun 24, 2020, 5:06:03 PM CEST
+INSERT INTO AD_Process (AD_Process_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,Name,Description,IsReport,Value,IsDirectPrint,Classname,AccessLevel,EntityType,Statistic_Count,Statistic_Seconds,IsBetaFunctionality,IsServerProcess,ShowHelp,CopyFromProcess,AD_Process_UU,AllowMultipleExecution,XXA_IsDossierAction) VALUES (200119,0,0,'Y',TO_TIMESTAMP('2020-06-24 17:06:02','YYYY-MM-DD HH24:MI:SS'),0,TO_TIMESTAMP('2020-06-24 17:06:02','YYYY-MM-DD HH24:MI:SS'),0,'Insert Parameters','This process will add required parameters for current web service type','N','WebServiceTypeCreateParameters','N','org.idempiere.webservices.process.WebServiceTypeCreateParameters','6','D',0,0,'N','N','S','N','2b94e537-3bb4-4e25-8277-13c16061e3a6','P','N')
```
globalqss, line 71: @nmicoud - I tried to apply this migration script and it failed because of the column XXA_IsDossierAction (same in oracle).

PR #128, opened 2020-06-19
IDEMPIERE-1026 Button on table to create automatically the window

d-ruiz:


hieplq:
1. maybe missing p_AD_Window_ID mean p_isNewWindow = true
1. window.name can't duplicate, so should add window name parameter and replace window.setName(table.getName());
1. window.setEntityType(entityType); i prefer set it by parameter when i do new window for core table
1 in case new window then p_TabLevel should = 0 other case should p_TabLevel > 0 tab.setTabLevel(p_TabLevel);

d-ruiz:
>     1. maybe missing p_AD_Window_ID mean p_isNewWindow = true
> 
>     2. window.name can't duplicate, so should add window name parameter and replace window.setName(table.getName());
> 
>     3. window.setEntityType(entityType); i prefer set it by parameter when i do new window for core table
>        1 in case new window then p_TabLevel should = 0 other case should p_TabLevel > 0 tab.setTabLevel(p_TabLevel);

1. Is already on the code, p_AD_Window_ID is only mandatory (and shown in the parameters) if p_isNewWindow = true.
2. I thought about this case, as well. I designed it this way because this way could also help to avoid re-creating windows by mistake. If you want to copy or customize an existing window you can always do it manually with the copy tabs process or Window Customization.
3. The process creates the new records with the EntityType of the Window, this would be to keep consistency. The second part is exactly how it works (if New Window TabLevel = 0, if not new - defined by the user), there can be a window with no Tabs and the user could choose a TabLevel 0 on an existing window.

globalqss:
Hi @d-ruiz - some additional tests for this:

1 - In my first tests related above I didn't set the "Parent Key Column" flag - on a second test I set Parent Key Column and it works OK
So, maybe is good to enforce the parent link as suggested above - or verify there is a parent valid link, this is, a parent link that exists in the tab above (tab above according to tab level, not seqno)

2 - in other tests also I made a "mistake" (that is very common) - some columns were created with one entity type and others with another - then the process just created fields for some of the columns.
I would suggest the process pass an empty EntityType as parameter, and also to change the default for the process when running from the button.

These are the migration scripts that you can join at the end of yours:
[migscr.zip](https://github.com/idempiere/idempiere/files/4813182/migscr.zip)

globalqss:
About my suggestion:

>  2 - The column CUST_Master_ID was created at the bottom of the window, before IsActive.
> The standard is usually to create the master link in the third position, below the AD_Client_ID field

I think the following patch can make the trick:
[TCF.patch.zip](https://github.com/idempiere/idempiere/files/4813213/TCF.patch.zip)

d-ruiz:
Hi Carlos, new commit with your suggestions was added. I've added just the validation for the parent link, not checking if it is a valid link with the master table, that might be worth it to add directly to the Create Fields process and not just in the new one. 

For the tab level, the process checks that the selected tab level is at most one level higher of the highest tab level record in the window. If it's not, for example WIndow with a tabl level of 2, and the user sets a tabl level of 5, a message is shown stating the maximum tab level value.

PR #130, opened 2020-06-23
IDEMPIERE-4342:[2pack]trim space on tables name of sql on packout detail

hieplq:


d-ruiz:
Tested with multiple details and it is working: 
F.e: SELECT * FROM C_Order WHERE C_Order.C_Order_ID > 100000; C_Orderline ; C_OrderTax

PR #132, opened 2020-06-24
IDEMPIERE-4343 : records created by InitialClientSetup must be on Org *

nmicoud:


nmicoud:
Hi @globalqss ; updated !

PR #139, opened 2020-06-25
IDEMPIERE-3508:support more case (2pack: support export from normal s…

hieplq:
cover 2 more case at last comment of https://idempiere.atlassian.net/browse/IDEMPIERE-3508
1. on 2pack is blob and pack in to text column
2. value of column has pattern like binary (fileCount + ".dat" + "|")

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/WFactReconcile.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/WFactReconcile.java
@@ -186,7 +186,9 @@ implements IFormController, EventListener<Event>, WTableModelListener, ValueChan
 		
 		// Parameter Panel
 		North north = new North();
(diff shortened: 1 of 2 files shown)


org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
```diff
@@ -49,14 +50,55 @@ protected String getStringValue (String columnName){
...
+			return TagValueType.BINARY;
 		
-		String strParts [] = value.split("[|]");
-		return strParts.length == 2;
+		Element pe = element.properties.get(qName);
+		if (pe == null) // in case binary, clob this data always not null
+			return TagValueType.TEXT;
+		
+		String value = pe.contents.toString();
+		if (StringUtils.isBlank(value)) // in case binary, clob this data always not blank
```
globalqss, line 89: is StringUtils.isBlank the same as org.compiere.util.Util.isEmpty - I think there is no need to add dependency to an apache commons library here

PR #140, opened 2020-06-25
IDEMPIERE-1674:In physical inventory line. organization isn't readonly

hieplq:
@globalqss pull request for IDEMPIERE-1674:In physical inventory line. organization isn't readonly

### DIFF
--- a/migration-historic/i3.1z/oracle/201403131604_IDEMPIERE-1674.sql
+++ b/migration-historic/i3.1z/oracle/201403131604_IDEMPIERE-1674.sql
@@ -1,0 +1,22 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Mar 13, 2014 1:44:41 AM ICT
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+UPDATE AD_Column SET ReadOnlyLogic='@C_Charge_ID@<1', IsUpdateable='Y',Updated=TO_DATE('2014-03-13 01:44:41','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=0 WHERE AD_Column_ID=3557
+;
+UPDATE AD_Column SET Callout='org.compiere.model.CalloutInventory.setCharge',Updated=TO_DATE('2014-03-13 02:26:56','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=9950
+;
+
+-- Mar 13, 2014 6:41:26 AM ICT
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+UPDATE AD_Field SET ReadOnlyLogic='@InventoryType@!C',Updated=TO_DATE('2014-03-13 06:41:26','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Field_ID=2691
+;
+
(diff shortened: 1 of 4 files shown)


org.adempiere.base/src/org/compiere/model/MInventoryLine.java
```diff
@@ -289,6 +289,10 @@ protected boolean beforeSave (boolean newRecord)
 		MDocType dt = MDocType.get(getCtx(), getParent().getC_DocType_ID());
 		String docSubTypeInv = dt.getDocSubTypeInv();
 
+		//	Set AD_Org to parent if not charge
```
globalqss, line 292: Moving these lines above seems useless and potentially dangerous - in line 330 the charge is being set to zero, so in case that line is hit this code movement will miss the change on ad_org

org.adempiere.base.callout/src/org/compiere/model/CalloutInventory.java
```diff
@@ -173,5 +173,50 @@ private BigDecimal setQtyBook (int M_AttributeSetInstance_ID, int M_Product_ID,
...
+	 *  @param ctx      Context
+	 *  @param WindowNo current Window No
+	 *  @param mTab     Model Tab
+	 *  @param mField   Model Field
+	 *  @param value    The new value
+	 *  @return Error message or ""
+	 */
+	public String setInventoryType (Properties ctx, int WindowNo, GridTab mTab, GridField mField, Object value){
+		String inventoryType = Env.getContext(ctx, WindowNo, "InventoryType").toString();
+		if ("D".equals(inventoryType)){
```
globalqss, line 215: it's better to use here the constant MInventoryLine.INVENTORYTYPE_InventoryDifference instead of the hardcoded value "D" - it makes the codes more readable and easy to maintain

org.adempiere.base.callout/src/org/compiere/model/CalloutInventory.java
```diff
@@ -173,5 +173,50 @@ private BigDecimal setQtyBook (int M_AttributeSetInstance_ID, int M_Product_ID,
...
+	 * 	reset value of org when not set charge 
+	 * 
+	 *  @param ctx      Context
+	 *  @param WindowNo current Window No
+	 *  @param mTab     Model Tab
+	 *  @param mField   Model Field
+	 *  @param value    The new value
+	 *  @return Error message or ""
+	 */
+	public String setCharge (Properties ctx, int WindowNo, GridTab mTab, GridField mField, Object value){
```
globalqss, line 187: the names of the methods here are misleading
- setCharge is not setting a charge - maybe a correct name would be setOrgFromParent
- samely setInventoryType - as both have the same purpose setting the Org from parent maybe they can be the same method and discover from which field is called

migration-historic/i3.1z/oracle/201403131604_IDEMPIERE-1674.sql
```diff
@@ -0,0 +1,22 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Mar 13, 2014 1:44:41 AM ICT
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
```
globalqss, line 5: must be moved to i7.1z for integrating into current version

globalqss:
Closing pull request after 82 days of inactivity - please feel free to create a new one with the comments solved.

PR #141, opened 2020-06-25
IDEMPIERE-4279 Error on packout from the toolbar on certain windows (…

globalqss:
…Table & Column, Window, Tab & Field)

I changed the classes
CommonTranslationHandler
ReferenceElementHandler
ReferenceTableElementHandler
TableElementHandler
WorkflowElementHandler

to avoid using direct JDBC and use the DB and Query methods, it seems the issue is gone, so, this reinforces the best practice about not using JDBC.


d-ruiz:
Tested the cases reported in the ticket and everything is working fine now. Even when the "Export current row only?" flag is unchecked

PR #143, opened 2020-06-26
IDEMPIERE-1518 2Pack shoud not try to reduce column size

globalqss:


hengsin:
what's the option here if there's really the need to reduce column size ?

globalqss:
> what's the option here if there's really the need to reduce column size ?

The option would be to do the "alter table" manually via SQL.

I think that would be the less common case as typically the changes on varchar columns is to increase the size.

PR #144, opened 2020-06-26
IDEMPIERE-3062 Unable to complete MR with qty >1 with Attribute Set o…

globalqss:
…f Always Mandatory


hengsin:
shouldn't there be a check for m_attributesetinstance_id > 0 for minoutlinema as well ?

globalqss:
> shouldn't there be a check for m_attributesetinstance_id > 0 for minoutlinema as well ?

is not necessary because m_inoutlinema.m_attributesetinstance_id cannot be saved as null or zero.

hengsin:
I know null is not possble but zero is a valid ASI value in many place.

PR #147, opened 2020-06-27
IDEMPIERE-3226:env: add method to set Timestamp to tab level

hieplq:


org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -373,6 +373,40 @@ public static void setContext (Properties ctx, int WindowNo, String context, Tim
...
+	 *  @param ctx context
+	 *  @param WindowNo window no
+	 *  @param context context key
+	 *  @param value context value
+	 */
+	public static void setContext (Properties ctx, int WindowNo, int tabNo, String context, Timestamp value)
+	{
+		if (ctx == null || context == null)
+			return;
+		Object oldValue = ctx.getProperty(WindowNo+"|"+tabNo+"|"+context);
```
globalqss, line 387: Warning: The value of the local variable oldValue is not used

org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -373,6 +373,40 @@ public static void setContext (Properties ctx, int WindowNo, String context, Tim
...
+		if (value == null)
+		{
+			ctx.remove(WindowNo+"|"+tabNo+"|"+context);
+			if (log.isLoggable(Level.FINER)) log.finer("Context("+WindowNo+") " + context + "==" + value);
+		}
+		else
+		{	//	JDBC Format	2005-05-09 00:00:00.0
```
globalqss, line 394: I see these comments are just copied from the other methods setContext(...Timestamp, but maybe is a good time to clean up a bit this code in the three methods?

- The comments about JDBC format and BUG are useless now
- The inialization of stringValue to "" is useless, it can be initialized directly below with the format
- The comments about chopping off nanos is useless too

So, basically all those lines are just these three:

			Calendar c1 = Calendar.getInstance();
			c1.setTime(value);
			String stringValue = DisplayType.getTimestampFormat_Default().format(c1.getTime());

PR #150, opened 2020-06-30
IDEMPIERE-1848 can delete org, account record in "Account Schema Dime…

globalqss:
…nsions"

Set Accounting Schema > Account Schema Dimensions > Element Type readonly when the type is Account or Org


hengsin:
Hi @globalqss, I think that shouldn't apply to new C_AcctSchema_Element record.

globalqss:
> Hi @globalqss, I think that shouldn't apply to new C_AcctSchema_Element record.

Thanks @hengsin, solved in a different way - please check my comments on ticket:
https://idempiere.atlassian.net/browse/IDEMPIERE-1848?focusedCommentId=45386&oldIssueView=true&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45386

PR #151, opened 2020-07-01
IDEMPIERE-4354 : Web Services : setDocAction error message not transl…

nmicoud:
…ated

Translation of the error message

### DIFF
--- a/org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
+++ b/org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
@@ -264,7 +264,7 @@ public class ModelADServiceImpl extends AbstractService implements ModelADServic
 			try {
 				if (!((org.compiere.process.DocAction) po).processIt(docAction))
 					return rollbackAndSetError(trx, resp, ret, true,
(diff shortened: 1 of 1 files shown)


org.idempiere.webservices/WEB-INF/src/org/idempiere/adinterface/ModelADServiceImpl.java
```diff
@@ -264,7 +264,7 @@ public StandardResponseDocument setDocAction(ModelSetDocActionRequestDocument re
 			try {
 				if (!((org.compiere.process.DocAction) po).processIt(docAction))
 					return rollbackAndSetError(trx, resp, ret, true,
-							"Couldn't set docAction: " + ((org.compiere.process.DocAction) po).getProcessMsg());
+							"Couldn't set docAction: " + Msg.parseTranslation(ctx, ((org.compiere.process.DocAction) po).getProcessMsg()));
```
globalqss, line 267: Thanks @nmicoud - I understand the goal of this - but I think still it creates a multi-language output for the user - is it better if we translate also the "Couldn't set docAction" part.

Maybe you can use the message FailedProcessingDocument

Something like
Msg.parseTranslation(ctx, "@FailedProcessingDocument@: " + ((org.compiere.process.DocAction) po).getProcessMsg() ....
--> !! Not tested !!

PR #156, opened 2020-07-02
IDEMPIERE-3508:2pack: support export from normal string column to tex…

hieplq:
hi @globalqss i make new pull request with reduce dependency to apache lib

### DIFF
--- a/org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
+++ b/org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
@@ -49,14 +49,55 @@ public class PoFiller{
 		return value;
 	}
 	
-	protected boolean isBlobOnPackinFile (String columnName){
-		String value = getStringValue(columnName);
-		if(value == null)
-			return false;
+	private enum TagValueType {
+		BINARY,
+		CLOB,
+		TEXT
+	}
+	
+	/**
+	 * test to see can treatment value as normal String to set to String column
+	 * 
+	 * value can be string or path to binary file
+	 * in case binary has 2 type
+	 * 1. binary is (attach file, image, binary content) then 
+	 * 		tag name = "BinaryData"
+	 * 		value by format	file path (fileCount + ".dat")
+	 * 			part 1 file name: fileCount + ".dat" // refer: PackOut.writeBlob 
+	 *      	separate char : "|"
+	 *      	part 2 data type: "byte[]" // refer:PoExporter.addBlob
+	 * 2. binary is long string (Clob) then Packout then
+	 * 		tag name is columnName
+	 * 		value by format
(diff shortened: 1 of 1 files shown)


org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
```diff
@@ -338,13 +379,13 @@ public int setTableReference(String qName) {
 					setInteger(qName);
 				} else if (info.getColumnClass(index) == Timestamp.class) {
 					setTimestamp(qName);
-				}else if(DisplayType.TextLong == info.getColumnDisplayType(index)) {// export column from system have type is normal string, but import to system have this column but type is textlong (mean blob)
-					if (getStringValue (qName) != null && !isBlobOnPackinFile(qName)) {
+				}else if(DisplayType.isText(info.getColumnDisplayType(index))) {// data is text then test to see on 2pack can get text to setter or not
+					if (getTreamentType(qName) == TagValueType.TEXT)
```
globalqss, line 383: Hi @hieplq - one doubt here, I see the getTreatmentType has a possible return of BINARY when the column is named BinaryData, but the if here ignores that case, 

is that on purpose?  wouldn't be better then to use a simple else for BINARY and CLOB cases?

hieplq: in case it return BINARY mean we try to import a binary file (image, attach file) on 2pack into string column
it's mismatch just ignore this case

maybe a warning log?

globalqss: Is that check necessary?

I haven't tested, but I mean, what if somebody creates a column named BinaryData and set it as String?  or as a CLOB?

I remember there is something special about that name in core, because there is the Test.BinaryData column, but not sure how would that work in other tables.  Anyways your change is to validate "if(DisplayType.isText" - so, I think you can treat it as String or CLOB (I think both, Binary and CLOB are saved in zip file)

hieplq: check binary to void packin binary (on source database) on 2pack into a string column (on target database)
check depend to columnName = "BinaryData" is weak, but at moment all core table use that name for binary column
i will improve this point, can use data type (byte[] or String) on Element.contents to detect binary type

logic is:
1. attachment save to zip (because we attach many file per record)
2. normal binay (image) save to .dat file
3. clob save to .date file

data type on Element.contents is "string" for clob, "byte[]" for other

globalqss: OK, as you describe it then maybe a warning is fine - it would be a wrong configuration to have a BinaryData column with String/Text content.

globalqss:
Closing pull request after 62 days of inactivity - please feel free to create a new one or reopen with the comments solved.

PR #157, opened 2020-07-02
IDEMPIERE-3226:env: add method to set Timestamp to tab level

hieplq:
hi @globalqss i refine commit for IDEMPIERE-3226

### DIFF
--- a/org.adempiere.base.callout/src/org/compiere/model/CalloutInOut.java
+++ b/org.adempiere.base.callout/src/org/compiere/model/CalloutInOut.java
@@ -369,7 +369,7 @@ public class CalloutInOut extends CalloutEngine
 				//	Locator
 				ii = Integer.valueOf(rs.getInt(2));
 				if (rs.wasNull())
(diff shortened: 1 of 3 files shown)


globalqss:
@hieplq - I preferred the previous version of the patch, in the new patch you're adding extra String concatenation operation (which is CPU expensive) when most of the time is not required.
I mean, these lines:
    String contextLog = "Context(" + WindowNo + ") ";
    String contextLog = "Context(" + WindowNo + "," + tabNo + ") ";
are executed everytime, but that value is just used when the log is FINER, which is rarely the case.

Also, this change is breaking also the swing client, so I would expect some plugins to be broken too - that can be solved with a migration note in the wiki.

However, I would like a peer review from @hengsin better about.
The previous pull request I'm refering is this one:
https://github.com/idempiere/idempiere/pull/147

Regards,

Carlos Ruiz

globalqss:
Agree - setContextTS sounds safer (less potentially disruptive)

globalqss:
Closing pull request after 78 days of inactivity - please feel free to create a new one or reopen with the comments solved.

PR #163, opened 2020-07-07
Idempiere-4162

mschnbeck:
IDEMPIERE-4162 ignore inactive products when looking for duplicate upc, handle errors on update without disturbing the import run.


hengsin:
For use case where UPC is support to be uniqued, having the same UPC for 2 product even though one of them is inactive is not good.

mschnbeck:
> 
> 
> For use case where UPC is support to be unique, having the same UPC for 2 product even though one of them is inactive is not good.

May be it's not a good idea. But the system allows that. If the error were handled correctly it had been entered into the line so that the user had a chance to correct it. Because this happens on other errors, too, I also changed the error handling so that error messages could be committed in any case.

mschnbeck:
better solution found

PR #164, opened 2020-07-08
IDEMPIERE-4281 : Credit memo linked to an invoice should generate aut…

nmicoud:
…omatically an allocation


org.adempiere.base/src/org/compiere/model/MInvoice.java
```diff
@@ -1943,7 +1943,42 @@ else if (!isSOTrx()
 		if (matchPO > 0)
 			info.append(" @M_MatchPO_ID@#").append(matchPO).append(" ");
 
+		// Generate and complete an allocation if is a credit memo linked to an invoice
```
globalqss, line 1946: There is a problem here when the PaymentRule of the credit note is Cash, because a Payment and an Allocation are created in line 1768.

So, code here or there must take into account that fact.

org.adempiere.base/src/org/compiere/model/MInvoice.java
```diff
@@ -1943,7 +1943,42 @@ else if (!isSOTrx()
...
+			if (!inv.isPaid()) {
+				BigDecimal invOpenAmt = inv.getOpenAmt();
+				BigDecimal cmOpenAmt = getOpenAmt();
+				StringBuilder aDescription = new StringBuilder(Msg.getElement(getCtx(), "C_Invoice_ID")).append(": ").append(inv.getDocumentNo()).append("/").append(getDocumentNo());
+
+				MAllocationHdr ah = new MAllocationHdr(getCtx(), true, TimeUtil.max(getDateAcct(), inv.getDateAcct()), getC_Currency_ID(), aDescription.toString(), get_TrxName());
+				ah.setAD_Org_ID(getAD_Org_ID());
+				ah.saveEx();
+
+				if (invOpenAmt.abs().compareTo(cmOpenAmt.abs()) == 0) {
```
globalqss, line 1959: The usage of .abs() is mostly a bad practice for these kind of comparisons.

Problem here is that an invoice of +10 and a credit memo of +10 or -10 (a negative credit memo) - will compare as equal.

I think the comparison must be exact, not with .abs() - and same goes for the other two comparisons below - which I think it must take into account if is sales or purchase.

org.adempiere.base/src/org/compiere/model/MInvoice.java
nmicoud: I don't know how to test (I don't use the Cash payment rule, but I think changing the first line of the patch 
`if (isCreditMemo() && getRelatedInvoice_ID() > 0) {`
by
`if (isCreditMemo() && getRelatedInvoice_ID() > 0 && !isPaid()) {`
should be enough.

wdyt ?

globalqss: > I don't know how to test (I don't use the Cash payment rule, but I think changing the first line of the patch

on pull requests you always **must** test in a vanilla idempiere



>  should be enough.

no idea if at that point the invoice arrives paid, you need to conduct tests for that - but I think as the condition is specific for cash payment rule maybe you need to check that, and the impact that causes

nmicoud: When I said I don't use Cash payment rule, I meant I have no idea what is behind, what is the logic, how to use it... and therefore what needs to be tested.
I think this was discussed in last marketing commitee, but that could be useful to have a kind of 'community help needed' to help on analysis/test/....
(I agreee, is not the best place to discuss)

globalqss:
Closing pull request after 61 days of inactivity - please feel free to create a new one or reopen with the comments solved.

Tried to integrate into eclipse and is throwing errors:
The method addAllocationLine(MAllocationHdr, int, BigDecimal, BigDecimal) is undefined for the type MInvoice

PR #165, opened 2020-07-12
IDEMPIERE-4365:[2pack] pack out translate for MTable MColumn

hieplq:


hengsin:
For exception log, should use log.log(Level.INFO, e.getMessage(), e) instead of log.info(e.toString()). Stacktrace matter.

PR #166, opened 2020-07-14
Idempiere 2999

Soundbytes:
I think the feature is now working again as originally intended and documented here: 
https://wiki.idempiere.org/en/NF3.1_Attribute_All_Reference
I am afraid I did not find the time to extensively test all available reference options. 
I did test a number of list and table references and also a couple of the other stuff like multiple choice lists, color selection, amount and did not run into any problems so far.

If anybody finds the time to test some of the available options please leave a note here. Please do not only post bug reports. It is also important to know what is working fine to avoid duplicated efforts. Also if you think that an option does not make any sense feel free to post that too.
Thanks! 

Here is what I believe needs to be done next:

- Evaluate the available table and list options then remove those that do not make any sense. (the field validation code can be modified to do that)
- The product attribute dialog window is currently a child of the main window. It would be great if it could be made a child of the product tab instead. This way a referenced attribute, table or list whose view is opened by clicking on the attribute label could be viewed and edited without having to close the attribute dialog first. 
- Currently the Sysadmin has no means to exclude any custom table references from a plugin installation or from a customization from the reference list. I do not think that is really an issue. Still if anybody needs that feature please make your voice heard.
- The Oracle migraton scripts need testing. I am not using oracle and won't be able to do that.

I hope you like that stuff

Andreas


Soundbytes:
The ad_languange table reference has been excluded from the menu because it does not use the ad_language_id as the key column.
Todo: Add a new ad_language reference that can be used as an attribute reference. (or find a better solution to the issue)

globalqss:
Hi @Soundbytes,

According to https://wiki.idempiere.org/en/Contributing_to_Trunk
* Mixing new features code and re-factoring code is hard to review. Please make different commits for both cases.

The peer review is better and easier if your code just focus on the required changes of this ticket.
I mean, for example, adding the "Utility scripts to backup and restore the database within the develoment" is out of the scope of this ticket, and that needs to be discussed if is OK in a different ticket.

I can help with the tests here, but please clean up the commit just focusing on the changes.


Just a visual check on scripts 202007132043_IDEMPIERE-2999.sql
* the constant dates are not good, you better use the method TO_TIMESTAMP (postgresql) and TO_DATE (oracle)

Regards,

Carlos Ruiz

Soundbytes:
> requesting changes as per comment above

@globalqss  Hi Carlos. Please excuse the delay. 

I have now implemented the requested changes. I am also going to clean up the commit history once the pull request is ready to be merged.

Do you think the two migration files should be consolidated into one? For better overview I have for the moment kept the original patch apart from my own changes.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
```diff
@@ -535,140 +534,176 @@ private void addAttributeLine (Rows rows, MAttribute attribute, boolean product,
...
+			row.appendChild(fieldEditor);
+			editor.showMenu();
 			if (readOnly)
-				editor.setEnabled(false);
+				editor.setReadWrite(false);
 			else
-				m_editors.add (editor);
+				m_editors.add(editor);
 		}
+//		else if(MAttribute.ATTRIBUTEVALUETYPE_Date.equals(attribute.getAttributeValueType()))
```
Soundbytes, line 583: This block still needs to be translated and moved to the proper place. The date attribute type was not available when the modification was first introduced.So the original patch file does not deal with that.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
```diff
@@ -535,140 +534,176 @@ private void addAttributeLine (Rows rows, MAttribute attribute, boolean product,
...
+		GridFieldVO vo = GridFieldVO.createParameter(Env.getCtx(), m_WindowNo, AEnv.getADWindowID(m_WindowNo), 0, 0,
+				attribute.getName(), Msg.translate(Env.getCtx(), attribute.get_Translation("Name")),
+				DisplayType.Number, 0, false, false, "");
+		String desc = attribute.get_Translation("Description");
+		vo.Description = desc != null ? desc : "";
+		GridField gridField = new GridField(vo);
+		return gridField;
+	}	
+	
+//	private void setDateAttribute(MAttribute attribute,Datebox editor)
```
Soundbytes, line 632: the same applies here

dpansheriya:
@globalqss can we create clean pull request. This pull request has many commits and missing some commits in logiliteERP. Also can we now drop date type support we added and we can cover same by all reference type. If we do this, we needs to add migration for existing attribute of date type.

globalqss:
I think yes @dpansheriya - @Soundbytes are OK about that?   would you be willing to help with peer review?

CarlosRuiz-globalqss:
closing as per @dpansheriya request

PR #168, opened 2020-07-15
IDEMPIERE-4365:[2pack] pack out translate for MTable MColumn

hieplq:
fixed comment from @hengsin 
update pull request by conflict to master

### DIFF
--- a/org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/ColumnElementHandler.java
+++ b/org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/ColumnElementHandler.java
@@ -87,6 +87,7 @@ public class ColumnElementHandler extends AbstractElementHandler {
 				mColumn.setFieldLength(mColumn.get_ValueOldAsInt(MColumn.COLUMNNAME_FieldLength));
 			}
 
+			element.recordId = mColumn.get_ID();
 			if (!mColumn.is_new() && !mColumn.is_Changed()) {
 				boolean syncDatabase = "Y".equalsIgnoreCase(getStringValue(element, "IsSyncDatabase"));
 				if (syncDatabase) {
@@ -95,7 +96,7 @@ public class ColumnElementHandler extends AbstractElementHandler {
 				}
 				return;
 			}
-			
+
 			X_AD_Package_Imp_Detail impDetail = createImportDetail(ctx, element.qName, X_AD_Column.Table_Name, X_AD_Column.Table_ID);
 			String action = null;
 			if (!mColumn.is_new()) {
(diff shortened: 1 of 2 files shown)


org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/ColumnElementHandler.java
```diff
@@ -348,6 +349,14 @@ public void create(PIPOContext ctx, TransformerHandler document)
 		addTypeName(atts, "table");
 		document.startElement("", "", I_AD_Column.Table_Name, atts);
 		createColumnBinding(ctx, document, m_Column);
+		
+		ctx.ctx.put("Table_Name",X_AD_Column.Table_Name);
+		try {
+			new CommonTranslationHandler().packOut(ctx.packOut, document, null, AD_Column_ID);
+		} catch(Exception e) {
+			if (log.isLoggable(Level.INFO)) log.log(Level.INFO, e.getMessage());
```
hengsin, line 357: Should be log.log(Level.INFO, e.getMessage(), e). As I've mention before, StackTrace matter. Same goes for the log below

hengsin:
Hi @hieplq , I tested this and while it is working as expected, I'm wondering what's the use case for this when translation of column and table is usually overwritten by translation from ad_element. Can you clarify on this.

hieplq:
has some reason for this path
1. for consistent to method export record as zip file
2. for consistent to method packout with type = "Data Single"
3. for consistent with field
4. not much but sometime i reuse core element for new customize column and it has new meaning. i can maintain translate at filed level, but why not column level

globalqss:
I was thinking similar as Heng Sin, IMO Column Translation is something that we could simply remove from the system, from what I know it doesn't do any job, and just causes extra space in translations and in database.

hieplq:
it's fine for me to remove column translate
but in case not yet remove it then this PR still helpful for consistent 2pack

hengsin:
hi @hieplq ,

how do you maintain translation at column level ? there's no field equivalent of iscentralizedmaintain flag and sync terminology doesn't check column_trl.istranslate either.

globalqss:
I would suggest instead of this commit, to deprecate the AD_Table_Trl and  AD_Column_Trl tables, AFAIK they have no use and is just consuming space in every translation, and confusing people if they think it requires translation.

nmicoud:
Hello,
AD_Table_Trl is used on WAcctViewer and I think also on the FactAcctReset process (there, you can select a table).
Having it translated is better. So I don't think is a good idea to deprecate it.

FWIW, I'm using AD_Column_Trl on specific windows, to allow users to select a column to setup some specific stuff. Here again, having items translated is better, but I can change that to a reference.

globalqss:
Ah - you're right @nmicoud - I see AD_Table_ID element is used in 72 tables and 10 process parameters.

About AD_Column_ID is being used in some user tables/windows, like AD_ChangeLog, AD_ImpFormat_Row, AD_PrintFormatItem, AD_WF_NextCondition, AD_WF_Node, C_AcctSchema_Element

So, I think my proposal doesn't make sense - it has some value to have the translation there, or we would need to change those references to use element translation, which seems a lot more difficult.

globalqss:
Closing pull request after 38 days of inactivity - please feel free to create a new one or reopen with the comments solved.

After re-reviewing this I think the table AD_Column_Trl can be simply replaced by a view, as explained by Heng Sin above it's always getting the value from AD_Element_Trl.

PR #171, opened 2020-07-17
Idempiere 4162

mschnbeck:
it appeared that errormessages on subtable of products are included into rollback if subsequent products have an error. Correcting this also led to a smother solution for the missing errormessages on update-error of the product itself.


hengsin:
Hi @mschnbeck ,

I think it is usually not good to have duplicate upc even if the other product is inactive. I would suggest to change the validation instead to validate that the matched product is Active (for e.g I_ErrorMsg='Product is not active').

mschnbeck:
Hi @hengsin,
it's not the problem that the product is inactive. Even then it could be updated by finding it with it's upc. The problem is that there is an active and an inactive product with the same upc. That's leads to an error (too many row IIRC) which then leads to follow up errors.

If you don't want to have active and inactive products with the same upc (I think it's worth thinking about) that's a different case and it would be an incompatible change.

Regards,
Martin

hengsin:
Hi Martin,

You added the condition "and p.IsActive = 'Y'" for the UPC search and that create the possibility to create a new product that with share the same UPC with an existing inactive product. I means instead of doing that, we flag the I_Product record as having error if the UPC value matched to an existing inactive  product. That should help to avoid the possibility above.

mschnbeck:
Hi @hengsin,

that would not hinder users to manually enter a new product with the same upc as an inactive one. That was done by my customer. So as long as we allow to enter such products, we must except inactive products to get this select running.

And even if we go to add an unique constraint to this field to hinder entering such cases, the p.IsActive = 'Y' doesn't hurt. In this case the user will get a clear error message and can decide to reactivate the product or remove the upc. But as I said before, this unique constraint isn't possibly not compatible with current databases.
Regards,
Martin

mschnbeck:
Hi @hengsin,

ok, now I got your idea. We add an additional selection before the one I changed and this selection sets the errormessage, right? So that in the next selection the line isn't evaluated due to being flagged as error, right?
Regards,
Martin

hengsin:
Hi @mschnbeck ,

Yes, that would help. I think it probably worth another ticket to address that the system didn't have a means to enforce uniqueness of product upc.

mschnbeck:
Hi @hengsin,

I'll change it accordingly.

Regards,
Martin

mschnbeck:
Hi @hengsin,

now I have implemented the check for duplicate UPC (connected to import products) in m_product and removed the check for inactive products. AFAICS the new commit is automatically added to the push request, right?
Regards,
Martin

hengsin:
Hi @mschnbeck,
I think we should add the following inactive checking as well (before // Set Product Category):

//check product is inactive
		sql = new StringBuilder ("UPDATE I_Product i ")
				.append("SET I_IsImported='E', I_ErrorMsg=I_ErrorMsg||'ERR=Product is not active,' ")
				.append("WHERE I_IsImported<>'Y' AND M_Product_ID IS NOT NULL")
				.append(" AND EXISTS (SELECT 1 FROM M_Product mp")
				.append(" WHERE i.AD_Client_ID=mp.AD_Client_ID AND i.M_Product_ID= mp.M_Product_ID AND mp.IsActive='N')").append(clientCheck);
		no = DB.executeUpdate(sql.toString(), get_TrxName());
		if (no != 0)
			log.warning("Product inactive");

Regards,
Low

mschnbeck:
Hi @hengsin,

very good idea. Added, tested, works as expected.

PR #172, opened 2020-07-18
IDEMPIERE-4372 Performance: only add log record and sql statement to …

hengsin:
…context if trace/debug level <= debug/info

Minor clean up and thread safe fix for CLogErrorBuffer and MIssue
Only add LogRecord and SQL to context if TraceLevel is <= INFO

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -424,8 +424,10 @@ public class GridTable extends AbstractTableModel
 			m_SQL += " ORDER BY " + m_orderClause;
 		}
 		//
-		log.fine(m_SQL_Count);
-		Env.setContext(m_ctx, m_WindowNo, m_TabNo, GridTab.CTX_SQL, m_SQL);
+		if (log.isLoggable(Level.FINE))
+			log.fine(m_SQL_Count);
+		if (log.isLoggable(Level.INFO))
+			Env.setContext(m_ctx, m_WindowNo, m_TabNo, GridTab.CTX_SQL, m_SQL);
 		return m_SQL;
 	}	//	createSelectSql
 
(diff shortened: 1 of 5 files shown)


org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -424,8 +424,10 @@ private String createSelectSql()
 			m_SQL += " ORDER BY " + m_orderClause;
 		}
 		//
-		log.fine(m_SQL_Count);
-		Env.setContext(m_ctx, m_WindowNo, m_TabNo, GridTab.CTX_SQL, m_SQL);
+		if (log.isLoggable(Level.FINE))
+			log.fine(m_SQL_Count);
+		if (log.isLoggable(Level.INFO))
```
globalqss, line 429: Having this in context is very useful to find sometimes why the expected records are not appearing - but as the workaround is now to set the log level on INFO, I think is OK and we can live with that.

org.adempiere.base/src/org/compiere/model/MIssue.java
```diff
@@ -53,10 +53,10 @@
 	 */
 	public static MIssue create (LogRecord record)
 	{
-		s_log.config(record.getMessage());
+		if (s_log.isLoggable(Level.CONFIG))
+			s_log.config(record.getMessage());
 		MSystem system = MSystem.get(Env.getCtx()); 
-		if (!DB.isConnected()
```
globalqss, line 58: Hi @hengsin - I don't understand why to drop the validation if the DB is not connected, I think it is there because sometimes an Issue is created before the DB plugin is started, or because of database failure, and in such case trying to save an Issue will create a new Issue and probably it will end in a recursive endless loop there.

hengsin: that's actually redundant as DB.isConnected check is already done in the CLogErrorBuffer.publish method. anyway, I've add it back since it is harmless to have it there as extra guard.

org.adempiere.base/src/org/compiere/model/GridTable.java
hengsin: Yes, very  usefull for debug but we shouldn't have that in production. Performance aside, those sql can include sensitive data that user shouldn't see.

PR #173, opened 2020-07-18
IDEMPIERE-4322 Sort tab does not support virtual column identifier an…

hengsin:
…d identifiers with different data type

add support for virtual column identifier
concat multiple identifier instead of taking the first not null value


globalqss:
Hi @hengsin - please check comments in ticket.

PR #181, opened 2020-07-22
IDEMPIERE-4380 Inconsistencies with selected filter query on tab navigation

d-ruiz:
https://idempiere.atlassian.net/browse/IDEMPIERE-4380


d-ruiz:
Hi @hengsin, you're right, I just pushed the corresponding changes to address the issue you mention.

hengsin:
Hi @d-ruiz ,
Notice one thing in my testing - enter query for both header and detail. At header, when I navigate to the next record, the query for the detail tab is clear. However, if I don't make any record navigation and just create a new query at header that returns a different set of records, the detail tab query is keep. Is that the intended behaviour ?

d-ruiz:
@hengsin,

To clear the detail query on navigation is intended because the detail grid mode filters the rows based on that query. I think on navigation those filters should be cleared. 
If the new query on the master tab returns a different set of records but the currently active record does not change, the detail tab query is kept. If the master query returns a different set of records and the currently active record changes and the detail tab query is kept, that would be an error from my side. Is that the case?

hengsin:
Hi @d-ruiz , yeah, you are right, detail query is clear when current record change due to new query or record navigation.

PR #184, opened 2020-07-24
IDEMPIERE-4383 Performance: CalloutInOut for bPartner field has bad p…

hengsin:
…erformance for BP with many locations and contacts

use separate query to get location and contact


globalqss:

PR #188, opened 2020-07-27
IDEMPIERE-4386 setAD_User_ID() breaks when logged in as System user

globalqss:
Allow saving Tab Customization for System user

https://idempiere.atlassian.net/browse/IDEMPIERE-4386

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MUser.java
+++ b/org.adempiere.base/src/org/compiere/model/MUser.java
@@ -1092,12 +1092,7 @@ public class MUser extends X_AD_User
 				MPasswordHistory passwordHistory = new MPasswordHistory(this.getCtx(), 0, this.get_TrxName());
 				passwordHistory.setSalt(this.getSalt());
 				passwordHistory.setPassword(this.getPassword());
-				// http://wiki.idempiere.org/en/System_user
-				if (!this.is_new() && this.getAD_User_ID() == 0){
-					passwordHistory.set_Value(MPasswordHistory.COLUMNNAME_AD_User_ID, 0);
-				}else{
-					passwordHistory.setAD_User_ID(this.getAD_User_ID());
-				}
+				passwordHistory.setAD_User_ID(this.getAD_User_ID());
 				passwordHistory.setDatePasswordChanged(this.getUpdated());
 				passwordHistory.saveEx();
 			}
--- a/org.adempiere.base/src/org/adempiere/model/MTabCustomization.java
+++ b/org.adempiere.base/src/org/adempiere/model/MTabCustomization.java
@@ -74,7 +74,7 @@ public class MTabCustomization extends X_AD_Tab_Customization {
 		{
 			tabCust = new MTabCustomization(ctx, 0, trxName);
 			tabCust.setAD_Tab_ID(AD_Tab_ID);
-			tabCust.setAD_User_ID(AD_User_ID);
+			tabCust.set_Value(MTabCustomization.COLUMNNAME_AD_User_ID,0); // allow saving for System
 			tabCust.setCustom(Custom);
 			tabCust.setIsDisplayedGrid(DisplayedGrid);
 			tabCust.setIsQuickForm(isQuickForm);
--- a/org.adempiere.base/src/org/adempiere/model/MTabCustomization.java
+++ b/org.adempiere.base/src/org/adempiere/model/MTabCustomization.java
@@ -11,7 +11,7 @@ public class MTabCustomization extends X_AD_Tab_Customization {
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 786885789239156678L;
+	private static final long serialVersionUID = -4986336833193761507L;
 
 	public MTabCustomization(Properties ctx, int AD_Tab_Customization_ID, String trxName) {
 		super(ctx, AD_Tab_Customization_ID, trxName);
@@ -74,7 +74,7 @@ public class MTabCustomization extends X_AD_Tab_Customization {
 		{
 			tabCust = new MTabCustomization(ctx, 0, trxName);
 			tabCust.setAD_Tab_ID(AD_Tab_ID);
-			tabCust.set_Value(MTabCustomization.COLUMNNAME_AD_User_ID,AD_User_ID); // allow saving for System
+			tabCust.setAD_User_ID(AD_User_ID);
 			tabCust.setCustom(Custom);
 			tabCust.setIsDisplayedGrid(DisplayedGrid);
 			tabCust.setIsQuickForm(isQuickForm);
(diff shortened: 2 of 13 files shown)


globalqss:
> IMO the proposed change does not improve the application. It does not allow saving for System as stated. Instead it just throws the used ID away that had been handed in as a parameter and replaces it with the System users ID.

Thank @Soundbytes for your peer review - glad that you reviewed it and found my copy/paste error.  My mistake I just did the white test case and didn't check with another user, this time I tested with System and GardenAdmin.

Soundbytes:
As I already seid in the Jira ticket: To me the cleanest short term fix would be to override setAD_User_ID () as well as the getAD_User_ID() methods: 

@Override
public void setAD_User_ID (int AD_User_ID)
{
    if (AD_User_ID < 0)
        set_Value (COLUMNNAME_AD_User_ID, null);
    else
        set_Value (COLUMNNAME_AD_User_ID, Integer.valueOf(AD_User_ID));
}

@Ocerride
public int getAD_User_ID () 
{
	Integer ii = (Integer)get_Value(COLUMNNAME_AD_User_ID);
	if (ii == null)
		 return 0;
	return ii.intValue();
}

To me a sustainable long term solution would involve replacing the blueprint functions used by the model generator as well as all instances of the functions by the fixed versions. A search for 'public void setAD_User_ID' resulted in 182 Hits.

globalqss:
Thanks, answered my opinion in the ticket
https://idempiere.atlassian.net/browse/IDEMPIERE-4386?focusedCommentId=45503&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45503

Soundbytes:
@globalqss Thanks again for your patience. Rereading my comments I realized some of them are a bit harsh. I really appreciate your efforts and I admire the speed that you guys have in providing fixes. 
Thanks also for sharing your thoughts on the long term maintenance of the iDempiere project. I am definitely going to have a close look into the server.product.functionaltest project and see if I can add something there.

globalqss:
Hi @Soundbytes - I changed my commit to implement better your suggestion and expanded to all the other classes having the same approach.

org.adempiere.base/src/org/compiere/model/MUser.java
```diff
@@ -1092,12 +1092,7 @@ protected boolean afterSave(boolean newRecord, boolean success) {
 				MPasswordHistory passwordHistory = new MPasswordHistory(this.getCtx(), 0, this.get_TrxName());
 				passwordHistory.setSalt(this.getSalt());
 				passwordHistory.setPassword(this.getPassword());
-				// http://wiki.idempiere.org/en/System_user
-				if (!this.is_new() && this.getAD_User_ID() == 0){
-					passwordHistory.set_Value(MPasswordHistory.COLUMNNAME_AD_User_ID, 0);
-				}else{
-					passwordHistory.setAD_User_ID(this.getAD_User_ID());
-				}
+				passwordHistory.setAD_User_ID(this.getAD_User_ID());
```
Soundbytes, line 1095: Here is the line that exemplifies what I am not happy with. setAD_User_ID() now accepts zero as a valid value. Still getAD_User_ID() does not distinguish between a null and a zero in the column. It will in both cases return zero. A null that is read from one table (or that could be in place because of an initialization error)  gets written as a zero in the other table. This mechanism could potentially introduce or propagate an error that could not be spotted and might lead to  corrupt database entries. I'd suggest to also override the getAD_User_ID() method to remove the possible ambivalent state.

Soundbytes:
Thanks Carlos for considering my suggestions. Of course I still think the best solution would be to to just retire the System user. However I do understand if this is not an option for you and I am glad that you are now taking a broader approach in tackling the Bug. I still have two issues with the solution: 
A) The ambivalence that is introduced by not using a modified getAD_User_ID() method. I have explained that in the post above. 
B) I am a great enemy to code duplication. I understand that it comes as a heritage of JJ original design around the model generator. While I do not love it I have found some peace in accepting the status quo because the duplicated code is in the I_Model and X_Model interfaces and classes that are never modified by any human. With the overriden accessors IMO this is another story. here it might make sense to put the modification in a single place as a static member of a helper class and to call that function from the override instead of copying the function body all over the place. 
IMO this should greatly improve the maintainability. Changes would only have to be made in one place and what is more important - the overrides could be retrived very quickly because they all reference the helper class. It also improves code readability for new developers which to me is a crucial factor. Do you think this would be worth it or do you think my suggestion would just overcomplicate things? 

Here is an example implementation:
```
class Helper
{
    public static int getAD_User_ID (Po po)
    {
        Integer ii = (Integer)po.get_Value(COLUMNNAME_AD_User_ID);
        if (ii == null)
            return -1;
        return ii.intValue();
    }
}
```

in the model class:

```
    public int getAD_User_ID ()
    {
        return Helper.getAD_User_ID(this);
    }
```

globalqss:
I don't think that's a good idea - changing getAD_User_ID to return -1 when null introduces something that nobody expects at this moment from the API.  I think developers know about the existence of the System record with zero ID, and if they want to check for nulls they can directly use .get_Value("AD_User_ID")

Soundbytes:
OK, should we then leave the ambiguity in place or use .get_Value("AD_User_ID") to work around it?
I just saw an instance where the code might break eventuallly. However if you can confirm that this a non issue I will happily rely on your experience and approve the change. (BTW: Does my vote even count here?)

globalqss:
> I just saw an instance where the code might break eventuallly.

If there is some point in core that needs to be fixed, please report it.

> However if you can confirm that this a non issue I will happily rely on your experience and approve the change. (BTW: Does my vote even count here?)

Of course your peer review counts, indeed we are asking for help on this task:
https://wiki.idempiere.org/en/Pull_Request_Review

It still needs the approval from a core committer, but knowing that a community member tested and reviewed is very valuable, and you create trust around you.

Soundbytes:
> > I just saw an instance where the code might break eventuallly.
> 
> If there is some point in core that needs to be fixed, please report it.

What I meant to say was: I just saw **this as** an instance where the code might break eventuallly.
`passwordHistory.setAD_User_ID(this.getAD_User_ID());`
(see my posting above.)

globalqss:
Although the column ad_password_history.ad_user_id is nullable (probably wrongly defined) - I don't see a use case where the code can arrive there without a user, I would think that column must be made "not null"

hengsin:
Just curious what really is the existence of System user is for ? I've never log in as System user before.

globalqss:
> Just curious what really is the existence of System user is for ? I've never log in as System user before.

I don't use System either, I suppose it was JJ attempt to have a "root" with sudo permissions (SuperUser) and a plain "root" without sudo (System) - but of course, if you have System access, with enough knowledge you can manage to give yourself access to other clients.

Soundbytes:
Maybe it would just solve that problem once and for all if the System user were set inactive.
The standard behavior could then be restored or left in place.  

That would enhance the consistency of the api and to me that is a most valuable asset.

To you core developers it is just a small effort to get along with the quirks of an api that you yourself have been working on  over the years. For a relatively unexperienced api user like me inconsistencies like that just make life harder than it needs to be. Thanks!

hengsin:
hi @globalqss , any further changes for this ?

globalqss:
> hi @globalqss , any further changes for this ?

Hi @hengsin - I don't think there is a need for more changes, the current commits solves the bug on tab customization, and I extended that to clean the implementation in all the other classes to make it standard.

I don't agree about disabling System user in core - and is really easy to disable it in an specific implementation if anybody wants that.

Regards,

Carlos Ruiz

PR #190, opened 2020-07-30
IDEMPIERE-874 Suspense Account is Posted with Wrong Org on GL Journal

globalqss:
[IDEMPIERE-874](https://idempiere.atlassian.net/browse/IDEMPIERE-874)

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Fact.java
+++ b/org.adempiere.base/src/org/compiere/acct/Fact.java
@@ -310,6 +310,7 @@ public final class Fact
 		FactLine line = new FactLine (m_doc.getCtx(), m_doc.get_Table_ID(), 
 			m_doc.get_ID(), 0, m_trxName);
 		line.setDocumentInfo(m_doc, null);
+		line.setAD_Org_ID(m_doc.getAD_Org_ID());
 		line.setPostingType(m_postingType);
 
 		//	Account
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/acct/Fact.java
```diff
@@ -310,6 +310,7 @@ public FactLine balanceSource()
 		FactLine line = new FactLine (m_doc.getCtx(), m_doc.get_Table_ID(), 
 			m_doc.get_ID(), 0, m_trxName);
 		line.setDocumentInfo(m_doc, null);
+		line.setAD_Org_ID(m_doc.getAD_Org_ID());
```
hengsin, line 313: any idea why this is not part of the setDocumentInfo above ?

globalqss: > any idea why this is not part of the setDocumentInfo above ?

I noticed the same, and was thinking to add the code in setDocumentInfo instead, but when I saw that in setDocumentInfo the AD_Org_ID is explicitly set to zero (with the sales region), then maybe there is a reason beyond that - so, I refrained from touching that method and focused on solving isolated the two problems found.

hengsin: seem that's due to the logic in FactLine.getAD_Org_ID and FactLine.getC_SalesRegion_ID

PR #193, opened 2020-07-30
IDEMPIERE-3675:Create new AD_Session Record for Scheduler run.

globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-3675


globalqss:
Tested - the scheduler is now creating change log properly after this change, generating one session for each call.

PR #198, opened 2020-07-30
IDEMPIERE-2624 Cant confirm 0 qty on Movement Confirmation

globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-2624

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -69,7 +69,7 @@ public class MInOut extends X_M_InOut implements DocAction
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 1226522383231204912L;
+	private static final long serialVersionUID = 5791054523079936837L;
 
 	/**
 	 * 	Create Shipment From Order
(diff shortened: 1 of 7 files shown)


hengsin:
This looks confusing - isn't that the equivalent of removing the mandatory validation for getMovementQty().signum() == 0 ?

globalqss:
> This looks confusing - isn't that the equivalent of removing the mandatory validation for getMovementQty().signum() == 0 ?

The movement line cannot be saved with zero qty, still the validation is correct as it doesn't make sense to register a movement without quantity.

But the ticket is about the movement with confirmation, when the confirmation party doesn't receive the item, they can register they received zero items.  Without the patch that situation is not supported, the patch allows the qty to become zero when the document is InProgress, which is the status when there is a confirmation pending.

hengsin:
Hi @globalqss ,

It does open a loop hole with the following test case:

1. Create a movement document with confirmation with 1 movement line with qty=1, product=Mulch
2. Complete the movement document, due to confirmation, it will change to docstatus Prepare/In Progress.
3. Add a new movement line with qty=1, product=Fertilizer 50. Issue here - can save without mandatory movement qty error.
4. Open the confirmation document, change confirm qty to 0 and complete it.
5. Reload the movement doc, now it is with docstatus=Completed and both line wity movementqty=0

Regards,
Low

globalqss:
>     4. Open the confirmation document, change confirm qty to 0 and complete it.

Yes, I'm not too convinced about allowing editing a document In Progress (which iDempiere allows by default).
It can create too many issues.

In the case of sales orders, the edit on Prepare is somewhat controlled because there are reservations (and I remember we had a lot of problems and needed to add controls to fix the loopholes), but in the case of an inventory documents (we have confirmations in receipts, shipments, returns and moves).

I just checked for example changing the Qty on a material receipt In Progress, and then completing the confirmation, the numbers got messed up.

So, I think in general, we better forbid to edit those "in-transit inventory" documents when they are in progress status, WDYT?

hengsin:
I think in the current use case and the one you mention above, can't edit a document if there's pending confirmation is sufficient.

globalqss:
Ready @hengsin - validations added in the second commit

PR #202, opened 2020-08-03
IDEMPIERE-4395 Cannot save valid strings in oracle - ORA-12899: value…

globalqss:
… too large for column

[IDEMPIERE-4395](IDEMPIERE-4395)

Hi @hengsin - can you please peer review this commit?

I tested it in oracle, postgres and native postgresql and it works fine.  Migration scripts are also correctly generated.

But I'm not sure if the approach is the best, or if you have a better suggestion.
Instead of writing a complex algorithm on the convert layer, I simply added a regular expression to replace the string "VARCHAR2(30 CHAR)" with "VARCHAR(30)" - for this, I used a pattern that was not used before reusing a group of the pattern being replaced.

Applying this fix will help for new columns created with the migration script or 2packs, but I'm working on a possible fix for the whole database in [IDEMPIERE-3862](https://idempiere.atlassian.net/browse/IDEMPIERE-3862), I had an initial script, but it became too complex as there are indexes and constraints that require attention - we can postpone that, but at least stop the problem with the new columns, and users requiring this can also synchronize the specific column they need to fix.

### DIFF
--- a/org.adempiere.base/src/org/compiere/db/AdempiereDatabase.java
+++ b/org.adempiere.base/src/org/compiere/db/AdempiereDatabase.java
@@ -380,7 +380,14 @@ public interface AdempiereDatabase
 	 * @return variable length character data type name
 	 */
 	public String getVarcharDataType();
-	
+
+	/**
+	 * @return variable length character data type suffix
+	 */
+	public default String getVarcharLengthSuffix() {
+		return "";
+	};
+
 	/**
 	 * 
 	 * @return binary large object data type name
@@ -460,6 +467,7 @@ public interface AdempiereDatabase
 	 * @return alter column sql
 	 */
 	public String getSQLModify (MTable table, MColumn column, boolean setNullOption);
-		
+
+	
 }   //  AdempiereDatabase
 
--- a/org.adempiere.base/src/org/compiere/dbPort/Convert.java
+++ b/org.adempiere.base/src/org/compiere/dbPort/Convert.java
@@ -384,7 +384,7 @@ public abstract class Convert
 	
 				} catch (Exception e) {
 					String error = "Error expression: " + regex + " - " + e;
-					log.info(error);
+					log.warning(error);
 					m_conversionError = error;
 				}
 			}
--- a/org.adempiere.base/src/org/compiere/util/DisplayType.java
+++ b/org.adempiere.base/src/org/compiere/util/DisplayType.java
@@ -624,7 +624,7 @@ public final class DisplayType
 	{
 		if (columnName.equals("EntityType")
 			|| columnName.equals ("AD_Language"))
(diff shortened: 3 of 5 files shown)


hengsin:
This goes against the best practice recommended in Oracle doc, from https://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1824 - "The BYTE and CHAR qualifiers shown earlier should be avoided when possible, because they lead to mixed-semantics databases. Instead, the NLS_LENGTH_SEMANTICS initialization parameter should be set appropriately in the server parameter file (SPFILE) or initialization parameter file, and columns should use the default semantics."

hengsin:
Also, "NCHAR, NVARCHAR2, CLOB, and NCLOB columns are always character-based", perhaps a better alternative here is to always use nvarchar2 instead ?

globalqss:
Thanks @hengsin - I'll try to do tests today about using NLS_LENGTH_SEMANTICS - that's a great tip and probably the right answer to the problem.

About NVARCHAR2 - we already have a mixed-semantics database and that creates some problems.  The ticket [IDEMPIERE-3862](https://idempiere.atlassian.net/browse/IDEMPIERE-3862) was opened to migrate all NVARCHAR2 to VARCHAR2 and solve that issue.  AFAIR the NVARCHAR2 have caused issues to us, but I'll analyze also that case.

hengsin:
Hi @globalqss , I've actually deleted my earlier comments on NLS_LENGTH_SEMANTICS as I found out that's from old doc (11g) and now (19c doc) they(Oracle) have change their recommendation to specify the length semantics (i.e 10 CHAR or 10 BYTE) instead. Also, they have now specifically say you shouldn't set NLS_LENGTH_SEMANTICS at spfile as some of their installation scripts is not compatible with the CHAR semantics (I think part of the sys schema as well). Their recommendation is to use ALTER SESSION OR ALTER SYSTEM instead (https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/NLS_LENGTH_SEMANTICS.html#GUID-221B0A5E-A17A-4CBC-8309-3A79508466F9).

Anyway, if possible, the NVARCHAR2 solution is the cleanest approach.

globalqss:
> Anyway, if possible, the NVARCHAR2 solution is the cleanest approach.

Thanks @hengsin - added my analysis at [IDEMPIERE-3862](https://idempiere.atlassian.net/browse/IDEMPIERE-3862?focusedCommentId=45738&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45738)

globalqss:
This is ready for peer review, please see my comments at [IDEMPIERE-4395](https://idempiere.atlassian.net/browse/IDEMPIERE-4395?focusedCommentId=45752&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-45752)

PR #203, opened 2020-08-04
IDEMPIERE-4396 Allow Idempiere forms sort by header

igorpojzl:
Force ZK Model Sync after Sorting in forms.

https://idempiere.atlassian.net/browse/IDEMPIERE-4396


hengsin:
Can't approve this without as test case.

globalqss:
Hi @hengsin - test case is in [IDEMPIERE-4367](https://idempiere.atlassian.net/browse/IDEMPIERE-4367).

I checked Payment Allocation and it works fine after the patch, can you please check if the approach is correct, this is an issue on zk9

hengsin:
replace with https://github.com/idempiere/idempiere/pull/205 that is less likely to have side-effect.

PR #206, opened 2020-08-05
IDEMPIERE-4389

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4389
Implement Copy Constructor support for PO

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
+++ b/org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
@@ -21,10 +21,10 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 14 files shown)


org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Random;
 import java.util.logging.Level;
 
+import org.bouncycastle.util.Arrays;
```
globalqss, line 27: Hi @hengsin - is this OK or is it a typo?
Maybe you wanted to use the java.util.Arrays instead of introducing bouncycastle dependency?

hengsin: good catch. wrong import.

PR #215, opened 2020-08-19
IDEMPIERE-4413 Change Role, disfunction all keyboard shortcuts

hengsin:
Fix key listener error after change role

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/AdempiereWebUI.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/AdempiereWebUI.java
@@ -561,6 +561,12 @@ public class AdempiereWebUI extends Window implements EventListener<Event>, IWeb
     			root.detach();
     	}
     	this.setPage(page);
+    	
+    	//clear key listeners
+    	Iterable<EventListener<? extends Event>> keylisteners = keyListener.getEventListeners(Events.ON_CTRL_KEY);
+    	for(EventListener<? extends Event> kl : keylisteners) {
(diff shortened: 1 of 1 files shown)


hieplq:
solution on PR is ok it keep Keylistener when change to role

i attach my solution by fix javascript side, it don't conflict with this solution
now can consider my solution as improve on case we need to remove/re-add Keylistener

[0001-IDEMPIERE-4413-remove-client-side-handle-key-event.patch.txt](https://github.com/idempiere/idempiere/files/5097465/0001-IDEMPIERE-4413-remove-client-side-handle-key-event.patch.txt)

hieplq:
after review patch @globalqss attach on ticket then i consider apply only my patch is better
1. it resolve issue report by Norbert
2. don't need worry what @globalqss mention, because Keylistener, HeaderPanel, UserPanel all be GC so don't need to call remove listen
3. it reset change role panel to same case we login from home and choose "Select Role"

globalqss:
Hi @hengsin - I tested your commits and they are working fine.

However, I think the patch posted by @hieplq here:
[0001-IDEMPIERE-4413-remove-client-side-handle-key-event.patch.txt](https://github.com/idempiere/idempiere/files/5097465/0001-IDEMPIERE-4413-remove-client-side-handle-key-event.patch.txt)
seems like a cleaner solution, and possibly better for future in case somebody does something similar preserving the Keylistener object.

Or, it can be as Hiep suggested that we add the patch on top of this, however, according with my tests, the patch from Hiep alone is enough and looks like a good solution.

WDYT?

Regards,

Carlos Ruiz

hengsin:
yeah, if that's working fine, it is a cleaner solution.

PR #216, opened 2020-08-20
IDEMPIERE-4413:Change Role, disfunction all keyboard shortcuts

hieplq:
remove keydown handle at client side


globalqss (approved):
tested

PR #224, opened 2020-08-22
IDEMPIERE-4425 Find Window improvements

globalqss:
Fix for printing report parameters - it is printing OPERATOR_EQUAL on reports instead of =


hengsin (approved):
looks good

PR #225, opened 2020-08-22
IDEMPIERE-1482 Invoice with Payment Schedule is not displayed in Invo…

globalqss:
…ice Not Realized Gain/Loss

Fix problem reported by Anozi Mada
Fix issue found when testing - postgresql was failing with error
   org.postgresql.util.PSQLException: ERROR: la columna «c_doctypereval_id» es de tipo numeric pero la expresión es de tipo text
Avoid creating empty journal


hengsin (approved):
Looks good.

PR #227, opened 2020-08-22
IDEMPIERE-1709 / IDEMPIERE-2624 Cant confirm 0 qty on Movement Confir…

globalqss:
…mation

Implement void for movement confirm

https://idempiere.atlassian.net/browse/IDEMPIERE-1709


hengsin:
hi @globalqss , I'm able to Complete the Movement document after I've voided the Movement Confirmation document. Is that the expected behaviour ?

globalqss:
Thanks @hengsin - I added code to void the inventory move when the confirmation is voided.

globalqss:
Hi @hengsin - pushed a third commit because during tests found that MInOut is suffering from the same catch-22 situation, cannot void the InOut because there are confirmations, cannot void the confirmation because the InOut is not void.

As I saw the implementation here was different and it was to void the confirmations when the InOut is void, so I added a bit of code to restore that flow.

hengsin (approved):
good to go

PR #228, opened 2020-08-22
IDEMPIERE-4429 BOM Drop to Invoice is throwing error

globalqss:
IDEMPIERE-1314 getDocumentInfo methods should return translated String


hengsin (approved):
looks good.

PR #229, opened 2020-08-23
IDEMPIERE-3197 Tab still keep editing indicator when move to next record

hengsin:
Update tab title after save


globalqss (approved):
Tested successfully

PR #235, opened 2020-08-25
Idempiere 4066

dpansheriya:
This enhancement adds RelativePeriodTo on report column. Using RelativePeriod and RelativePeriodTo, Range of periods (Inclusive) can specified.

If RelativePeriodTo is null, then it behaves as prior to this enhancement.
This enhancement are part of Adaxa Suite's Financial report improvements.  Test cases are mentioned on Ticket.

### DIFF
--- a/migration/i7.1z/postgresql/201912301730_IDEMPIERE-4066.sql
+++ b/migration/i7.1z/postgresql/201912301730_IDEMPIERE-4066.sql
@@ -4,7 +4,7 @@ INSERT INTO AD_Element (AD_Element_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,Cr
 ;
 
 -- Oct 14, 2019 4:06:50 PM IST
-INSERT INTO AD_Column (AD_Column_ID,Version,Name,Description,AD_Table_ID,ColumnName,DefaultValue,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure) VALUES (214087,0,'Relative Period To','Period offset (0 is current)',446,'RelativePeriodTo','0',22,'N','N','N','N','N',0,'N',22,0,0,'Y',TO_TIMESTAMP('2019-10-14 16:06:50','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2019-10-14 16:06:50','YYYY-MM-DD HH24:MI:SS'),100,203370,'Y','N','D','N','N','N','Y','48d4dd26-197e-4a9a-add3-f2e1d5ed0e22','Y',0,'N','N')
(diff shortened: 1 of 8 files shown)


migration/i7.1z/postgresql/201912301730_IDEMPIERE-4066.sql
```diff
@@ -0,0 +1,35 @@
...
+-- Oct 14, 2019 4:09:10 PM IST
+INSERT INTO AD_Field (AD_Field_ID,Name,Description,AD_Tab_ID,AD_Column_ID,IsDisplayed,DisplayLogic,DisplayLength,SeqNo,IsSameLine,IsHeading,IsFieldOnly,IsEncrypted,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,IsReadOnly,IsCentrallyMaintained,EntityType,AD_Field_UU,IsDisplayedGrid,SeqNoGrid,XPosition,ColumnSpan,NumLines,IsQuickEntry,IsDefaultFocus,IsAdvancedField) VALUES (206213,'Relative Period To','Period offset (0 is current)',374,214087,'Y','@ColumnType@=R',26,145,'N','N','N','N',0,0,'Y',TO_TIMESTAMP('2019-10-14 16:09:10','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2019-10-14 16:09:10','YYYY-MM-DD HH24:MI:SS'),100,'N','Y','D','ebe78b9f-d030-46fa-b1c0-eaafe2521d3a','Y',145,4,2,1,'N','N','N')
+;
+
+-- Nov 26, 2019 4:04:43 PM IST
+UPDATE AD_Column SET DefaultValue=NULL,Updated=TO_TIMESTAMP('2019-11-26 16:04:43','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=214087
+;
+
+-- Nov 26, 2019 4:04:46 PM IST
+INSERT INTO t_alter_column values('pa_reportcolumn','RelativePeriodTo','NUMERIC',null,'NULL')
```
globalqss, line 23: There is a mismatch between the oracle and postgres scripts - this line is not there, so maybe is better to review if is OK, or needs to be fixed, or not needed

org.adempiere.base/src/org/compiere/model/GridField.java
```diff
@@ -156,6 +156,8 @@ protected void dispose()
 	/** Indicator for new Value				*/
 	public static final String  INSERTING = "FieldValueInserting";
 
+	private static final int	PA_REPORTCOLUMN_RelativePeriodTo_COLUMN_ID	= 214087;
```
globalqss, line 159: This hardcode must go to the org.compiere.model.SystemIDs class

org.adempiere.base/src/org/compiere/report/MReportColumn.java
```diff
@@ -442,6 +442,20 @@ public BigDecimal getRelativePeriod()
...
 		return null;
 	}	//	getRelativePeriod
+
+	/**
+	 * Get Relative Period To
+	 * 
+	 * @return relative period to
+	 */
+	@Override
+	public BigDecimal getRelativePeriodTo()
```
globalqss, line 452: when adding public methods - the serialVersionUID changes

globalqss (changes requested):
I didn't test because it has compilation errors, please fix the requested and test.

dpansheriya:
Ohk, Let me work on this fix.

migration/i7.1z/postgresql/201912301730_IDEMPIERE-4066.sql
dpansheriya: This line is setting default value to null, while same is taken care in oracle script by 
`ALTER TABLE PA_ReportColumn ADD RelativePeriodTo NUMBER DEFAULT NULL
;`

dpansheriya:
I committed changes on IDEMPIERE-4066 branch and looks pull request auto updated. This is great features. @globalqss  now you are good to test.

globalqss:
@dpansheriya - you didn't answer - why is that code required in GridField?  what is special with the default of that column?

dpansheriya:
> @dpansheriya - you didn't answer - why is that code required in GridField? what is special with the default of that column?

RelativePeriod_To is numeric. It was always getting defaulted to 0 but we want to use null as default. Null means not provided. 0 means selected current period on report process parameter. To avoid blank getting converted to 0, we added this.

globalqss:
> > @dpansheriya - you didn't answer - why is that code required in GridField? what is special with the default of that column?
> 
> RelativePeriod_To is numeric. It was always getting defaulted to 0 but we want to use null as default. Null means not provided. 0 means selected current period on report process parameter. To avoid blank getting converted to 0, we added this.

OK, then that code is not required, you can define the default to NULL and the numeric fields will behave as expected.
That was added by @nmicoud in ticket IDEMPIERE-2678

dpansheriya:
@globalqss Let me test this and if work then will revert.

dpansheriya:
@globalqss 

I restored seed DB, applied corrected migration script on postgresql DB and tested below scenarios with reverted GridField code.
1. After Applying migration script, existing records in PA_ReportColumn have null value in relativePeriodTo (Verified in DB)
2. Set value relativePeriodTo = 5 and save
3. Making relativePeriodTo field blank and save. It reset value to null (Verified in DB)

globalqss (approved):

PR #236, opened 2020-08-25
IDEMPIERE-4187 implement the usage of workflows in processes

hengsin:
Implement chaining of workflow through the use of Process Action
Implement overwrite of Record_ID for Process through Node Parameter


globalqss (approved):
tested successfully

PR #237, opened 2020-08-26
IDEMPIERE-4433 - MTable error - ConcurrentModificationException

igorpojzl:
Fix Putting into Cache outside Synchronized methods in MTable.

https://idempiere.atlassian.net/browse/IDEMPIERE-4433


hengsin (approved):
looks good.

PR #240, opened 2020-08-30
IDEMPIERE-4361:Date is subtracted by 1 day in certain cases

hieplq:
sun.util.calendar.ZoneInfo consider offset and "daylight saving time" changed on range 1900-2037
for out of that range then offset always return raw offset, make date incorrect 1 day on some days

offset change by history can reference at:
https://en.wikipedia.org/wiki/Time_in_Vietnam#Time_in_French_Indochina
https://www.timeanddate.com/time/zone/vietnam/hanoi

database for all change
https://www.iana.org/time-zones


globalqss:
The patch solves the issue for user.timezone=Asia/Ho_Chi_Minh, but still issues with user.timezone=Europe/Berlin, see comments in ticket.

globalqss:
Tested, still issues - more comments on ticket [IDEMPIERE-4361](https://idempiere.atlassian.net/browse/IDEMPIERE-4361)

globalqss (approved):

PR #241, opened 2020-09-02
IDEMPIERE-4440 NPE in DocAllocationHdr

globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4440

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
+++ b/org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
@@ -420,7 +420,8 @@ public class Doc_AllocationHdr extends Doc
 						getC_Currency_ID(), null, line.getAmtSource().negate());
 					if (fl != null && payment != null)
 						fl.setAD_Org_ID(payment.getAD_Org_ID());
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -420,7 +420,8 @@ else if (line.getC_CashLine_ID() != 0)
 						getC_Currency_ID(), null, line.getAmtSource().negate());
 					if (fl != null && payment != null)
 						fl.setAD_Org_ID(payment.getAD_Org_ID());
-					allocPayAccounted = allocPayAccounted.add(fl.getAcctBalance().negate());
+					if (fl != null)
+						allocPayAccounted = allocPayAccounted.add(fl.getAcctBalance().negate());
```
kriegfrj, line 424: `fl` can't spontaneously become null after the first call, so what about this as slightly more efficient: 

```
if (fl != null) {
  if (payment != null) {
    fl.setAD_Org_ID(payment.getAD_Org_ID());
  }
  allocPayAccounted = allocPayAccounted.add(fl.getAcctBalance().negate());
}

PR #243, opened 2020-09-04
IDEMPIERE-4447 Add offset to Query

globalqss:
Add method Query.setRecordstoSkip

https://idempiere.atlassian.net/browse/IDEMPIERE-4447


d-ruiz (approved):
looks good and it doesn't break previous functionality

PR #244, opened 2020-09-05
IDEMPIERE-4067 : Support for Under/Over line and Blank line in finaci…

dpansheriya:
…al report line.

For testing, After applying migration script, Report Line Set -> Report line has two new field
1. Overline Stroke type
2. Underline Stroke Type
Selecting value on this field and refreshing PDF report will shows elected type of lines on column section.

Also in Line Type there is new option Blank line visible. Selecting it will make blank line printed.

Original work done by @pawlbouden from Adaxa in adaxasuite. Logilite has bring his work to iDempiere.

### DIFF
--- a/migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
+++ b/migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
@@ -39,7 +39,7 @@ INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLen
 ;
 
 -- Oct 16, 2019 2:43:31 PM IST
-ALTER TABLE PA_ReportLine ADD OverlineStrokeType VARCHAR2(3) DEFAULT NULL 
+ALTER TABLE PA_ReportLine ADD OverlineStrokeType VARCHAR2(3 CHAR) DEFAULT NULL 
 ;
 
 -- Oct 16, 2019 2:44:08 PM IST
@@ -51,7 +51,7 @@ INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLen
 ;
 
 -- Oct 16, 2019 2:45:39 PM IST
-ALTER TABLE PA_ReportLine ADD UnderlineStrokeType VARCHAR2(3) DEFAULT NULL 
+ALTER TABLE PA_ReportLine ADD UnderlineStrokeType VARCHAR2(3 CHAR) DEFAULT NULL 
 ;
 
 -- Oct 16, 2019 2:48:07 PM IST
--- a/migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
+++ b/migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
@@ -1,0 +1,72 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Adding Over/Under line in Report
+-- Oct 16, 2019 2:39:54 PM IST
+INSERT INTO AD_Element (AD_Element_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,ColumnName,Name,PrintName,EntityType,AD_Element_UU) VALUES (203371,0,0,'Y',TO_DATE('2019-10-16 14:39:53','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2019-10-16 14:39:53','YYYY-MM-DD HH24:MI:SS'),100,'OverlineStrokeType','Overline Stroke Type','Overline Stroke Type','D','18f9193c-b4a9-4dc2-ba69-8dcefc4d8e88')
+;
+
+-- Oct 16, 2019 2:40:47 PM IST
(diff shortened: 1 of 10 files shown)


migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
```diff
@@ -0,0 +1,72 @@
...
+;
+
+-- Oct 16, 2019 2:43:26 PM IST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Reference_Value_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure) VALUES (214088,0,'Overline Stroke Type',448,'OverlineStrokeType',3,'N','N','N','N','N',0,'N',17,200174,0,0,'Y',TO_DATE('2019-10-16 14:43:26','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2019-10-16 14:43:26','YYYY-MM-DD HH24:MI:SS'),100,203371,'Y','N','D','N','N','N','Y','b1529a0e-96a5-43e6-94b6-d78feea8e8c8','Y',0,'N','N')
+;
+
+-- Oct 16, 2019 2:43:31 PM IST
+ALTER TABLE PA_ReportLine ADD OverlineStrokeType VARCHAR2(3) DEFAULT NULL
```
globalqss, line 42: Please change this line to:
ALTER TABLE PA_ReportLine ADD OverlineStrokeType VARCHAR2(3 **CHAR**) DEFAULT NULL

Now we are creating the columns in oracle using that CHAR syntax

migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
```diff
@@ -0,0 +1,72 @@
...
+;
+
+-- Oct 16, 2019 2:45:23 PM IST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Reference_Value_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure) VALUES (214089,0,'Underline Stroke Type',448,'UnderlineStrokeType',3,'N','N','N','N','N',0,'N',17,200174,0,0,'Y',TO_DATE('2019-10-16 14:45:23','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2019-10-16 14:45:23','YYYY-MM-DD HH24:MI:SS'),100,203372,'Y','N','D','N','N','N','Y','d3504292-068e-4126-944c-dc3d165019c0','Y',0,'N','N')
+;
+
+-- Oct 16, 2019 2:45:39 PM IST
+ALTER TABLE PA_ReportLine ADD UnderlineStrokeType VARCHAR2(3) DEFAULT NULL
```
globalqss, line 54: Same here - 3 CHAR

org.adempiere.base/src/org/compiere/print/layout/TableElement.java
```diff
@@ -1725,6 +1813,22 @@ private void addPrintLines (int row, int col, Serializable data)
...
+	 */
+	private MReportLine getReportLine(int row, int col)
+	{
+		// First specific position
+		MReportLine rLine = (MReportLine) m_rowColReportLine.get(new Point(row, col));
+		if (rLine != null)
+			return rLine;
+
+		return null;
+	} // getFont
```
globalqss, line 1831: This is:
// getReportLine

globalqss (changes requested):
The tests went OK.

The only strange thing I noticed is:
- for blank lines - the Name and Description are hidden in PDF format, but they are printed in all the other formats (HTML, XLS, etc.)

Maybe is good to hide the Name in all the formats.

But I think is interesting to allow showing the Description in all formats (also in PDF), as that could be exploited to add some annotations to the reports.
If the description is left blank, then there is no harm, if is filled, then it can help as annotation or comments to the report.

WDYT?

migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
dpansheriya: Sure, will fix in next commit.

migration/i7.1z/oracle/201912301800_IDEMPIERE-4067.sql
dpansheriya: Sure, will fix in next commit.

org.adempiere.base/src/org/compiere/print/layout/TableElement.java
dpansheriya: Yes, this copy paste error. Good catch.

dpansheriya:
@globalqss I am agree with you on Always showing description. As it is not mandatory.
Blank Type is not respected on HTML and XLS, XLSX are our miss. will fix this in next commit.

adaxa-steven:
When adding a 'blank line' in the Report Line Set I always had to enter 3 space characters in the Name filed because Name is mandatory and I could not save without some input.  Although it worked it always seemed an ugly kludge. Perhaps it has already been improved?

dpansheriya:
@adaxa-steven Name is used to identify line so it is required. With this feature, it will not printed so we can give meaningfull name.

@globalqss Please note that second commit taking care of your feedback.

globalqss (approved):

PR #248, opened 2020-09-12
IDEMPIERE-4449 FreeBSD sh shell rc init script

globalqss:
Integrate rc script contributed by James B. Byrne

https://idempiere.atlassian.net/browse/IDEMPIERE-4449


globalqss:
Closed as a new version was added in IDEMPIERE-4457 - we need a tester for this

PR #253, opened 2020-09-14
Idempiere 4441

globalqss:


globalqss:
bad pull request - closing to recreate it

PR #254, opened 2020-09-14
IDEMPIERE-4441 More shellcheck

globalqss:
Fix issue caused by previous commit - RUN_SyncDB broken


globalqss:
Committed with admin permission as current master is broken

PR #257, opened 2020-09-16
IDEMPIERE-4460 : User suggestions for field/context help : check perm…

nmicoud:
…ission

new method to check exclusion of a table for current role, used to determine if popups should be displayed or not

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRole.java
+++ b/org.adempiere.base/src/org/compiere/model/MRole.java
@@ -3349,16 +3349,20 @@ public final class MRole extends X_AD_Role
 	 * @param tableID ID of the table
 	 * @return
 	 */
-	public boolean isTableExcluded(int tableID)
+	public boolean isTableAccessExcluded(int tableID)
 	{
 		loadTableAccess(false);
 
 		for (int i = 0; i < m_tableAccess.length; i++) {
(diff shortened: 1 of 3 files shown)


org.adempiere.base/src/org/compiere/model/MRole.java
```diff
@@ -3344,4 +3344,21 @@ private boolean isAccessLevelCompatible(String accessLevel) {
...
+	 * Does the table is excluded for current role (this method doesn't check the level of the table, use isTableAccess for this purpose)
+	 * @param tableID ID of the table
+	 * @return
+	 */
+	public boolean isTableExcluded(int tableID)
+	{
+		loadTableAccess(false);
+
+		for (int i = 0; i < m_tableAccess.length; i++) {
+			if (m_tableAccess[i].isExclude() && m_tableAccess[i].getAD_Table_ID() == tableID)
```
globalqss, line 3357: Hi @nmicoud - I tested the patch successfully, however, because the table access can be defined for many other things is better to rename this method to isTableAccessExcluded, and the validation must verify access type and readonly, like suggested here:
[mod257.txt](https://github.com/idempiere/idempiere/files/5251702/mod257.txt)

globalqss (changes requested):

nmicoud:
hi @globalqss ; done
Thanks for the review + patch

Nicolas

globalqss (approved):

PR #260, opened 2020-09-17
IDEMPIERE-4442 FIX NPE when locatorField is Empty

igorpojzl:
Fix NPE when we confirm Locator Dialog with empty Locator Field.

To test:
1. Open any Product which have its Locator selected.
2. Deselect Locator through Locator Dialog.


hengsin (approved):
tested

PR #261, opened 2020-09-18
IDEMPIERE-4461 : Feedback email recipient(s) : add flexibility

nmicoud:
Search recipients in SysConfig and if not found, fallback on System.SupportEMail

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
@@ -116,18 +116,20 @@ public class DefaultFeedbackService implements IFeedbackService {
 			
 			MSystem system = MSystem.get(Env.getCtx());
 
-			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_TO).split(","))
-				if(EMail.validate(s.trim()))
-					dialog.addTo(s.trim(), Util.isEmpty(dialog.getTo()));
+			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_TO).split(",")) {
+				if (!Util.isEmpty(s.trim()) && EMail.validate(s.trim()))
+					dialog.addTo(s.trim(), false);
+			}
 
 			if (Util.isEmpty(dialog.getTo()) && !Util.isEmpty(system.getSupportEMail())) 
 			{
 				dialog.addTo(system.getSupportEMail(), true);
 			}
 
-			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_CC).split(","))
-				if(EMail.validate(s.trim()))
-					dialog.addCC(s.trim(), Util.isEmpty(dialog.getCc()));
+			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_CC).split(",")) {
+				if (!Util.isEmpty(s.trim()) && EMail.validate(s.trim()))
+					dialog.addCC(s.trim(), false);
+			}
 
 			AEnv.showWindow(dialog);
 			if (imageBytes != null && imageBytes.length > 0) {
@@ -136,6 +138,11 @@ public class DefaultFeedbackService implements IFeedbackService {
 				dialog.addAttachment(screenShot, true);
 			}
 			dialog.focus();
+		}
+		
+		protected String getFeedbackRecipient(String scValue) {
+			String retValue = MSysConfig.getValue(scValue, Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
+			return Util.isEmpty(retValue) ? "" : retValue;
 		}
 	}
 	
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
```diff
@@ -113,10 +115,20 @@ protected void showEmailDialog(byte[] imageBytes) {
 			dialog.setAttribute(Window.MODE_KEY, Mode.OVERLAPPED);			
 			
 			MSystem system = MSystem.get(Env.getCtx());
-			if (!Util.isEmpty(system.getSupportEMail())) 
+
+			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_TO).split(","))
+				if(EMail.validate(s.trim()))
```
hengsin, line 120: Should check whether s is empty since EMail.validate is kind of heavy and can product warning to server log if s is empty string.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
```diff
@@ -113,10 +115,20 @@ protected void showEmailDialog(byte[] imageBytes) {
...
+					dialog.addTo(s.trim(), Util.isEmpty(dialog.getTo()));
```
hengsin, line 121: if it is always append, the second argument should always be false

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
```diff
@@ -113,10 +115,20 @@ protected void showEmailDialog(byte[] imageBytes) {
...
+
+			if (Util.isEmpty(dialog.getTo()) && !Util.isEmpty(system.getSupportEMail())) 
 			{
 				dialog.addTo(system.getSupportEMail(), true);
 			}
+
+			for (String s : getFeedbackRecipient(MSysConfig.FEEDBACK_EMAIL_CC).split(","))
+				if(EMail.validate(s.trim()))
```
hengsin, line 129: same as above, don't call EMail.validate if s is an empty string

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
```diff
@@ -113,10 +115,20 @@ protected void showEmailDialog(byte[] imageBytes) {
...
+					dialog.addCC(s.trim(), Util.isEmpty(dialog.getCc()));
```
hengsin, line 130: same here, Util.isempty check is unnecessary, should just pass false as parameter.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
```diff
@@ -169,4 +181,9 @@ protected void showRequestDialog(byte[] imageBytes) {
 			window.focus();
 		}
 	}
+	
+	static String getFeedbackRecipient(String scValue) {
```
hengsin, line 185: this should be private or protected and without the static qualifier.

hengsin (changes requested):

CarlosRuiz-globalqss:
Closing after 28 days of inactivity - please reopen after fixing the suggested changes

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
nmicoud: Condition is true with the first recipient, then is false.
If I hardcode it to false, we can have only one recipient

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
nmicoud: If I remove static, I got following error:
`Cannot make a static reference to the non-static method getFeedbackRecipient(String) from the type DefaultFeedbackService`
for getFeedbackRecipient

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
hengsin: No, that only determine whether the email is add to front or append to end.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultFeedbackService.java
hengsin: move the method inside the static class EmailSupportAction

nmicoud:
Hello there,
Sorry fot the lateness, I didn't get email notifications
Regards,
Nicolas

hengsin (approved):
tested

PR #265, opened 2020-09-20
IDEMPIERE-4213 Window Toolbar attached processes are doesn't validate…

globalqss:
… role access

https://idempiere.atlassian.net/browse/IDEMPIERE-4213

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridField.java
+++ b/org.adempiere.base/src/org/compiere/model/GridField.java
@@ -505,6 +505,13 @@ public class GridField
 					return false;
 				if (!MRole.getDefault(ctx, false).isColumnAccess(AD_Table_ID, m_vo.AD_Column_ID, false))
 					return false;
+				if (getDisplayType() == DisplayType.Button && getAD_Process_ID() > 0) {
+					// Verify access to process for buttons
+					Boolean access = MRole.getDefault().getProcessAccess(getAD_Process_ID());
+					if (access == null || !access.booleanValue())
+						return false;
+				}
+				
 			}
 		}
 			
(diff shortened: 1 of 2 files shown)


igorpojzl:
Hi, i tested this PR. It is Working.
I have one question. This solution will apply on Menu Buttons too. So for example if User do not have Access to Complete Sales Order, in Sales Order Window he will not see Document Action button in Grid too. Is this functionality desirable?
Screenshot 2020-09-21 at 12 59 58

hengsin:
I guess that could cause some confusion since we do have access control for
document action too.

globalqss:
The permissions for document action is for the options shown - so, it is about if a user can Complete/Void ... etc

I think is OK if the button is not shown when the user doesn't have access to the process.

I still managed the case if we want to show the button read-only, then you can define the access to the process as read-only.

globalqss:
In other words:

The actual behavior is like this:

- if the user doesn't have access (or just read-only access) to the process - the button is shown and active, when the user pushes the button and select complete then an error is thrown: "You cannot access process [Process Order] with your role : GardenWorld User"

- if the user have access to the process, but doesn't have access to any document action - then the button is shown and when the user pushes the button it doesn't do anything - not even open the dialog

With this patch:

- if the user doesn't have access to the process - the button is not shown

- if the user has read-only access to the process - the button is shown disabled

- the behavior based on document action was not modified

So, I think in terms of usability is better - you have the two options - hide it, or show it disabled - what I think is bad on usability is showing a button active that is going to throw an error to the user when pushed.

hengsin (approved):
tested by Igor and Carlos

PR #266, opened 2020-09-21
IDEMPIERE 4451 Executions.schedule should  be executed asynchronously

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4451


globalqss (approved):
tested

PR #267, opened 2020-09-21
IDWS-8 Cannot update user info in WebStore

globalqss:
IDEMPIERE-4287

https://idempiere.atlassian.net/browse/IDWS-8


hengsin (approved):
looks good.

PR #271, opened 2020-09-21
Idempiere 4452

mschnbeck:
Create purchase pricelists for a vendor which is not the current vendor and allow to select entries by vendorcategory.

While at it corrected a missing m_AD_PInstance_ID when reading the price list lines.


migration/i7.1z/oracle/202009171400_IDEMPIERE-4452.sql
```diff
@@ -0,0 +1,51 @@
...
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Description,Help,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,IsHtml) VALUES (214316,0,'Partner Category','Product Category of the Business Partner','The Business Partner Category identifies the category used by the Business Partner for this product.',477,'VendorCategory',30,'N','N','N','N','N',0,'N',10,0,0,'Y',TO_DATE('2020-09-17 08:16:18','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2020-09-17 08:16:18','YYYY-MM-DD HH24:MI:SS'),100,622,'Y','N','D','N','N','N','Y','d3bd3b17-d48d-43d7-9d79-f58aad219bcc','Y',0,'N','N','N')
+;
+
+-- Sep 17, 2020, 8:19:51 AM UTC
+ALTER TABLE M_DiscountSchemaLine ADD IsIgnoreIsCurrentVendor CHAR(1) DEFAULT 'N' CHECK (IsIgnoreIsCurrentVendor IN ('Y','N')) NOT NULL
+;
+
+-- Sep 17, 2020, 8:20:37 AM UTC
+ALTER TABLE M_DiscountSchemaLine ADD VendorCategory VARCHAR2(30) DEFAULT NULL
```
hengsin, line 22: we have change to use 30 char (i.e Varchar2(30 Char)) for Oracle

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -218,11 +215,13 @@ private String create() throws Exception {
...
 						+ " INNER JOIN M_DiscountSchemaLine dl ON (dl.M_DiscountSchemaLine_ID=?) "	//	#2
 						+ "WHERE p.AD_Client_ID IN (?, 0)"		//	#3
-						+ " AND p.IsActive='Y' AND po.IsActive='Y' AND po.IsCurrentVendor='Y'"
+						+ " AND p.IsActive='Y' AND po.IsActive='Y'"
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR po.C_BPartner_ID=dl.C_BPartner_ID)"
+						+ " AND (dl.VendorCategory IS NULL OR po.VendorCategory=dl.VendorCategory)"
+						+ " AND (dl.C_BPartner_ID IS NULL AND po.IsCurrentVendor='Y'"
```
hengsin, line 223: please use parenthesis to make this easier to read and remove any potential ambiguaties

migration/i7.1z/oracle/202009171400_IDEMPIERE-4452.sql
mschnbeck: Ok.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Yes, it's better.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -117,10 +117,7 @@ private String checkPrerequisites() {
 				"UPDATE M_Product_PO SET PriceLastPO = 0 WHERE PriceLastPO IS NULL AND AD_Client_ID=?",
 				clientParam, get_TrxName());
 		DB.executeUpdateEx(
-				"UPDATE M_Product_PO SET PricePO = PriceLastPO WHERE (PricePO IS NULL OR PricePO = 0) AND PriceLastPO <> 0 AND AD_Client_ID=?",
-				clientParam, get_TrxName());
-		DB.executeUpdateEx(
-				"UPDATE	M_Product_PO SET PricePO = 0 WHERE PricePO IS NULL AND AD_Client_ID=?",
+				"UPDATE M_Product_PO SET PricePO = PriceLastPO WHERE PricePO IS NULL OR (PricePO = 0 AND PriceLastPO <> 0) AND AD_Client_ID=?",
```
hengsin, line 120: what does it fix here ?

mschnbeck: Hi,
not a fix. Only a small enhancement because it replaces the previously following update.

globalqss: Doesn't sound like an enhancement.

The change in the WHERE sounds wrong:
was:
(PricePO IS NULL OR PricePO = 0) AND PriceLastPO <> 0 AND AD_Client_ID=?
changed to:
PricePO IS NULL OR (PricePO = 0 AND PriceLastPO <> 0) AND AD_Client_ID=?

That's a very different set of records, I have not tested, but I have the feeling it breaks some scenarios.

I think is better not to mix features in the middle of another ticket, if a small enhancement is possible, then better to discuss it.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -117,10 +117,7 @@ private String checkPrerequisites() {
 				"UPDATE M_Product_PO SET PriceLastPO = 0 WHERE PriceLastPO IS NULL AND AD_Client_ID=?",
 				clientParam, get_TrxName());
 		DB.executeUpdateEx(
-				"UPDATE M_Product_PO SET PricePO = PriceLastPO WHERE (PricePO IS NULL OR PricePO = 0) AND PriceLastPO <> 0 AND AD_Client_ID=?",
-				clientParam, get_TrxName());
-		DB.executeUpdateEx(
-				"UPDATE	M_Product_PO SET PricePO = 0 WHERE PricePO IS NULL AND AD_Client_ID=?",
```
globalqss, line 123: This UPDATE is important, why it was deleted?

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -240,10 +239,12 @@ private String create() throws Exception {
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR EXISTS "
-						+ "(SELECT * FROM M_Product_PO po "
-						+ "WHERE po.M_Product_ID=p.M_Product_ID AND po.C_BPartner_ID=dl.C_BPartner_ID))"
-						+ " AND (dl.M_Product_ID IS NULL OR p.M_Product_ID=dl.M_Product_ID)");
+						+ " AND (dl.C_BPartner_ID IS NULL AND dl.VendorCategory IS NULL OR EXISTS "
```
globalqss, line 242: The change here looks wrong:
from
dl.C_BPartner_ID IS NULL OR EXISTS
to
dl.C_BPartner_ID IS NULL AND dl.VendorCategory IS NULL OR EXISTS

The AND takes precedence, if is intentional better to make it explicit with parenthesis, but I think is not, what if a record has set the BP but not the category?  or viceversa?

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -295,7 +296,9 @@ else if (M_Pricelist_Version_Base_ID == 0) {
 						+ " INNER JOIN M_DiscountSchemaLine dl ON (dl.M_DiscountSchemaLine_ID=?) "	//	#3
 						//
 						+ "WHERE EXISTS (SELECT * FROM T_Selection s WHERE s.AD_PInstance_ID=? AND po.M_Product_ID=s.T_Selection_ID)"	//	#4
-						+ " AND po.IsCurrentVendor='Y' AND po.IsActive='Y'";
+						+ " AND (dl.C_BPartner_ID IS NULL AND po.IsCurrentVendor='Y'"
```
globalqss, line 299: This change doesn't seem equivalent:
from:
AND po.IsCurrentVendor='Y'
to:
AND (dl.C_BPartner_ID IS NULL AND po.IsCurrentVendor='Y'
OR po.C_BPartner_ID=dl.C_BPartner_ID AND (dl.IsIgnoreIsCurrentVendor='Y' OR po.IsCurrentVendor='Y'))"

Seems like a change on the previous feature, where is checking IsCurrentVendor independently of the C_BPartner configuration.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -336,7 +339,7 @@ else if (M_Pricelist_Version_Base_ID == 0) {
 			/** Calculations	**/
 			MProductPrice[] pp = m_plv.getProductPrice(
 					"AND EXISTS (SELECT * FROM T_Selection s "
-							+ "WHERE s.T_Selection_ID=M_ProductPrice.M_Product_ID)");
+							+ "WHERE s.AD_PInstance_ID = " + m_AD_PInstance_ID + " AND s.T_Selection_ID=M_ProductPrice.M_Product_ID)");
```
globalqss, line 342: This one looks correct, like fixing a bug, I think is better to fix bugs in independent ticket - as is possible that the improvement is delayed, but the bug needs to be fixed a.s.a.p.

globalqss (changes requested):

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Hi @globalqss 
I had done so if I had detected it earlier but I found it while testing and wondering why prices where changed which were not part of the selection. If adoption of this pull request really takes longer, I'll separate it.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Hi Carlos, @globalqss 
the part behaves exactly as before if IsIgnoreIsCurrentVendor is N. If bpartner is null then IsCurrentVendor defines the result and if bpartner is not null it defines result also as long as IsIgnoreIsCurrentVendor doesn't do it. So it's independent of bpartner but not independent of IsIgnoreIsCurrentVendor

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: its function is included in the update before. PriceLastPO is always defined two updates earlier so it could be assigned to PricePO even if it's not non zero. 
The enhencement of this is so small that I didn't find it worth a seperate ticket. I could remove it, if requested.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Ok, yet another point where I should include brackets. It's intentional to take precedence. Former only if a bpartner was present, existence of a M_Product_PO record must be ensured. Now it must be ensured also if VendorCategory is not null. And of course only the not null part then must be used to decide whether an appropriate M_Product_PO record is present.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
globalqss: Yes - I think it would be good to separate it - bugs are treated different than improvements

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
globalqss: yes - I think is better to remove this - still not convinced the change is harmless because the WHERE there seems wrong
-> WHERE PricePO IS NULL OR (PricePO = 0 AND PriceLastPO <> 0) AND AD_Client_ID=?

is hard to follow OR and AND without proper parenthesis, so, first, that WHERE can be updating records from a different clientID.

I would suggest to drop these two changes to the UPDATEs that are not adding value - but made the peer review more complex (hard to follow what is intended here).

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -240,10 +239,12 @@ private String create() throws Exception {
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR EXISTS "
-						+ "(SELECT * FROM M_Product_PO po "
-						+ "WHERE po.M_Product_ID=p.M_Product_ID AND po.C_BPartner_ID=dl.C_BPartner_ID))"
-						+ " AND (dl.M_Product_ID IS NULL OR p.M_Product_ID=dl.M_Product_ID)");
+						+ " AND ((dl.C_BPartner_ID IS NULL AND dl.VendorCategory IS NULL) OR EXISTS "
```
globalqss, line 242: not convinced about the logic here.
what if a record just defines c_bpartner_id but not vendorcategory?
or the opposite - just vendor category and not business partner?

Maybe this comment from @hengsin makes sense:
-> I’m not sure VendorCategory is necessary when we already have > 1 way to group vendor.

globalqss (changes requested):

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Ok, I'll change that. And you are right: there is a missing bracket which will lead to all PricePO which are NULL will be updated to PriceLastPO even if not in the current client. It wouldn't do harm, but it's wrong.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: If only one of them is defined then only the defined one will select M_Product_PO records, because in the subselect the parts which are NULL are not evaluated.
AND (dl.C_BPartner_ID IS NULL OR po.C_BPartner_ID=dl.C_BPartner_ID)
AND (dl.VendorCategory IS NULL OR po.VendorCategory=dl.VendorCategory)))

mschnbeck: And actually there is no vendor specific grouping other than the vendor itself which at least for my customer isn't enough.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Ok, I'll create a separate ticket and pull request. I'll leave this one unchanged, as it doesn't change anything as soon as the bug is corrected.

mschnbeck: as IDEMPIERE-4469 created and pull request prepared

mschnbeck: Hi @CarlosRuiz-globalqss 
I thought that after committing IDEMPIERE-4469 this difference would disappear. Do I have to regenerate the pull request to get the changes taken into account?

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -218,11 +218,13 @@ private String create() throws Exception {
 						+ " INNER JOIN M_Product p ON (p.M_Product_ID=po.M_Product_ID)"
 						+ " INNER JOIN M_DiscountSchemaLine dl ON (dl.M_DiscountSchemaLine_ID=?) "	//	#2
 						+ "WHERE p.AD_Client_ID IN (?, 0)"		//	#3
-						+ " AND p.IsActive='Y' AND po.IsActive='Y' AND po.IsCurrentVendor='Y'"
+						+ " AND p.IsActive='Y' AND po.IsActive='Y'"
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR po.C_BPartner_ID=dl.C_BPartner_ID)"
```
hengsin, line 225: this should be keep

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -218,11 +218,13 @@ private String create() throws Exception {
 						+ " INNER JOIN M_Product p ON (p.M_Product_ID=po.M_Product_ID)"
 						+ " INNER JOIN M_DiscountSchemaLine dl ON (dl.M_DiscountSchemaLine_ID=?) "	//	#2
 						+ "WHERE p.AD_Client_ID IN (?, 0)"		//	#3
-						+ " AND p.IsActive='Y' AND po.IsActive='Y' AND po.IsCurrentVendor='Y'"
+						+ " AND p.IsActive='Y' AND po.IsActive='Y'"
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR po.C_BPartner_ID=dl.C_BPartner_ID)"
+						+ " AND (dl.VendorCategory IS NULL OR po.VendorCategory=dl.VendorCategory)"
```
hengsin, line 225: dl.VendorCategory should have no effect if dl.C_BPartner_ID > 0

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -218,11 +218,13 @@ private String create() throws Exception {
...
 						+ "WHERE p.AD_Client_ID IN (?, 0)"		//	#3
-						+ " AND p.IsActive='Y' AND po.IsActive='Y' AND po.IsCurrentVendor='Y'"
+						+ " AND p.IsActive='Y' AND po.IsActive='Y'"
 						//	Optional Restrictions
 						+ " AND (dl.Group1 IS NULL OR p.Group1=dl.Group1)"
 						+ " AND (dl.Group2 IS NULL OR p.Group2=dl.Group2)"
-						+ " AND (dl.C_BPartner_ID IS NULL OR po.C_BPartner_ID=dl.C_BPartner_ID)"
+						+ " AND (dl.VendorCategory IS NULL OR po.VendorCategory=dl.VendorCategory)"
+						+ " AND ((dl.C_BPartner_ID IS NULL AND po.IsCurrentVendor='Y')"
+						+ "      OR (po.C_BPartner_ID=dl.C_BPartner_ID AND (dl.IsIgnoreIsCurrentVendor='Y' OR po.IsCurrentVendor='Y')))"
```
hengsin, line 227: I think it should be just AND (dl.IsIgnoreIsCurrentVendor='Y' OR po.IsCurrentVendor='Y')

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
hengsin: I think we should keep the existing condition for "dl.C_BPartner_ID IS NULL OR EXISTS ..." and add another condition for dl.VendorCatory:
AND (dl.C_BPartner_ID IS NULL AND (dl.VendorCategory IS NULL OR EXISTS
(select * from m_product_po po where po.m_product_id=p.m_product_id and po.vendorcategory=dl.vendorcategory)))

As I've posted above, dl.C_BPartner_ID should take precedence over dl.VendorCategory

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
```diff
@@ -295,7 +299,9 @@ else if (M_Pricelist_Version_Base_ID == 0) {
 						+ " INNER JOIN M_DiscountSchemaLine dl ON (dl.M_DiscountSchemaLine_ID=?) "	//	#3
 						//
 						+ "WHERE EXISTS (SELECT * FROM T_Selection s WHERE s.AD_PInstance_ID=? AND po.M_Product_ID=s.T_Selection_ID)"	//	#4
-						+ " AND po.IsCurrentVendor='Y' AND po.IsActive='Y'";
+						+ " AND ((dl.C_BPartner_ID IS NULL AND po.IsCurrentVendor='Y')"
+						+ "      OR (po.C_BPartner_ID=dl.C_BPartner_ID AND (dl.IsIgnoreIsCurrentVendor='Y' OR po.IsCurrentVendor='Y')))"
+						+ " AND po.IsActive='Y'";
```
hengsin, line 304: for consistency, should have added another "AND " clause here for VendorCategory

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Of course there may be situations where all bpartner use the same vendorcategories. But normally you can only trust that a given vendorcategory hits the required products for a given bpartner. Therefore the normal usage would be to select a bpartner together with a vendorcategory.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: But that would be an incompatible change. Currently you can build a pricelist which contains prices only for a given bpartner but only those prices where this bpartner is the currentvendor. I don't know whether this will break someones pricelists, but I think it isn't necessary to break compatibility if it could be easily maintained.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: but it's not necessary, because there may be only one po-line for a given bpartner or one with IsCurrentVendor so because T_Selection contains only lines where VendorCategory matches it will match here anyway.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: If both are defined, both must match. Otherwise we would select product where by chance the vendorcategory is the same but the real product is totally different (vendor A uses category 47 for screws, vendor B uses 47 for shoes). And only if both are null there should be the OR alternative. Otherwise we would find Products which don't match the Vendorcategory.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
hengsin: why is this break the behaviour you mention above ? when dl.IsIgnoreIsCurrentVendor='N', it is checking whether po.IsCurrentVendor='Y' and that should be compatible with existing behaviour.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
hengsin: alright, change it to just 
AND (dl.VendorCategory IS NULL OR EXISTS
(select * from m_product_po po where po.m_product_id=p.m_product_id and po.vendorcategory=dl.vendorcategory))

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: I assume you meant to replace lines 226 and 227 by AND (dl.IsIgnoreIsCurrentVendor='Y' OR po.IsCurrentVendor='Y'). If not, i missunderstood you. Then given a dl.C_BPartner_ID is set products will be included, where there is no po line for the required vendor. Currently they won't be included.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
mschnbeck: Just to be sure to have the correct idea of your suggested change: 
	 AND (dl.C_BPartner_ID IS NULL OR EXISTS 
	         (SELECT * FROM M_Product_PO po 
	            WHERE po.M_Product_ID=p.M_Product_ID AND po.C_BPartner_ID=dl.C_BPartner_ID))
	                AND (dl.M_Product_ID IS NULL OR p.M_Product_ID=dl.M_Product_ID)
         AND (dl.VendorCategory IS NULL OR EXISTS
                   (select * from m_product_po po where po.m_product_id=p.m_product_id 
                        and po.vendorcategory=dl.vendorcategory))

But this would select products where a po line exists with the correct vendor but a wrong vendorcategory and another po line with the correct vendorcategory but a wrong vendor. Of course this will happen very seldom but those errors are worst to find.

org.adempiere.base.process/src/org/compiere/process/M_PriceList_Create.java
hengsin: yes, replace 226 and 227 and keep line 225 above.

mschnbeck: Ok, now I remember why I did it this way. The user interface hides 'IsIgnoreIsCurrentVendor' as long as there is no Vendor given and it is set to 'N' by default. But of course it would be possible to set it to 'Y' even if no vendor is set. So I wanted to ignore it in this case.

But as we are only selecting the products and there is always on po-record with 'IsCurrentVendor' the product would be selected by this record anyway, it doesn't help to ignore others.

I've changed it.

mschnbeck: I just see I had an error in the above. If IsIgnoreIsCurrentVendor is set by accident (and not visible to user) and the VendorCategory is set, then products will be included where IsCurrentVendor is not set for the po-record with the correct VendorCategory. So products are included which are no want to be included. Of course it shouldn't happen that IsIgnoreIsCurrentVendor is set invisible but I personally would prefer to take it right then.

hengsin: With the displaylogic of "@DiscountType@=P & @C_BPartner_ID@!0" for "IsIgnoreIsCurrentVendor", we have to make sure "IsIgnoreIsCurrentVendor" is set to "N" when it is hidden due to the display logic above (perhaps in the beforesave event of the model class). It is easy to have unwanted side effect with a hidden IsIgnoreIsCurrentVendor that's set to 'Y' in db, best to avoid that.

mschnbeck: Ok, to make sure what to do: I have to regenerate X_M_DiscountSchemaLine and I_M_DiscountSchemaLine, add a beforeSave() function in MDiscountSchemaLine and fill IsIgnoreCurrentVendor with 'N' if C_BPartner_ID is 0. Right?

hengsin: yes, that should work. to test - 1) set IsIgnoreCurrentVendor=Y, C_BPartner_ID>0, save.  2) clear C_BPartner_ID to 0, save 3). check db, should have IsIgnoreCurrentVendor=N

mschnbeck: @hengsin, @CarlosRuiz-globalqss, 
I have added and tested setting IsIgnoreIsCurrentVendor to false (or N) if no businesspartner is given. I think, now all amendments are processed.
Regards,
Martin

hengsin:
Hi @mschnbeck , There's a merge conflict that needs to be resolved (i.e git merge master and resolve the merge conflict)

hengsin (approved):
tested

PR #272, opened 2020-09-22
IDEMPIERE-4084 - Reposition default toolbar buttons

d-ruiz:
After talking with @globalqss, this commit repositions some of the buttons that were originally placed on the show more menu. 

After 7 months of using it, we noticed some common buttons need to be brought back to the main toolbar.

This is the final result after it:
image

In summary, the toolbar would look like this:
Find (Because it is next to the search Combobox) - [New - Copy - Save - Delete] (To have first the CRUD operations, this was also mentioned by @hengsin on the ticket) - Undo - Refresh - GridMode  - Attachment - Zoom Across - Rerpot - Process - Print - Parent - Detail - ShowMore.


hengsin:
ok, I did like this new layout better. one thing that bother me is the out of the box look of this is poor - you have 1 big empty drop down box at all your data entry window that did nothing.

globalqss (approved):

d-ruiz:
> ok, I did like this new layout better. one thing that bother me is the out of the box look of this is poor - you have 1 big empty drop down box at all your data entry window that did nothing.

I think we can think about potential solutions for this: 
1. Hide or inactivate that Searchbox if no saved queries exist. But I think this would be not good for UX, as the toolbar would change from window to window.
2. Maybe we could add a default option - if no query exists, we show a ** new query ** option that opens the Find window for a new query.

image

globalqss:
>     1. Hide or inactivate that Searchbox if no saved queries exist. But I think this would be not good for UX, as the toolbar would change from window to window.

This one sounds bad - the memory of toolbar is positional.

>     2. Maybe we could add a default option - if no query exists, we show a ** new query ** option that opens the Find window for a new query.

Alternatively, we could change the placeholder to something different than "Query Name", like:
* if the list is empty: "Click Find button to query"
* if the list have saved searches: "Select Query"

norbertbede:
regarding to saved queries. for example, in salesforce they have defaults, like all records, x days, my records - so dropdown is never empty.

d-ruiz:
Copying the comments to the right ticket for discussion: https://idempiere.atlassian.net/browse/IDEMPIERE-4085. To be able to close the pull request

hengsin:
probably a good idea to always show the ** new query ** option.
I like what norbert suggested above as that make it looks useful from the get go.

PR #278, opened 2020-09-26
IDEMPIERE-4441 More fixes to the SyncDB command

CarlosRuiz-globalqss:
Oracle fix - actual script is ignoring SP2 error messages
Postgresql improvement to show the SQL command that throws error


CarlosRuiz-globalqss:
Pushing this as is a bug causing installations to leave corrupted dictionary.

PR #279, opened 2020-09-30
IDEMPIERE-4468 - Pass the current value of numberbox to the textfield…

pmassing:
… of calculator

Tested 
- in different languages. (english, german)
- with different values
- with different fields 

Without redraw, the content of calculator-textfield 
is alternating empty or filled, when the 
calculator is reopened.


hengsin (approved):
tested

PR #283, opened 2020-10-03
IDEMPIERE-4083 currency rate by document or by transaction

CarlosRuiz-globalqss:
Fix wrong pl/sql functions


CarlosRuiz-globalqss:
Forced merging as is a bug fix for a broken oracle function

PR #284, opened 2020-10-03
IDEMPIERE-3862 ORACLE - Update datatype NVARCHAR2 to VARCHAR2

CarlosRuiz-globalqss:
Fix view to avoid compilation errors


CarlosRuiz-globalqss:
Forced merging - is an unused view with compilation errors in oracle

PR #286, opened 2020-10-04
IDEMPIERE-4475 Can't Run Junit Plugin Test with Eclipse 2020-09

hengsin:
added "org.eclipse.pde.junit.runtime_3.5.700.v20200116-1804"


CarlosRuiz-globalqss (approved):
Tested mvn compilation

PR #291, opened 2020-10-09
Idempiere 4058 - tab editor: new options for editing and customizing fields (window customization)

muriloht:


hengsin (changes requested):
Tested working fine. However, you need to move the migration script from i6.2z to i7.1z and if it is not a problem for you, update the timestamp as well (i.e from 201910081353 to current date and time).

CarlosRuiz-globalqss (changes requested):

CarlosRuiz-globalqss:
replaced by https://github.com/idempiere/idempiere/pull/312

PR #293, opened 2020-10-12
IDEMPIERE-4287 Fix for AD_Table caching.

igorpojzl:
@hengsin 
On Table we had method not using new Immutable Caching.


hengsin (approved):
looks good.

PR #295, opened 2020-10-13
IDEMPIERE-4211 Schedulers launched twice - problem with RunOnlyOnIP n…

hengsin:
…ull running on load balancer

add 1 to 3 minute delay to wait for hazelcast service

### DIFF
--- a/org.adempiere.server/src/main/servlet/org/compiere/web/AdempiereMonitor.java
+++ b/org.adempiere.server/src/main/servlet/org/compiere/web/AdempiereMonitor.java
@@ -1276,6 +1276,20 @@ public class AdempiereMonitor extends HttpServlet
 		WebEnv.initWeb(config);
 		log.info ("");
 		
+		//sleep 1 min to give cluster service time to start first
+		try {
+			Thread.sleep(60000);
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @hengsin - I tested this one and I think it impacts too heavily to most of the implementations which are not clustered in hazelcast.

In my test the whole server starting is stopped waiting for the monitor (so, no homepage, no webui, etc).

I would suggest the following patch on top of your commit:
[pr295_suggest.txt](https://github.com/idempiere/idempiere/files/5370919/pr295_suggest.txt)

It creates an initial wait of 5 seconds - modifiable via SysConfig  MONITOR_INITIAL_WAIT_FOR_CLUSTER_IN_SECONDS - because I think this wait must be higher depending on the environment (hazelcast in amazon for example)

And a max wait of 180 seconds modifiable with SysConfig MONITOR_MAX_WAIT_FOR_CLUSTER_IN_SECONDS

WDYT?

hengsin:
will investigate how to do this without much impart on the other startup activities of the server. will consider your path as well.

PR #302, opened 2020-10-16
IDEMPIERE-4287 Cache API not thread safe and inconsistent with contex…

CarlosRuiz-globalqss:
…t parameter

Fix issue with
java.lang.IllegalStateException: PO is Immutable: org.compiere.model.MRecentItem
	at org.compiere.model.PO.checkImmutable(PO.java:1573)
	at org.compiere.model.PO.set_Value(PO.java:733)
	at org.compiere.model.PO.set_Value(PO.java:721)
	at org.compiere.model.X_AD_RecentItem.setAD_Role_ID(X_AD_RecentItem.java:123)
	at org.compiere.model.MRecentItem.addModifiedField(MRecentItem.java:234)


hengsin (approved):
looks good.

PR #313, opened 2020-10-22
IDEMPIERE-4128 Matched invoice posting with wrong precision based on transaction currency instead of schema currency

etantg:
https://idempiere.atlassian.net/browse/IDEMPIERE-4128


hengsin (approved):
tested

PR #314, opened 2020-10-22
IDEMPIERE-4504:[JR] can't show QR code on jasper report (org/apache/b…

hieplq:
…atik/bridge/UserAgent)

1. DynamicImport don't need anymore, it's done on jasper engine already
2. seem batik-11 don't exchange object org.w3c.dom.* with svg anymore, so don't need to force it (and jasper) to use "org.w3c.dom.events version 3.0.0"
with this change all batik lib use org.w3c.dom.* from jdk so jasper engine can wire to it, so can see org.apache.batik.bridge.UserAgent

change of (2) enough to resolve this issue, next time when we get it back then down org.w3c.dom.smil to 1.0.0 to get rid of "org.w3c.dom.events version 3.0.0" will help on complete that hell


CarlosRuiz-globalqss (approved):
tested successfully

PR #319, opened 2020-10-24
IDEMPIERE-2902

CarlosRuiz-globalqss:
Hi @hengsin - I integrated this code from TrekGlobal into [IDEMPIERE-2902](https://idempiere.atlassian.net/browse/IDEMPIERE-2902).

The [commit 8263265](https://github.com/CarlosRuiz-globalqss/idempiere/commit/8263265) fixes one issue in DisplayType that was making the ColumnSync process to fail when there was a default value without quotes.

Also, I fixed the validation rule and 15 display logics to include other references that were missing.

For testing and UX purposes I configured as Radio Group the fields:
Broadcast Message > Broadcast Message > Target
Field Group > Field group > Field Group Type
Info Window > Process > LayoutType

Now, visually the component looks good and is working fine, but I found some issues, if you can please take a look if is easily solvable:

* The context menu options _Zoom_, _Value Pref_, _Change Log_ are not being set
* Not sure if is worthy and possible to implement the context menu _Requery_
* The context menu _Field Suggestions_ appears twice

I also found a big issue when the component is rendered on a grid first, this is the case:

Broadcast Message > Broadcast Message > Target
Field Group > Field group > Field Group Type
* These two are in the first tab of the window and it's working perfect

Info Window > Process > LayoutType
* This field is in a detail tab and it has some problems depending on how you navigate on the window:
  * When you open the window _Info Window_, then navigate to _Column_ tab (detail), and from there you change to the _Process_ tab, the field works perfect
  * But when you open the window _Info Window_, then double click on the tab _Process_, navigating to the tab, the field behaves badly:
    * The options are Read-Only, cannot be edited
    * If you save, the default on the database will save a _B_, but if you reopen the window and try to edit that field, it cannot be saved as it's setting always the value to null, and the red color of the label never goes away

I tested setting a breakpoint in WRadioGroupEditor.setEnabled:371 and noticed that in the first case (not detail navigation) is creating one read-write editor, but when navigating as the second case, it seems to create several editors, and one of them is read-only.  I suspect maybe this is not related to this specific Editor, but maybe something bigger happening when navigating to a detail that way.

Can you please take a look?

Regards,

Carlos Ruiz

### DIFF
--- a/migration/i7.1z/oracle/201804111700_TicketAP2-19.sql
+++ b/migration/i7.1z/oracle/201804111700_TicketAP2-19.sql
@@ -1,0 +1,27 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Apr 10, 2018 7:56:35 PM GMT+08:00
+-- Radio Group List
+INSERT INTO AD_Reference (AD_Client_ID,AD_Org_ID,AD_Reference_ID,AD_Reference_UU,Created,CreatedBy,EntityType,IsActive,IsOrderByValue,Name,Updated,UpdatedBy,ValidationType) VALUES (0,0,200152,'bde88a42-3601-499e-a2ee-2185eb6fc785',TO_DATE('2018-04-10 19:56:29','YYYY-MM-DD HH24:MI:SS'),100,'D','Y','N','Radio Group List',TO_DATE('2018-04-10 19:56:29','YYYY-MM-DD HH24:MI:SS'),100,'D')
+;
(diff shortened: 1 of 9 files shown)


hengsin:
got it fixed here but in the process, the changes become much bigger than I initially anticipated. needs to review the changes again...

hengsin:
created a new pull #326

CarlosRuiz-globalqss:
replaced

PR #320, opened 2020-10-24
IDEMPIERE-2999 Attribute, Table Direct and date field support

CarlosRuiz-globalqss:
Fix security issue about non advanced roles

https://idempiere.atlassian.net/browse/IDEMPIERE-2999?focusedCommentId=46481&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-46481

### DIFF
--- a/migration/i7.1z/oracle/202010241637_IDEMPIERE-2999.sql
+++ b/migration/i7.1z/oracle/202010241637_IDEMPIERE-2999.sql
@@ -1,0 +1,39 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-2999 Attribute, Table Direct and date field support
+-- Oct 24, 2020, 4:33:46 PM CEST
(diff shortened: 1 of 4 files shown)


dpansheriya (approved):

dpansheriya:
I agree on idea to allow only advanced role to configure reference type attribute.

hengsin:
This looks fine with one key difference from the other advance role configurations - the beforeSave checking make it not possible for implementation that wants to open up more through configuration only.

CarlosRuiz-globalqss:
> This looks fine with one key difference from the other advance role configurations - the beforeSave checking make it not possible for implementation that wants to open up more through configuration only.

Yes, that's exactly the intention, it is configured in the dictionary, but also I added the beforeSave because I want to stop the ability for a non-advanced role to be able to configure that by other means (webservices, import, etc).

There are some other cases where we have done the same in before*, like MWFNode and MUserRoles

hengsin:
Hi @CarlosRuiz-globalqss ,

Yes, your intention here is pretty obvious. I'm just pointing out the lack of consistency here.

Regards,
Low

CarlosRuiz-globalqss:
Thanks @hengsin - so, is it OK to push this?  Or you suggest something different?

hengsin:
Yeah, I'm fine with this. Just curious why the sudden change in approach - there's no beforeSave checking in other advanced setting. i.e in other existings advanced stuffs, the implementation can choose to make it not advance by a simple change of flag.

PR #322, opened 2020-10-24
IDEMPIERE-4499:zk9.5 - Popup menu disappear quickly for buttons on th…

hieplq:


hengsin:
I tested this but I'm not sure what have been improved here - it is still a poor experience where the submenu for "Private Record Lock" popup after the menu popup have been closed. It will be good if it can be make behave the same as the overflow popup for mobile where the main popup menu stay open.

hieplq:
it's not improve, it's work-around for zk issue
agree moment design isn't great, it should is sub-menu as Norbert do

CarlosRuiz-globalqss:
Tested successfully too - agree would be better if the behavior is similar to mobile, if is too hard to implement, then the workaround is OK.

CarlosRuiz-globalqss (approved):

hengsin:
lets hold this open first, I will try to build an alternative that uses the same mobile implementation.

hengsin:
created the alternative implementation that uses the existing mobile overflow popup #329 , please compare and pick one

CarlosRuiz-globalqss:
replaced by https://github.com/idempiere/idempiere/pull/329

PR #323, opened 2020-10-26
IDEMPIERE-4509 Grid View Horizontal scrolling is not working well

hengsin:
Disable the use of Frozen


CarlosRuiz-globalqss (approved):

CarlosRuiz-globalqss:
Hi @hengsin - tested successfully to solve the reported issue.

Is a shame to lose the functionality of freezing the selection column, but if the solution is too hard then I think is a good compromise.

Regards,

Carlos Ruiz

hengsin:
It will be quite an effort to reverse engineer the zk ee fix (I've some initial findings of that post in the ticket, if someone have time to work on that, can take that as the start) and it looks like zk is not going to fix the broken zk ce implementation.

PR #331, opened 2020-10-29
Idempiere-4492 update library (eclipse-4.17, jasper 6-15, jetty-9.4.33, atmosphere-2.6.4) 

hieplq:


CarlosRuiz-globalqss:
Hi @hieplq - can you please add also the updates for JDBC in the attached patch:
[pr_331_jdbc.txt](https://github.com/idempiere/idempiere/files/5458532/pr_331_jdbc.txt)

I did basic tests with postgresql and oracle and both seems to be working fine.

BTW, the oracle JDBC is now in maven, so, no need to do extra.jar

Regards,

Carlos Ruiz

CarlosRuiz-globalqss:
Hi @hengsin - I tested this successfully - it was required a complete install from eclipse (delete the .metadata and reconfigure).

I tested jasper reports, copy client, initial client, and a POS sales order, working fine.

You made this comment:
> since the jetty project doesn't test jetty 9 against the jakarta bundles, I know it will build and run
> but I'm concern whether it is safe/reliable to use jetty 9.x with those jakarta bundles

So, please check if we must wait for that concern to be solved or is OK to push like this.

Regards,

Carlos Ruiz

hengsin (approved):
looks good, lets get it out.

hengsin:
Hi @CarlosRuiz-globalqss , it looks fine, only annotation jar is move to jakarta, not the servlet bundles.

PR #345, opened 2020-11-02
IDEMPIERE-4211 Schedulers launched twice - problem with RunOnlyOnIP n…

hengsin:
…ull running on load balancer

- log errors from background init.
- add explicit Env Context init for AdempiereMonitor thread and
Background init thread.


CarlosRuiz-globalqss (approved):
Tested successfully

PR #346, opened 2020-11-02
IDEMPIERE-3340 Fixes for oracle migration scripts

CarlosRuiz-globalqss:


CarlosRuiz-globalqss:
Pushed with admin rights as is a wrong script failing in oracle

PR #349, opened 2020-11-03
IDEMPIERE-4515 Mobile: Paging control at detail pane overlapped with …

hengsin:
…toolbar

- Use more compact layout for small width screen
- Always enable click to popup for status message for small width screen


CarlosRuiz-globalqss (approved):
looks fine

PR #350, opened 2020-11-03
IDEMPIERE-4519 Zk: Implement optional defer rendering for Grid View

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4519


CarlosRuiz-globalqss (changes requested):
comments in ticket

PR #356, opened 2020-11-05
IDEMPIERE-4525 Translate MailText in Print Invoices to the correspond…

d-ruiz:
…ing BP Language

https://idempiere.atlassian.net/browse/IDEMPIERE-4525


CarlosRuiz-globalqss:
Hi @d-ruiz - it looks good - can you please do the same in DunningPrint (also 2 classes).

d-ruiz:
DunningPrint does not have this issue because it sets the Business Partner before any call to the text, therefore, it doesn't need any change. DunningPrint line 194 sets the BPartner

CarlosRuiz-globalqss (approved):
looks good

PR #358, opened 2020-11-06
IDEMPIERE-4514 Grid View Layout Bug

hengsin:
Fix layout issue after drag and drop of tab


hengsin:
Note that when the dragged panel is in grid view mode, it takes a big of delay before it fix the sizing of the grid.

CarlosRuiz-globalqss:
Hi @hengsin - tested successfully moving the tab to the left, but now I noticed moving the tab to the right has the same issue.

CarlosRuiz-globalqss (approved):

PR #361, opened 2020-11-08
IDEMPIERE-4083 currency rate by document or by transaction

CarlosRuiz-globalqss:
Fix NPE when the field is not present in the window
13:56:08.612===========> CalloutPayment.start: start: amounts [147]
java.lang.NullPointerException
        at org.compiere.model.CalloutPayment.amounts(CalloutPayment.java:599)


hengsin (approved):
trivial NPE fix, looks fine.

PR #363, opened 2020-11-09
IDEMPIERE-2490 java.lang.IllegalStateException: Form too large

hengsin:
replace the now defunc setting at jetty.xml with jetty-web.xml


CarlosRuiz-globalqss (approved):
tested successfully with user picture capture

PR #373, opened 2020-11-10
IDEMPIERE-4538 Dropdown button is slightly shifted on FindWindow and …

d-ruiz:
…Payment Rule


hengsin (approved):
trivia fix, good to go.

PR #374, opened 2020-11-11
IDEMPIERE-4540 Translate generated mail from Send Mail Text according…

d-ruiz:
… to the Business Partner Language

https://idempiere.atlassian.net/browse/IDEMPIERE-4540


hengsin (approved):
trivial fix, good to go.

PR #376, opened 2020-11-12
IDEMPIERE-4542 Cannot report from window _My Unprocessed Documents_

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4542


hengsin (approved):
tested

PR #378, opened 2020-11-12
IDEMPIERE-4545 Workflow activities - OK button cannot be reached in m…

CarlosRuiz-globalqss:
…obile

https://idempiere.atlassian.net/browse/IDEMPIERE-4545


hengsin (approved):
trivial fix, good to go.

PR #380, opened 2020-11-12
IDEMPIERE-4549 Detect broken theme configuration

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4549


hengsin (changes requested):
with the use of static variables, you would have to synchronize the static getTheme() method.

CarlosRuiz-globalqss:
Thanks @hengsin - I added synchronized to the method - do you think the approach is OK?   I tried to make it in a way to impact the performance the less possible, still allowing to change the theme in a hot way (cache reset, or setprop), but I'm not sure if the approach is OK.

Also, can you please review the patch in https://idempiere.atlassian.net/browse/IDEMPIERE-4536?focusedCommentId=46582&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-46582
Maybe I'm missing something minor there - or maybe is simply not possible.

hengsin:
Hi Carlos, now I think the original version without synchronize is better. This method is call quite a lot and I think it is better to do away withe the synchronization performance penalty since the risk involved is very low.

CarlosRuiz-globalqss:
ready - reverted

hieplq:
@CarlosRuiz-globalqss why not check on save of MConfig?
or we can create a process run schedule on test server only (maybe each 10 minute), this process do
1. checking theme setting and reset it
2. checking password change and reset it
3. login by english is disabled

CarlosRuiz-globalqss:
thanks @hengsin - the test servers are not a big problem - when I detect that situation, I simply trigger the jenkins job to recreate the database - there are many ways people mess up the server with just one click (for example enabling SaaS on client).

In reality is not something that I'm too concerned - is just that I was playing with IDEMPIERE-4536 yesterday - and saw the opportunity to do this - but if is something expensive, I think we can simply leave it as it is - the happening is unlikely on a real production server.

hengsin (approved):
good to go

PR #382, opened 2020-11-14
IDEMPIERE-4536 Allow minor customizations in default theme

hengsin:
only include fragment/custom.css.dsp if the file exists


CarlosRuiz-globalqss (approved):
Tested successfully

PR #383, opened 2020-11-16
IDEMPIERE-2418 Record Access Dialog Layout issue

CarlosRuiz-globalqss:
Fix layout for desktop and mobile
Filter the role list to exclude System roles

https://idempiere.atlassian.net/browse/IDEMPIERE-2418


hengsin (approved):
looks good

PR #386, opened 2020-11-17
IDEMPIERE-4512 zk9.5 - Timeout page is not working

hengsin:
- Fix a javascript error when session timeout
- Add support for session timeout alert


CarlosRuiz-globalqss (approved):
tested successfully

PR #389, opened 2020-11-18
IDEMPIERE-4555 Incorrect price in intercompany document

AndreaCh83:
**Incorrect price in intercompany document**

Description
I post a question in google group:

https://groups.google.com/g/idempiere/c/taAfzmkSKgk/m/WpMNgt7oBAAJ

Andrea Bellotto

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MOrder.java
+++ b/org.adempiere.base/src/org/compiere/model/MOrder.java
@@ -2388,7 +2388,6 @@ public class MOrder extends X_C_Order implements DocAction
 		{
 			MOrderLine counterLine = counterLines[i];
 			counterLine.setOrder(counter);	//	copies header values (BP, etc.)
-			//counterLine.setPrice();
 			counterLine.setPrice(counterLine.getPriceEntered());
 			counterLine.setTax();
 			counterLine.saveEx(get_TrxName());
--- a/org.adempiere.base/src/org/compiere/model/MOrder.java
+++ b/org.adempiere.base/src/org/compiere/model/MOrder.java
@@ -2388,7 +2388,6 @@ public class MOrder extends X_C_Order implements DocAction
 		{
 			MOrderLine counterLine = counterLines[i];
 			counterLine.setOrder(counter);	//	copies header values (BP, etc.)
-			counterLine.setPrice(counterLine.getPriceEntered());
 			counterLine.setTax();
 			counterLine.saveEx(get_TrxName());
 		}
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/MOrder.java
```diff
@@ -2388,7 +2388,8 @@ protected MOrder createCounterDoc()
 		{
 			MOrderLine counterLine = counterLines[i];
 			counterLine.setOrder(counter);	//	copies header values (BP, etc.)
-			counterLine.setPrice();
+			//counterLine.setPrice();
```
hengsin, line 2391: If we are sure that this is not needed, we should remove instead of just comments it out.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MOrder.java
```diff
@@ -2388,7 +2388,7 @@ protected MOrder createCounterDoc()
 		{
 			MOrderLine counterLine = counterLines[i];
 			counterLine.setOrder(counter);	//	copies header values (BP, etc.)
-			counterLine.setPrice();
+			counterLine.setPrice(counterLine.getPriceEntered());
```
hengsin, line 2391: This is unnecessary (if priceentered have been copied over from source line, priceactual should have been copied over as well). If the concern is not all price related value have been copied over from source line (perhaps due to changes to ad_column's allow copy flag), changes should be added to copyLinesFrom instead, i.e if counter==true, make sure all price related value from source line have been copied over to counter line.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MOrder.java
idempiere-consulting: I was a proposal of Nicolas because you can have "allow copy" set to NO .

idempiere-consulting: @nmicoud    discussion

hengsin: yeah but as I mention above, that's insufficient for the "allow copy" set to no concern since you can have that set to no for price entered as well.

nmicoud: hello
nb: I did that change for a particular customer - not search for a nice/generic solution.
I would prefer hengsin's approach

CarlosRuiz-globalqss: IMO is simply just to drop that line - the price is already set above.

hengsin (approved):
approve to keep it simple for now. there's a broader issue that developer/implementer can make changes to ad_column.isallowcopy flag that can break some processes (for e.g, if you set c_orderline.m_product_id isallowcopy to false, this will break) that deserve further investigation in a separate ticket.

PR #390, opened 2020-11-18
IDEMPIERE-4559 Customize and Process detail toolbar buttons are not c…

d-ruiz:
…lickable after creating a new record in the header

https://idempiere.atlassian.net/browse/IDEMPIERE-4559


hengsin (approved):
looks good.

PR #393, opened 2020-11-20
IDEMPIERE-4554 PO is immutable - Workflow editor

CarlosRuiz-globalqss:
- fix the PO is immutable errors
- forbid that system nodes are changed by non-system users
- system nodes are shown in gray
- reorganize layout of system workflows in editor
- make the nodes smaller
- fix the layout problem of name and properties dialogs

https://idempiere.atlassian.net/browse/IDEMPIERE-4554


hengsin (approved):
tested

PR #395, opened 2020-11-20
IDEMPIERE-4557 Order document by DateTrx and Created on Auto Allocati…

CarlosRuiz-globalqss:
…on process

- change the order by for payments from "DateTrx" to "DateTrx, Created, C_Payment_ID"
- change the order by for invoices from "DateInvoiced" to "DateInvoiced, Created, C_Invoice_ID"
- refactor four JDBC managed queries to use the Query function

https://idempiere.atlassian.net/browse/IDEMPIERE-4557


hengsin (approved):
tested

PR #397, opened 2020-11-20
IDEMPIERE-4560 Autocomplete fills the value to be suggested to the us…

CarlosRuiz-globalqss:
…er with the cache and does not support multilanguage

Fixes the problem about caching results from previously used language, breaking the usage of base language when another language was used previously

https://idempiere.atlassian.net/browse/IDEMPIERE-4560


CarlosRuiz-globalqss:
same solution as  IDEMPIERE-4427

hengsin (approved):
looks good.

PR #399, opened 2020-11-21
IDEMPIERE-4558 Copy GardenWorld into a new Client using Copy Client -…

CarlosRuiz-globalqss:
… failing in postgresql native

Fix the problem in postgresql native

https://idempiere.atlassian.net/browse/IDEMPIERE-4558


hengsin (approved):
trivial fix, looks good.

PR #400, opened 2020-11-21
IDEMPIERE-4465 Native Postgres - Error on Convert Layer

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4465


CarlosRuiz-globalqss:
Also fixes that the current code was not logging correctly when the property org.idempiere.db.debug was set.

hengsin (approved):
looks good.

PR #402, opened 2020-11-21
IDEMPIERE-4572 Fix NPE

ocurieles:
Show message account info


hengsin (approved):
trivial fix, good to go.

PR #403, opened 2020-11-22
IDEMPIERE-4520: Attribute Instance not Refreshed on Create

dpansheriya:
This is bug fix reported by Norbert.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAttribute.java
+++ b/org.adempiere.base/src/org/compiere/model/MAttribute.java
@@ -280,7 +280,12 @@ public class MAttribute extends X_M_Attribute implements ImmutablePOSupport
 	{
 		MAttributeInstance instance = getMAttributeInstance(M_AttributeSetInstance_ID);
 		if (instance == null)
+		{
 			instance = new MAttributeInstance(getCtx(), getM_Attribute_ID(), M_AttributeSetInstance_ID, valueInt, get_TrxName());
+			if (value == null)
+				value = String.valueOf(valueInt);
+			instance.setValue(value);
+		}
 		else
 			instance.setValueInt(valueInt, value);
 		instance.saveEx();


org.adempiere.base/src/org/compiere/model/MAttribute.java
```diff
@@ -280,7 +280,12 @@ public void setMAttributeInstance(int M_AttributeSetInstance_ID, int valueInt, S
 	{
 		MAttributeInstance instance = getMAttributeInstance(M_AttributeSetInstance_ID);
 		if (instance == null)
+		{
```
hengsin, line 283: Going through the history, this originated from the changes for IDEMPIERE-2999. While the code works fine (including this fix), it is not tidy design wise. 1. You really want to take a KeyNamePair, not int valueInt + String value parameter that's hard to understand. 2. MAttributeInstance have getValueInt and setValueInt pair that doesn't follows common Java Bean rule and kind of confusing. 3. MAttributeInstance should have a constructor that take KeyNamePair (or valueInt + value for the current implementation). The lack of that constructor make it inconsistent with other part of the code.

CarlosRuiz-globalqss:
Replaced with https://github.com/idempiere/idempiere/pull/409

I reopened the ticket [IDEMPIERE-2999](https://idempiere.atlassian.net/browse/IDEMPIERE-2999) and added the comments from @hengsin 

@dpansheriya can you please check the recommendations there?

dpansheriya:
> Replaced with #409
> 
> I reopened the ticket [IDEMPIERE-2999](https://idempiere.atlassian.net/browse/IDEMPIERE-2999) and added the comments from @hengsin
> 
> @dpansheriya can you please check the recommendations there?

Should this suggestions committed about IDEMPIERE-2999?

CarlosRuiz-globalqss:
Yes, I think is better as is part of that ticket

org.adempiere.base/src/org/compiere/model/MAttribute.java
dpansheriya: 1. We follow syntas matching to public void setMAttributeInstance (int M_AttributeSetInstance_ID, MAttributeValue value) which was existing method. 
Do you here suggest we should replace all setMAttributeInstance by single one accepting KeyNamePair?
2. We are using same column named ValueNumber for storing Numeric, Integer or Foreignkey reference to reduce number of columns. This cause as to create method for Int but there is no column of Integer presents. Also we are passing string value along with Integer value for storing readable attribute descriptions to the attribute instance for Table Direct or Table type references.
3. Yes, I think we should do this. Will take care for this.

PR #407, opened 2020-11-24
IDEMPIERE-3340 : Tree structurable user favorite

dpansheriya:
Please note that this patch has revised as per discussion with Carlos and hengsin. Following item is not implemented yet and I believe that without same we may be good, So I want to differ till real user fedback.

"If Folder is empty, don’t show it. if user try to create Folder  with same name under same parents, then instead of creating new one, show existing one."


CarlosRuiz-globalqss:
Hi @dpansheriya - can you please resolve conflicts on this pull request?

dpansheriya:
@CarlosRuiz-globalqss Sure, I will work on same.

PR #408, opened 2020-11-24
IDEMPIERE-4566 Translation Import/Export - output message is not visible

CarlosRuiz-globalqss:
- Deprecate zk form WTranslationDialog and controller TranslationController
- Move the logic to a java process org.idempiere.process.TranslationImpExp
- As a process works in mobile, can be called via webservices, tested with fitnesse, etc
- Importing leaves a better and readable log of each file processed
- Importing/Exporting gives feedback to the user about the file being processed

Side note:
- To standardize the menu - renamed the "Initial Client Setup Process" to "Initial Client Setup"

Improvements over the previous form:
- can process the files from the zip in root folder, or in internal folder with the name of the language
- can process a URL http/https to download the zip file

https://idempiere.atlassian.net/browse/IDEMPIERE-4566

### DIFF
--- a/migration/i8.1z/oracle/202011241451_IDEMPIERE-4566.sql
+++ b/migration/i8.1z/oracle/202011241451_IDEMPIERE-4566.sql
@@ -1,0 +1,387 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-4566 Translation Import/Export - output message is not visible
+-- Nov 24, 2020, 2:45:45 PM CET
+UPDATE AD_Form SET Name='Initial Client Setup Form (deprecated)',Updated=TO_DATE('2020-11-24 14:45:45','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Form_ID=102
+;
+
+-- Nov 24, 2020, 2:45:45 PM CET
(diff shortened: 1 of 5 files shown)


migration/i8.1z/oracle/202011241451_IDEMPIERE-4566.sql
```diff
@@ -0,0 +1,387 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-4566 Translation Import/Export - output message is not visible
+-- Nov 24, 2020, 2:45:45 PM CET
+UPDATE AD_Form SET Name='Initial Client Setup Form (deprecated)',Updated=TO_DATE('2020-11-24 14:45:45','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Form_ID=102
```
hengsin, line 6: shouldn't the form is mark inactive as well ?

hengsin (changes requested):

migration/i8.1z/oracle/202011241451_IDEMPIERE-4566.sql
CarlosRuiz-globalqss: Hi @hengsin - the form is already inactive, I was just changing the name to make explicit that is not used anymore.

hengsin (approved):
good to go.

PR #412, opened 2020-11-25
IDEMPIERE-1710 Receipt create lines from not update when change locat…

CarlosRuiz-globalqss:
…or or purchase order, invoice

https://idempiere.atlassian.net/browse/IDEMPIERE-1710


hengsin (approved):
The UI is still confusing but it is better than what we have now.

PR #417, opened 2020-11-27
IDEMPIERE-4394 Date Field font color is not consistent with other rea…

d-ruiz:
…donly fields

https://idempiere.atlassian.net/browse/IDEMPIERE-4394


CarlosRuiz-globalqss (approved):
tested successfully

PR #420, opened 2020-11-29
IDEMPIERE-4421 Html asset versioning should allow fluent deployment

hengsin:
use classpath loading and lang-addon versioning for theme resources


CarlosRuiz-globalqss:
Hi @hengsin - please check the patch [IDEMPIERE-4421_backward_compatible_0f0ef9b7.txt](https://github.com/idempiere/idempiere/files/5612500/IDEMPIERE-4421_backward_compatible_0f0ef9b7.txt)

I made some tests and it preserves backward compatibility with old themes, except if the theme is using pointers to /theme/default - but that's easily fixed pointing correctly to the own theme - or changing the reference to ~./theme/default

Regards,

Carlos Ruiz

CarlosRuiz-globalqss (approved):

PR #424, opened 2020-11-30
IDEMPIERE-4268 Fix recursive call to PO

CarlosRuiz-globalqss:
Found a case where PO.checkValidClient fails, and then because of log.severe it tries to create a record with MIssue, and the PO.checkValidClient fails in MIssue, and is called recursively until it fails with a "Trx.run: Transaction timeout."

https://idempiere.atlassian.net/browse/IDEMPIERE-4268

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRegion.java
+++ b/org.adempiere.base/src/org/compiere/model/MRegion.java
@@ -16,27 +16,25 @@
  *****************************************************************************/
 package org.compiere.model;
 
+import static org.compiere.model.SystemIDs.COUNTRY_JAPAN;
+
 import java.io.Serializable;
 import java.sql.ResultSet;
-import java.sql.SQLException;
-import java.sql.Statement;
 import java.text.Collator;
 import java.util.Arrays;
 import java.util.Comparator;
+import java.util.List;
 import java.util.Properties;
 import java.util.logging.Level;
 
 import org.compiere.Adempiere;
 import org.compiere.util.CLogger;
-import org.compiere.util.DB;
 import org.compiere.util.Env;
-import org.idempiere.cache.ImmutablePOSupport;
 import org.idempiere.cache.ImmutablePOCache;
-
-import static org.compiere.model.SystemIDs.*;
+import org.idempiere.cache.ImmutablePOSupport;
 
 /**
- *	Localtion Region Model (Value Object)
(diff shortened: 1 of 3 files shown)


d-ruiz (approved):

org.adempiere.base/src/org/compiere/model/MRegion.java
```diff
@@ -47,39 +45,34 @@
...
-				r.markImmutable();
-				s_regions.put(String.valueOf(r.getC_Region_ID()), r);
-				if (r.isDefault())
-					s_default = r;
+		List<MRegion> regions;
+		int cid = Env.getAD_Client_ID(Env.getCtx());
+		try {
+			if (cid > 0) {
+				// forced potential cross tenant read - requires System client in context
+				Env.setContext(Env.getCtx(), Env.AD_CLIENT_ID, 0);
```
hengsin, line 61: I don't understand, why is this needed here ?

org.adempiere.base/src/org/compiere/wf/MWorkflow.java
```diff
@@ -130,11 +130,23 @@ public static MWorkflow getCopy(Properties ctx, int AD_Workflow_ID, String trxNa
...
-				.setParameters(new Object[]{WORKFLOWTYPE_DocumentValue, true})
-				.setOnlyActiveRecords(true)
-				.setOrderBy("AD_Client_ID, AD_Table_ID")
-				.list();
+			List<MWorkflow> workflows;
+			int cid = Env.getAD_Client_ID(Env.getCtx());
+			try {
+				if (cid > 0) {
+					// forced potential cross tenant read - requires System client in context
+					Env.setContext(Env.getCtx(), Env.AD_CLIENT_ID, 0);
```
hengsin, line 138: same as above, why do we have to do this here ?

CarlosRuiz-globalqss:
Hi @hengsin - MWorkflow has a mechanism for performance to read all records (from all tenants) and put them in CCache.

That comes since compiere times - if we would like to change that to per-client cache that requires some refactoring that is not straight (and probably will affect performance).

About MRegion, it has the same mechanism to read all regions in cache - in theory regions cannot be added per client, but as it is a cross-tenant query, I preferred to go on the safe side in case in some implementations they decide to add regions in tenant side.

Regards,

Carlos Ruiz

PR #425, opened 2020-11-30

CarlosRuiz-globalqss:
Implement cross tenant check on foreign keys for insert and update

https://idempiere.atlassian.net/browse/IDEMPIERE-4268


hengsin:
this looks fine but at the same time, I'm curious how big is the performance hit it is. also, what would be the performance difference if it is implemented at db level ?

hengsin (approved):
looks fine.

PR #427, opened 2020-12-01
IDEMPIERE-2999 : refactoring as per Hengsin's review comments

dpansheriya:


org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
```diff
@@ -108,32 +120,31 @@ public MAttributeInstance (Properties ctx, int M_Attribute_ID,
...
 			String trxName)
 	{
-		super(ctx, 0, trxName);
-		setM_Attribute_ID(m_Attribute_ID);
-		setM_AttributeSetInstance_ID(m_AttributeSetInstance_ID);
+		this(ctx, M_Attribute_ID, M_AttributeSetInstance_ID, trxName);
 		setValueDate(value);
 	} // MAttributeInstance
 
+	public MAttributeInstance(Properties ctx, int M_Attribute_ID, int M_AttributeSetInstance_ID, KeyNamePair value, String trxName)
```
hengsin, line 142: good to have basic javadoc here

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
```diff
@@ -213,11 +224,10 @@ else if (displayType == DisplayType.Time)
 	 * @param valueInt
 	 * @param value
 	 */
-	public void setValueInt(int valueInt, String value)
+	public void setValueInt(int valueInt)
 	{
 		super.setValueNumber(BigDecimal.valueOf(valueInt));
-		if (value == null)
-			value = String.valueOf(valueInt);
+		String	value = String.valueOf(valueInt);
```
hengsin, line 230: just curious, what's the use case here to convert valueInt to string and pass it to the string value field ?

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
```diff
@@ -234,6 +244,24 @@ public int getValueInt()
...
 	} // getValueInt
 
+	
+	/**
+	 * Set Value as keyNamePair
+	 * @param value
+	 */
+	public void setValueKeyNamePair(KeyNamePair value)
+	{
+		if (value.getName() == null)
```
hengsin, line 254: don't think the if is necessary here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
```diff
@@ -1185,7 +1185,7 @@ else if (displayType == DisplayType.Image
 				valueLable = editor.getDisplay();
```
hengsin, line 1185: looks like there's an existing typo here (valueLable instead of valueLabel), would be nice to fix it as well.

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
dpansheriya: @hengsin we are converting Int to String as Value column is of String type.

hengsin: yeah, I can read the code. I just don't get it why do we need to do that here ? what use case does it serve here ?

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
dpansheriya: @hengsin there may be some cases when display label is resolved to null, in that case it is good idea to default to _ID  value.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
dpansheriya: Yes, good case. Will fix.

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
dpansheriya: Sure will do.

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
hengsin: nvm, that seems what other set method is doing as well.

org.adempiere.base/src/org/compiere/model/MAttributeInstance.java
hengsin: that usually means there's issue to be fixed somewhere else and I don't think showing ID is useful for user (In a way it can mask the fact the ID doesn't resolve properly).

dpansheriya: Yes, we can go with assumption that name never null. this if block may be never traversed part. but here it was just safe exception handling.

dpansheriya:
@hengsin Please note that I pushed changes you suggested

hengsin (approved):
good to go

PR #429, opened 2020-12-02
IDEMPIERE-4579 Function altercolumn fails when a role name has a dash

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4579


CarlosRuiz-globalqss:
Tested successfully in TrekGlobal

PR #433, opened 2020-12-03
IDEMPIERE-4580 Refactor about credits

CarlosRuiz-globalqss:
Work in progress

The file Credits.html is just a template until defined

https://idempiere.atlassian.net/browse/IDEMPIERE-4580


hengsin (approved):
good to go

PR #434, opened 2020-12-04
IDEMPIERE-4581 Model generator should isn't dialog it should be window

hieplq:
use shutdown asyn to shutdown framework when close Model generator window

one unhappy thing remain:
when use -console then framework is shutdown but jdk still hang because "pipe-gosh --login --noshutdown" still live (for https://bugs.eclipse.org/bugs/show_bug.cgi?id=362412)
work-around: use -console 1234
by that way we can't access osgi command from console but can use "telnet localhost 1234" to use osgi command

### DIFF
--- a/org.adempiere.base-feature/model.generator.launch
+++ b/org.adempiere.base-feature/model.generator.launch
@@ -21,9 +21,9 @@
         <listEntry value="org.eclipse.debug.ui.launchGroup.debug"/>
     </listAttribute>
     <stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
-    <stringAttribute key="org.eclipse.jdt.launching.PROGRAM_ARGUMENTS" value="-os ${target.os} -ws ${target.ws} -arch ${target.arch} -nl ${target.nl} -consoleLog -console"/>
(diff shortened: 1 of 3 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @hieplq - I tested this:

- the positive effect is that the window is now seen by the operating system and I can switch to that window with Alt-Tab

- the negative effect is that the program doesn't end when I click on Cancel or close the window (note that the swing client can be closed, so maybe there is a mechanism there)

hieplq:
you need to wait for while for thread of timer at PackInApplicationActivator finish

on practice it isn't issue because time for user input to window will cover it

actual i see it's issue when packin folder run by model generator it should run only by main product (server application)

anyway can use replace 
Timer t = new Timer();
by
Timer t = new Timer(true);
to switch timer to deamon process to don't block exit jdk

CarlosRuiz-globalqss:
Hi @hieplq - I tested disabling pipo from the launcher and it worked fine.

Can you please check the attached patch (and integrate if you find it OK):
[launcher.txt](https://github.com/idempiere/idempiere/files/5647315/launcher.txt)

I disabled a lot of plugins from the launcher and still is running perfect.

Regards,

Carlos Ruiz

CarlosRuiz-globalqss:
Also, if you want to consider it - a minor improvement on the layout of the window:
[ModelGeneratorGBC.txt](https://github.com/idempiere/idempiere/files/5647378/ModelGeneratorGBC.txt)

CarlosRuiz-globalqss:
replaced with https://github.com/idempiere/idempiere/pull/463

PR #435, opened 2020-12-04
IDEMPIERE-4101 [Chosenbox] Multi select don't work in Firefox

hengsin:
- add display of button
- fix dropdown doesn't render properly at Firefox


CarlosRuiz-globalqss (approved):
Tested successfully in firefox 83.0 - ubuntu 20.04

PR #437, opened 2020-12-04
IDEMPIERE-4387 Cannot select C&W Business Partner in Multi Select Search

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4387

Hi @hengsin, can you please check this one?

I took a simple open source class to do the conversion from HTML to plain from:
http://gmailassistant.sourceforge.net/src/org/freeshell/zs/common/HtmlManipulator.java.html

But maybe is more recommended to use org.apache.commons.lang.StringEscapeUtils.unescapeHtml - however we don't have org.apache.commons.lang available in the target platform.

Maybe there is another class available that can do the same?

Regards,

Carlos Ruiz


hieplq:
@CarlosRuiz-globalqss we have org.apache.commons.lang3 on target

hengsin:
replace with pull 438 (base on this patch + hints from hieplq)

PR #438, opened 2020-12-05
IDEMPIERE-4387 Cannot select C&W Business Partner in Multi Select Search

hengsin:
- base on patch from Carlos and hints from hieplq


CarlosRuiz-globalqss (approved):
tested successfully

PR #447, opened 2020-12-08
IDEMPIERE-4289 Implement Support for JUnit5 Unit Testing

CarlosRuiz-globalqss:
Fix launcher to avoid adding all custom plugins every time I try to execute it


hengsin (approved):
trivial fix, good to go

PR #448, opened 2020-12-08
IDEMPIERE-4358 Processes to drop some objects from database

CarlosRuiz-globalqss:
Add process to rename a column directly from the Table and Column window

https://idempiere.atlassian.net/browse/IDEMPIERE-4358


d-ruiz (changes requested):
Comments added to the ticket

d-ruiz (approved):

PR #449, opened 2020-12-09
IDEMPIERE-4590 Scheduler Window: Start Scheduler action change AD_Cli…

hengsin:
…ent_ID to 0 when HazelCast Service is Down/Disable

https://idempiere.atlassian.net/browse/IDEMPIERE-4590


CarlosRuiz-globalqss (approved):
Tested successfully - scheduled a report and it worked without problems sending the PDF to notice

PR #451, opened 2020-12-09
IDEMPIERE-4421 Html asset versioning should allow fluent deployment

hengsin:
Fix missing downarrow for breadcrumb.


CarlosRuiz-globalqss (approved):
tested successfully

PR #453, opened 2020-12-10

hengsin:
- Use a different approach that doesn't add strange API to PO. The new
API reduce chance of mistake and give clearer picture of what's the
intention. Also, it sounds more correct to perform the cross tenant
operation with the active context's AD_Client_ID=0
- rename checkValidContext(boolean) to checkCrossTenant(boolean)

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MClient.java
+++ b/org.adempiere.base/src/org/compiere/model/MClient.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Properties;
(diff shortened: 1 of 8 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @hengsin - analyzed the changes I would prefer to keep it as it is now, and just commit the changes:
- the method name change to PO.checkCrossTenant
- the synchronized m_fullListLock that you implemented in MUserDefWin

I tested this change with the swing client and with the new approach is still broken - probably another different bug in swing, but that could mean that this proposed change can potentially break more stuff out there than the actual simpler approach.

hengsin:
drop this.

PR #459, opened 2020-12-10
IDEMPIERE-4596 Cannot create a product and an inventory document in t…

CarlosRuiz-globalqss:
…he same transaction

WIP - commit failing unit test

https://idempiere.atlassian.net/browse/IDEMPIERE-4596

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MInventory.java
+++ b/org.adempiere.base/src/org/compiere/model/MInventory.java
@@ -384,7 +384,7 @@ public class MInventory extends X_M_Inventory implements DocAction
 			//	Product requires ASI
 			if (line.getM_AttributeSetInstance_ID() == 0)
 			{
-				MProduct product = MProduct.get(getCtx(), line.getM_Product_ID());
+				MProduct product = MProduct.get(getCtx(), line.getM_Product_ID(), get_TrxName());
 				if (product != null && product.isASIMandatory(line.isSOTrx()))
 				{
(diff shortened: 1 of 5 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - can you please take a look to this one?

I tested the following fix:
[IDEMPIERE-4596_fix.txt](https://github.com/idempiere/idempiere/files/5676061/IDEMPIERE-4596_fix.txt)
and it works - but I think it can introduce a performance penalty as it keeps reading the product from database.

From what I see, since IDEMPIERE-4287 we cannot have objects in cache that have not been committed in database, and I think that's safe.

IMO I prefer the performance penalty imposed in the suggested patch than having a potential issue with cached objects not in DB.

Also, not being able to complete documents with just created products can affect the ability to conduct unit tests or fitnesse tests that initially set up all the required data and roll it back at the end (the usual case).

WDYT?

Regards,

Carlos Ruiz

hengsin:
hi @CarlosRuiz-globalqss ,

adding a get (Properties ctx, int M_Product_ID, String trxName) method should work for this, the only difference from existing get (Properties ctx, int M_Product_ID) method is to replace "retValue = new MProduct (ctx, M_Product_ID, (String)null);" with "retValue = new MProduct (ctx, M_Product_ID, trxName;". Note that this doesn't replace the getCopy method as the return product instance is still immutable (and without trxName although it is use to retrieve the product).

There's a small risk of having rollback product instance in cache but in practise, the possibility of that is extremely low.

d-ruiz (approved):

PR #460, opened 2020-12-11
IDEMPIERE-4584 : InfoProduct throws "Attempted to access nonexistent …

nmicoud:
…ListModelTable field at 2, -1" if no IsInstanceAttribute column

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoProductWindow.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoProductWindow.java
@@ -52,7 +52,7 @@ public class InfoProductWindow extends InfoWindow {
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -7892916038089331016L;
+	private static final long serialVersionUID = -640644572459126094L;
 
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoProductWindow.java
```diff
@@ -357,11 +357,13 @@ public void onEvent(Event event) throws Exception {
...
 					
-					Object value = contentPanel.getValueAt(row, findColumnIndex("IsInstanceAttribute"));
-					if (value != null && value.toString().equals("true")) {
-						m_PAttributeButton.setEnabled(true);
-					} else {
-						m_PAttributeButton.setEnabled(false);
+					m_PAttributeButton.setEnabled(false);
+					int colIdx = findColumnIndex("IsInstanceAttribute");
+
+					if (colIdx > 0) {
```
hengsin, line 363: column index is zero base so this should  be >= 0 instead.

hengsin (changes requested):

hengsin (approved):
trivial fix, good to go.

PR #463, opened 2020-12-12
IDEMPIERE-4581 Model generator should isn't dialog it should be window

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4581

comes from https://github.com/idempiere/idempiere/pull/434


hengsin (approved):
tested by carlos and hieplq

PR #464, opened 2020-12-12
IDEMPIERE-4597 : Make sure ConvertLead process generates an active BP…

nmicoud:
…artner


CarlosRuiz-globalqss (approved):
tested successfully

PR #465, opened 2020-12-12

hengsin:
- Refactoring to remove the need to have api to allow cross tenant
reading of data.
- Rename and reduce scope of crossTenantSafe to just the writing of
system tenant records.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MClient.java
+++ b/org.adempiere.base/src/org/compiere/model/MClient.java
@@ -105,14 +105,9 @@ public class MClient extends X_AD_Client implements ImmutablePOSupport
 	public static MClient[] getAll (Properties ctx, String orderBy)
 	{
 		List<MClient> list = null;
-		try {
-			PO.setCrossTenantSafe();
-			list = new Query(ctx,I_AD_Client.Table_Name,(String)null,(String)null)
-					.setOrderBy(orderBy)
-					.list();
-		} finally {
-			PO.clearCrossTenantSafe();
-		}
+		list = new Query(ctx,I_AD_Client.Table_Name,(String)null,(String)null)
+				.setOrderBy(orderBy)
+				.list();
 		for(MClient client:list ){
 			s_cache.put (Integer.valueOf(client.getAD_Client_ID()), client, e -> new MClient(Env.getCtx(), e));
 		}
--- a/org.adempiere.base/src/org/compiere/model/MRegion.java
+++ b/org.adempiere.base/src/org/compiere/model/MRegion.java
@@ -54,14 +54,9 @@ public class MRegion extends X_C_Region
 	{
 		s_regions.clear();
 		List<MRegion> regions;
-		try {
-			PO.setCrossTenantSafe();
-			regions = new Query(Env.getCtx(), Table_Name, "", null)
-					.setOnlyActiveRecords(true)
-					.list();
-		} finally {
-			PO.clearCrossTenantSafe();
-		}
+		regions = new Query(Env.getCtx(), Table_Name, "AD_Client_ID=0", null)
+				.setOnlyActiveRecords(true)
+				.list();
 		for (MRegion r : regions) {
 			r.markImmutable();
 			s_regions.put(r.getC_Region_ID(), r);
--- a/org.adempiere.base/src/org/compiere/model/MUser.java
+++ b/org.adempiere.base/src/org/compiere/model/MUser.java
@@ -798,7 +798,7 @@ public class MUser extends X_AD_User implements ImmutablePOSupport
 		// are found but also roles which delegate org access to the user level where
 		// this user has access to the org in question
 		String sql = "SELECT * FROM AD_Role r " 
-			+ "WHERE r.IsActive='Y' AND r.AD_Client_ID=? " 
+			+ "WHERE r.IsActive='Y' AND r.AD_Client_ID IN (0, ?) " 
 			+ " AND EXISTS (SELECT * FROM AD_User_Roles ur" 
 			+ " WHERE r.AD_Role_ID=ur.AD_Role_ID AND ur.IsActive='Y' AND ur.AD_User_ID=?) "
 			+ " AND ( ( r.isaccessallorgs = 'Y' ) OR "
--- a/org.adempiere.base/src/org/compiere/model/MUser.java
+++ b/org.adempiere.base/src/org/compiere/model/MUser.java
@@ -383,6 +383,8 @@ public class MUser extends X_AD_User implements ImmutablePOSupport
 	private MRole[] 			m_roles = null;
 	/**	Roles of User with Org	*/
 	private int		 			m_rolesAD_Org_ID = -1;
+	/**	AD_Client_ID fpr m_roles above	*/
+	private int					m_rolesAD_Client_ID = -1;
 	/** Is Administrator		*/
 	private Boolean				m_isAdministrator = null;
 	/** User Access Rights				*/
@@ -787,7 +789,7 @@ public class MUser extends X_AD_User implements ImmutablePOSupport
 	 */
 	public MRole[] getRoles (int AD_Org_ID)
 	{
-		if (m_roles != null && m_rolesAD_Org_ID == AD_Org_ID)
+		if (m_roles != null && m_rolesAD_Org_ID == AD_Org_ID && m_rolesAD_Client_ID == Env.getAD_Client_ID(Env.getCtx()))
 			return m_roles;
 		
 		ArrayList<MRole> list = new ArrayList<MRole>();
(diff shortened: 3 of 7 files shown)


org.adempiere.base/src/org/compiere/model/MUser.java
```diff
@@ -796,7 +798,7 @@ public boolean isNotificationNote()
 		// are found but also roles which delegate org access to the user level where
 		// this user has access to the org in question
 		String sql = "SELECT * FROM AD_Role r " 
-			+ "WHERE r.IsActive='Y'" 
+			+ "WHERE r.IsActive='Y' AND r.AD_Client_ID=? "
```
CarlosRuiz-globalqss, line 801: This change breaks the MUser.isAdministrator validation

org.adempiere.base/src/org/compiere/model/MUserDefWin.java
```diff
@@ -104,26 +105,25 @@ public MUserDefWin(Properties ctx, MUserDefWin copy, String trxName)
...
-			if (m_fullList == null) {
-				try {
-					PO.setCrossTenantSafe();
-					m_fullList = new Query(ctx, MUserDefWin.Table_Name, null, null)
-							.setOnlyActiveRecords(true)
-							.list();
-				} finally {
-					PO.clearCrossTenantSafe();
-				}
+		List<MUserDefWin> fullList = null;
```
CarlosRuiz-globalqss, line 108: tested this change successfully with two tenants both customizing Sales Order window

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -4982,35 +4982,35 @@ private void checkValidContext() {
...
 	 */
-	private static ThreadLocal<Boolean> isSafeCrossTenant = new ThreadLocal<Boolean>() {
+	private static ThreadLocal<Boolean> isSystemRecordsWritable = new ThreadLocal<Boolean>() {
 		@Override protected Boolean initialValue() {
 			return Boolean.FALSE;
 		};
 	};
-	public static void setCrossTenantSafe() {
-		isSafeCrossTenant.set(Boolean.TRUE);
+	public static void enterSystemRecordsWritable() {
```
CarlosRuiz-globalqss, line 4999: I still think is good to preserve an option for cross tenant safe readings - I know of implementations that have created a "summarizing" tenant and have extensions that collect data from multiple tenants and summarize (for example for fact_acct table) - we're leaving no option to do that programatically (for swing client for example - is an obsolete piece of code, is easier just to allow the cross-reading than the cost trying to refactor that code)

org.adempiere.base/src/org/compiere/wf/MWorkflow.java
```diff
@@ -125,46 +127,43 @@ public static MWorkflow getCopy(Properties ctx, int AD_Workflow_ID, String trxNa
...
-				PO.setCrossTenantSafe();
-				workflows = new Query(ctx, Table_Name, whereClause, trxName)
-						.setParameters(new Object[]{WORKFLOWTYPE_DocumentValue, true})
-						.setOnlyActiveRecords(true)
-						.setOrderBy("AD_Client_ID, AD_Table_ID")
-						.list();
-			} finally {
-				PO.clearCrossTenantSafe();
-			}
+			workflows = new Query(ctx, Table_Name, whereClause, trxName)
```
CarlosRuiz-globalqss, line 136: tested successfully - creating business partner in two clients with workflow DocValue same as GardenWorld

CarlosRuiz-globalqss (changes requested):
I added my comments in lines.

I would suggest to fix the issues pointed above - but still leave the PO.setCrossTenantSafe() escape way for the potential cases described.

org.adempiere.base/src/org/compiere/model/MClient.java
```diff
@@ -105,14 +105,9 @@ public static MClient get (Properties ctx, int AD_Client_ID)
...
 		List<MClient> list = null;
-		try {
-			PO.setCrossTenantSafe();
-			list = new Query(ctx,I_AD_Client.Table_Name,(String)null,(String)null)
-					.setOrderBy(orderBy)
-					.list();
-		} finally {
-			PO.clearCrossTenantSafe();
-		}
+		list = new Query(ctx,I_AD_Client.Table_Name,(String)null,(String)null)
```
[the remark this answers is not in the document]
hengsin: It should be straightforward to fix the swing client issue. It just doesn't make sense why the need to get all clients records when you login as GardenWorld tenant.

org.adempiere.base/src/org/compiere/model/MRegion.java
```diff
@@ -54,14 +54,9 @@ private static void loadAllRegions ()
...
 		List<MRegion> regions;
-		try {
-			PO.setCrossTenantSafe();
-			regions = new Query(Env.getCtx(), Table_Name, "", null)
-					.setOnlyActiveRecords(true)
-					.list();
-		} finally {
-			PO.clearCrossTenantSafe();
-		}
+		regions = new Query(Env.getCtx(), Table_Name, "AD_Client_ID=0", null)
```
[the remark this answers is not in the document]
hengsin: I will check that again, it should be straightforward fix. Also, the existing logic for s_default in loadAllRegions doesn't work correctly if you can create region records at tenant level - you can end up with region record from other tenant as your default.

org.adempiere.base/src/org/compiere/model/MUser.java
hengsin: so a user is still an administration even when the user is logging in as "GardenWorld User" role ? That doesn't sounds quite right to me.

org.adempiere.base/src/org/compiere/model/PO.java
hengsin: I disagree that we should add an API to make it easy for developer to perform cross tenant operation which is a big security concern in multi tenant, saas environment. For summarizing tenant ( which is not a well define and supported feature in idempiere), you still have the options of view, web service call, direct sql or import/export. Anyway, some customization to code is expected if you are doing something that have not been well defined in the core.

org.adempiere.base/src/org/compiere/model/MUser.java
CarlosRuiz-globalqss: Yes, users with System role are administrators when logged in any tenant - they have certain powers (and those are required) - for example change a password without knowing the previous.

org.adempiere.base/src/org/compiere/model/PO.java
CarlosRuiz-globalqss: I'm not saying to release security checks - we are adding security checks everywhere and we keep adding more to prevent the developers doing wrong things (sometimes prevent completely, sometimes warn, depending on the case).

But still we need to be flexible to allow the developers to do intended things when they know what they are doing, a cross-tenant read is a valid thing in certain scenarios, forbidding totally that option can break existing code out there, and we would be forcing them use JDBC (or fork).

Indeed you're adding an API to write in System, which sounds way more dangerous than cross-tenant reading, and we need to trust that developers use that enterSystemRecordsWritable API knowing what they are doing.

I think our difference of opinions is that you consider any cross-tenant code as wrong - but I think there are cases (uncommon) where that is valid and intended - and I would prefer to be open to such thing.

IMO having setCrossTenantSafe is a good compromise - the developer can do it - but he needs to consciously and explicitly tell the system that he's doing that.
We cannot pretend that iDempiere is protected just dropping the API setCrossTenantSafe, because the developer can simply do that in JDBC and that's something we cannot forbid.

hengsin: The cross-tenant API allows write in System in addition to cross tenant operation and I'm just reducing its scope to only write in System. In fact, it is possible to get rid of it in Core since the only place in core that's using it now is AD_Preference, i.e that can be changed to as an exception for just that table.

Yes, we have the existing JDBC hole in the system (no, it is a huge task but not something impossible to forbid for plugins) but that shouldn't be the reason to add more hole to the system. It would be hard to get rid of a public API once we added it to a release.

org.adempiere.base/src/org/compiere/model/MUser.java
hengsin: ok, that's a trivial fix, I will change that.

org.adempiere.base/src/org/compiere/model/PO.java
CarlosRuiz-globalqss: I have different opinion, JDBC is not a hole.

If JDBC is a hole, then pl/pgsql is a hole too, and the complete SQL language, those are tools to be used when the developer have the proper permissions and I guess somebody must be in charge of giving permissions wisely and review the proper usage of those permissions.

Samely, allowing cross tenant safe readings can be a valid operation, and a required one in certain scenarios.

hengsin:
alright, it seems it is just a waste of my time and effort to continue working on this. I will just close this and leave it as it is.

PR #466, opened 2020-12-12
IDEMPIERE-4287 Cache API not thread safe and inconsistent with contex…

hengsin:
…t parameter

- Fix PO immutable exception with Workflow Validate process.


CarlosRuiz-globalqss (approved):
tested successfully

PR #475, opened 2020-12-18
IDEMPIERE-1575 Increase some columns length

CarlosRuiz-globalqss:
AD_Rule.Script -> 4000
AD_ChangeLog.TrxName -> 1000
AD_SysConfig.Value -> 4000

https://idempiere.atlassian.net/browse/IDEMPIERE-1575


hengsin (approved):
trivial fix, good to go

PR #487, opened 2020-12-27
IDEMPIERE-4612

gsjeffen:
Fix the layout completely.
Tested using Chrome, Firefox, Edge and Opera browsers.
#resolve

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
+++ b/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
@@ -13,3 +13,24 @@
 .find-window-advanced .img-btn, .find-window-advanced .img-btn.btn-ok.z-button, .find-window-advanced .img-btn.btn-cancel.z-button {
 	margin: 0 2px;
 }
+ 
+.find-window .z-window-content{
+	padding-bottom: 0px;	
+}
+
+.find-window .status{
+    height: 20px;
+}
+
+.find-window .statusbar{
+    height: 22px;
+}
+
+.find-window .z-grid,
+.find-window .z-listbox.z-flex-item{
+	width: calc(100% - 1px) !important;
+}
+
+.find-window .z-toolbar{
+	border-bottom-width: 0px;
+}


org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
```diff
@@ -13,3 +13,24 @@
 .find-window-advanced .img-btn, .find-window-advanced .img-btn.btn-ok.z-button, .find-window-advanced .img-btn.btn-cancel.z-button {
 	margin: 0 2px;
 }
+ 
+.find-window .z-window-content{
+	padding-bottom: 0px;	
+}
+
+.find-window .status{
+    height: 20px;
```
hengsin, line 22: The amount of white space is a subjective matter and it is not appropriate to make this change as part of a bug fix of other issues.

org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
```diff
@@ -13,3 +13,24 @@
...
+}
+
+.find-window .statusbar{
+    height: 22px;
```
hengsin, line 26: The amount of white space is a subjective matter and it is not appropriate to make this change as part of a bug fix of other issues.

org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
```diff
@@ -13,3 +13,24 @@
...
+}
+
+.find-window .z-grid,
+.find-window .z-listbox.z-flex-item{
+	width: calc(100% - 1px) !important;
```
hengsin, line 31: you have to change the width of the .z-toolbar above the list box as well, otherwise there's a 1px difference between them.

hengsin (changes requested):
For theme changes, you have to change the "idempiere.theme.default" version org.adempiere.ui.zk/WEB-INF/src/metainfo/zk/lang-addon.xml as well (it is a GMT0 timestamp)

org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/find-window.css.dsp
hengsin: for consistency, it should be .z-grid.z-flex-item too.

hengsin:
replace by pr 495

PR #490, opened 2020-12-29
IDEMPIERE-4602 Encrypt passwords on properties files (FHCA-1982)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4602

* backward compatible with the previous idempiere.properties and idempiereEnv.properties version
* when the setup/console-setup is executed again the secret keys are stored obfuscated in a different file .idpass
  * Secret keys are ADEMPIERE_DB_PASSWORD, ADEMPIERE_DB_SYSTEM, ADEMPIERE_MAIL_PASSWORD
  * to add more is just adding keys to array ConfigurationData.secretVars
* the previous (unobfuscated) approach is still preserved passing -DIDEMPIERE_SECURE_PROPERTIES=false to the JVM in setup and server
* the approach just run on Linux - as is implemented using shell script, windows is out of the initial scope, but could be possible to implement .bat files to do similar
* the default approach is to use getVar.sh and setVar.sh that writes in .idpass obfuscated
  * is possible to extend and use custom secret managers implementing customSetVar.sh and customGetVar.sh
  * samples for amazon AWS secretsmanager are included
* jetty ADEMPIERE_KEYSTOREPASS is obfuscated
* avoid the email sent on setup sending the secret keys
* restore the setup and console-setup previous behavior that received a log level as parameter
* avoid creating duplicate logs in setup - create the log in log folder
* avoid creating the log on /tmp when the max logs are reached (100) - instead of creates a log using datetime in log folder
* implement creating log also in console-setup
* mark SilentSetup as deprecated
* enclose all variables in myEnvironment.sh within quotes (this avoids problems with variables containing spaces)
* add coreutils as required for debian installer (as we use base64 now to obfuscate passwords)
* swing client is not affected as it saves the idempiere.properties encrypted in user home folder


org.adempiere.base/src/org/compiere/util/DB.java
```diff
@@ -240,17 +245,30 @@ public static void updateMail()
...
 		//
+		FileOutputStream out = null;
 		try
 		{
 			env.setProperty("ADEMPIERE_MAIL_UPDATED", "Y");
-			FileOutputStream out = new FileOutputStream(envFile);
+			out = new FileOutputStream(envFile);
 			env.store(out, "");
-			out.flush();
-			out.close();
```
hengsin, line 249: the latest recommended pattern is try (FileOutputStream out = new FileOutputStream(envFile)) instead of close in finally clause.

org.adempiere.install/src/org/compiere/install/ConfigurationData.java
```diff
@@ -860,6 +915,25 @@ public boolean save()
 				System.err.println(ConfigurationPanel.res.getString("ErrorSave"));
 			return false;
 		}
+		finally
+		{
+			if (fos != null) {
+				try {
+					fos.flush();
+					fos.close();
```
hengsin, line 923: this is a bad pattern that defeat the purpose of having close in finally. you want to ensure close is try regardless whether flush is throwing exception.

org.adempiere.base/src/org/compiere/util/DBReadReplica.java
```diff
@@ -168,6 +168,8 @@ private static void setUserPass() {
 					m_pass = value;
 				}
 			}
+			if (m_pass == null && MSystem.isSecureProps())
+				m_pass = Ini.getVar("ADEMPIERE_DB_PASSWORD");
```
hengsin, line 172: this is a questionable one - now all read replica must use the same password as the master db ?

CarlosRuiz-globalqss:
Closing to reopen after https://github.com/idempiere/idempiere/pull/494 is approved

Patch file on top of IDEMPIERE-4618 at https://idempiere.atlassian.net/secure/attachment/38717/38717_IDEMPIERE-4602.patch

PR #493, opened 2020-12-30
IDEMPIERE-4620 Improvements for Setup programs

CarlosRuiz-globalqss:
* Allow receiving log level as parameter for setup and console-setup
* Avoid duplication of log file - just leave it in /log
* Implement logging for console-setup too
* mark SilentSetup as deprecated
* add runtime-*.app to .gitignore (sometimes files appear in these folders and can be wrongly committed)

https://idempiere.atlassian.net/browse/IDEMPIERE-4620

### DIFF
--- a/org.adempiere.install/src/org/compiere/install/Setup.java
+++ b/org.adempiere.install/src/org/compiere/install/Setup.java
@@ -150,17 +150,17 @@ public class Setup extends JFrame implements ActionListener
 	public static void main(String[] args)
 	{
 		CLogMgt.initialize(false);
-		// Duplicate log file - not required
-		// Handler fileHandler = new CLogFile(System.getProperty("user.dir"), false, false);
-		// CLogMgt.addHandler(fileHandler);
+
 		//	Log Level
-		if (args.length > 0)
-			CLogMgt.setLevel(args[0]);
-		else
-			CLogMgt.setLevel(Level.INFO);
-		//	File Logger at least FINE
-		// if (fileHandler.getLevel().intValue() > Level.FINE.intValue())
-		//	fileHandler.setLevel(Level.FINE);
+		Level logLevel = Level.INFO;
+		if (args.length > 0) {
+			try {
+				logLevel = Level.parse(args[0]);
+			} catch (IllegalArgumentException e) {
+				CLogger.get().warning("Unrecognized log level: " + args[0] + " defaulting to: " + logLevel);
+			}
+		}
+		CLogMgt.setLevel(logLevel);
 
 		instance = new Setup();
 	}	//	main
--- a/org.adempiere.install/src/org/compiere/install/Setup.java
+++ b/org.adempiere.install/src/org/compiere/install/Setup.java
@@ -22,7 +22,6 @@ import java.awt.Cursor;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.util.ResourceBundle;
-import java.util.logging.Handler;
 import java.util.logging.Level;
 
 import javax.swing.BorderFactory;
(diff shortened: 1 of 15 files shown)


org.adempiere.install/src/org/compiere/install/Setup.java
```diff
@@ -151,17 +149,18 @@ else if (e.getSource() == menuHelpInfo)
...
-		CLogMgt.initialize(true);
-		Handler fileHandler = new CLogFile(System.getProperty("user.dir"), false, false);
-		CLogMgt.addHandler(fileHandler);
+		CLogMgt.initialize(false);
+		// Duplicate log file - not required
+		// Handler fileHandler = new CLogFile(System.getProperty("user.dir"), false, false);
+		// CLogMgt.addHandler(fileHandler);
 		//	Log Level
 		if (args.length > 0)
 			CLogMgt.setLevel(args[0]);
```
hengsin, line 158: I think we should validate the arguments here.

org.adempiere.install/src/org/compiere/install/console/ConsoleInstallApplication.java
```diff
@@ -32,6 +34,12 @@
 	 */
 	@Override
 	public Object start(IApplicationContext context) throws Exception {
+		CLogMgt.initialize(false);
+		String[] args = (String[]) context.getArguments().get(IApplicationContext.APPLICATION_ARGS);
+		if (args.length > 0)
+			CLogMgt.setLevel(args[0]);
```
hengsin, line 40: should validate the arguments here

hengsin (approved):
please add validation for the log level parameter

org.adempiere.install/src/org/compiere/install/Setup.java
```diff
@@ -151,17 +149,18 @@ else if (e.getSource() == menuHelpInfo)
...
 		else
 			CLogMgt.setLevel(Level.INFO);
 		//	File Logger at least FINE
-		if (fileHandler.getLevel().intValue() > Level.FINE.intValue())
-			fileHandler.setLevel(Level.FINE);
+		// if (fileHandler.getLevel().intValue() > Level.FINE.intValue())
+		//	fileHandler.setLevel(Level.FINE);
```
hengsin, line 163: with history from git, it seems strange to me that we will still do remark instead of remove.

org.adempiere.install/src/org/compiere/install/SilentSetup.java
```diff
@@ -61,17 +60,18 @@ public SilentSetup()
...
 		//	Log Level
 		if (args.length > 0)
 			CLogMgt.setLevel(args[0]);
 		else
 			CLogMgt.setLevel(Level.INFO);
 		//	File Logger at least FINE
-		if (fileHandler.getLevel().intValue() > Level.FINE.intValue())
-			fileHandler.setLevel(Level.FINE);
+		// if (fileHandler.getLevel().intValue() > Level.FINE.intValue())
+		//	fileHandler.setLevel(Level.FINE);
```
hengsin, line 74: with history from git, it seems strange to me that we will still do remark instead of remove.

org.adempiere.install/src/org/compiere/install/Setup.java
CarlosRuiz-globalqss: This was legacy code - but is a good idea to check it - I added the check to the three setup programs

CarlosRuiz-globalqss:
Thanks  @hengsin - this is read for peer review.

Instead of deprecating silent setup I decided to implement it again as CLI - the approach of redirecting the stdin for console-setup has been prone to errors - sometimes it enters in an endless loop.

I tested the silent-setup, and also used this one for the debian installer, which was tested successfully in a fresh VM with debian 20.04.

Regards,

Carlos Ruiz

hengsin (approved):
looks fine.

PR #497, opened 2020-12-31
IDEMPIERE-4583 Fix OSGi split package

hengsin:
Use only Require-Bundle for dependency between idempiere module/bundles

https://idempiere.atlassian.net/browse/IDEMPIERE-4583


CarlosRuiz-globalqss (approved):
Tested

PR #498, opened 2020-12-31
IDEMPIERE-4602 Encrypt passwords on properties files (FHCA-1982)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4602

* backward compatible with the previous idempiere.properties and idempiereEnv.properties version
* when the setup/console-setup is executed again the secret keys are stored obfuscated in a different file .idpass
  * Secret keys are ADEMPIERE_DB_PASSWORD, ADEMPIERE_DB_SYSTEM, ADEMPIERE_MAIL_PASSWORD
  * to add more is just adding keys to array ConfigurationData.secretVars
* the previous (unobfuscated) approach is still preserved passing -DIDEMPIERE_SECURE_PROPERTIES=false to the JVM in setup and server
* the approach just run on Linux - as is implemented using shell script, windows is out of the initial scope, but could be possible to implement .bat files to do similar
* the default approach is to use getVar.sh and setVar.sh that writes in .idpass obfuscated
  * is possible to extend and use custom secret managers implementing customSetVar.sh and customGetVar.sh
  * samples for amazon AWS secretsmanager are included
* avoid the email sent on setup sending the secret keys
* enclose all variables in myEnvironment.sh within quotes (this avoids problems with variables containing spaces)
* add coreutils as required for debian installer (as we use base64 now to obfuscate passwords)
* swing client is not affected as it saves the idempiere.properties encrypted in user home folder


hengsin (approved):
not my cups of tea but anyway there are already many others existing one, doesn't really matter to add yet another one ...

org.adempiere.base/src/org/compiere/util/Ini.java
```diff
@@ -794,4 +798,76 @@ public static String getPropertyFileName()
...
+
+	public static void setVar(String secretVar, String secretValue) {
+		String cmd = getUtilsCmd("setVar");
+		String[] command = new String[] {
+				cmd,
+				secretVar,
+				secretValue
+		};
+		@SuppressWarnings("unused") // used for debugging purposes
+		String retValue = runCommand(command);
```
hieplq, line 820: @CarlosRuiz-globalqss don't need this one. eclipse already support
in case we use a IDE hasn't this feature then debug at end of runCommand better

image

org.adempiere.base/src/org/compiere/util/Ini.java
```diff
@@ -794,4 +798,76 @@ public static String getPropertyFileName()
...
+			while ((s = stdInput.readLine()) != null) {
+				msg.append(s);
+			}
+			// read any errors from the attempted command
+			while ((s = stdError.readLine()) != null) {
+				msg.append(s);
+			}
+			try {
+				p.waitFor(5, TimeUnit.SECONDS);
+			} catch (InterruptedException e) {
```
hieplq, line 860: from API of Process.waitFor
> Returns:
true if the process has exited and false if the waiting time elapsed before the process has exited.
Throws:
InterruptedException - if the current thread is interrupted while waiting.

so message "Timeout waiting for" isn't correct for InterruptedException

by above mistake then next check "if (p.exitValue() != 0) {" too soon process can be not yet done when code run to there

CarlosRuiz-globalqss:
Thanks @hieplq - pull request created
https://github.com/idempiere/idempiere/pull/508

PR #501, opened 2020-12-31
IDEMPIERE-4485 Info Window Columns has no Value preference

igorpojzl:
Problem is that Info Column has ReadOnly parameter which have 2 uses. So if we set Parameter as ReadOnly, this parameter is in table set as read only. But if this parameter is set as Query Parameter it will interfere with menu popup which checks ReadOnly Parameter.
So i created parameterTree to store new Gridfield copy which have forced read only false in Query Parameter.

https://idempiere.atlassian.net/browse/IDEMPIERE-4485

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
@@ -981,7 +981,7 @@ public class InfoWindow extends InfoPanel implements ValueChangeListener, EventL
 					builder.append(whereClause);
 				}
 			} else if (editor.getGridField() != null && editor.getValue() != null && editor.getValue().toString().trim().length() > 0) {
-				InfoColumnVO InfoColumnVO = findInfoColumn(editor.getGridField());
+				InfoColumnVO InfoColumnVO = findInfoColumnParameter(editor.getGridField());
 				if (InfoColumnVO == null || InfoColumnVO.getSelectClause().equals("0")) {
 					continue;
 				}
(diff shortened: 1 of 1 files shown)


igorpojzl:
- [ ] TODO Filter in InfoWindow is Broken. Wait for Fix.

hengsin:
hi @igorpojzl , is that change set complete ? I didn't see any additions that will show popup/context menu for info window query/parameter field.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -624,6 +626,40 @@ protected boolean loadInfoDefinition() {
...
+					if (infoColumn.getDisplayLogic() != null)					
+						vo.DisplayLogic =  infoColumn.getDisplayLogic();
+					if (infoColumn.isQueryCriteria() && infoColumn.getDefaultValue() != null)
+						vo.DefaultValue = infoColumn.getDefaultValue();
+					desc = infoColumn.getDescriptionTrl();
+					vo.Description = desc != null ? desc : "";
+					help = infoColumn.getHelpTrl();
+					vo.Help = help != null ? help : "";
+					vo.AD_FieldStyle_ID = infoColumn.getAD_FieldStyle_ID();
+					vo.IsAutocomplete = infoColumn.isAutocomplete();
```
hengsin, line 652: the code before vo.IsReadOnly can be shorten to the following:
vo = vo.clone(infoContext, p_WindowNo, 0, vo.AD_Window_ID, 0, false);

hengsin (changes requested):
tested working but it can be shorten quite a bit.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
igorpojzl: Commited changes

hengsin (approved):
tested

hengsin:
I wrote above "the code before vo.IsReadOnly can be shorten to the following", didn't notice @igorpojzl has remove the vo.IsReadOnly = false line as well. that's needed.

igorpojzl:
Oh, my bad sorry, Ive missread the clone method. I thought it set IsReadOnly parameter.

PR #502, opened 2020-12-31
IDEMPIERE-4621  Cross Tenant Parent Tax Error

igorpojzl:
https://idempiere.atlassian.net/browse/IDEMPIERE-4621

Saving OrderTax After Deleting it caused CrossTennant Error.
Why do we save deleted OrderTax?


hengsin (approved):
trivial fix, good to go.

PR #509, opened 2021-01-05
IDEMPIERE-4630  Global Search, Tab Search has no scroll

igorpojzl:
Added overflow style parameter to VLayout.

https://idempiere.atlassian.net/browse/IDEMPIERE-4630


hengsin (approved):
trivial fix, good to go.

PR #510, opened 2021-01-05
IDEMPIERE-4485 Info Window Columns has no Value preference

hengsin:
add back the line that's wrongly remove in the review of PR 501


norbertbede:
test: pass, time: jan 6 after jenkins build  at 00:09
thanks @igorpojzl @hengsin for cooperation.

PR #512, opened 2021-01-06
IDEMPIERE-4632 Print Invoices is not translating the invoice if the r…

d-ruiz:
…eport is a Jasper Report

https://idempiere.atlassian.net/browse/IDEMPIERE-4632


CarlosRuiz-globalqss (approved):
looks good

PR #518, opened 2021-01-09
Fix for the safari browser tile image positioning problem.

Soundbytes:
This is a fix for the safari tile image positioning issue:

Bildschirmfoto 2021-01-09 um 11 59 32

Discussion here: https://groups.google.com/g/idempiere/c/soHxAlytJMs


hengsin (approved):
tested

PR #523, opened 2021-01-12
IDEMPIERE-4644 Heading only Field is the cause of NullPointerException Error on Find Window

HideakiHagiwara:
This is pull request of IDEMPIERE-4644 : Heading only Field is the cause of NullPointerException Error on Find Window.
I added null check logic.
And test below.

・I confirmed that NullPointerException Error on Find Window Don't display when there is a Heading only Field. 
・In case of tick "Lookup Only Selection Columns" of tab setting,  it also seemed good .


CarlosRuiz-globalqss (approved):
tested

PR #524, opened 2021-01-13
IDEMPIERE-4645 Search by special field types failing

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4645


hengsin (approved):
tested

PR #525, opened 2021-01-15
IDEMPIERE-4625 Login Page refresh cause error message, randomly when …

hengsin:
…slow db response

Fix CLOSE-WAIT issue
(https://groups.google.com/g/idempiere/c/wppbB5jdByg/m/_VOzSZZyAQAJ)


CarlosRuiz-globalqss (approved):
Tests worked fine

PR #526, opened 2021-01-15
IDEMPIERE-917 : Fill City from Postal

nmicoud:


CarlosRuiz-globalqss:
Closing this - as per discussion in ticket is better to do a different design

PR #529, opened 2021-01-15
IDEMPIERE-4649 NPE when running Change base language with a fake language (as instructed)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4649


hengsin (approved):
trivial fix, good to go

PR #531, opened 2021-01-18
IDEMPIERE-4651 Max records for Search autocomplete list should be con…

hengsin:
…figurable

https://idempiere.atlassian.net/browse/IDEMPIERE-4651


CarlosRuiz-globalqss (approved):
looks OK

PR #535, opened 2021-01-20
IDEMPIERE-4655 WTimeEditor display time picker button when read only.

HideakiHagiwara:
I tested below case.

* In case of processed = 'Y' at Window.
* In case of IsActive = 'Y' at Window.
* In case of Read only field at Window.
* In case of apply Read only Logic at Window. 
* In case of apply Read only Logic at Process. 

It's seems good.

best regards.


hengsin (approved):
trivial fix, good to go.

PR #536, opened 2021-01-20
IDEMPIERE-4654 Sales Order After Receipt Delivery Rule is not working

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4654

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MOrder.java
+++ b/org.adempiere.base/src/org/compiere/model/MOrder.java
@@ -3010,8 +3010,7 @@ public class MOrder extends X_C_Order implements DocAction
 	public BigDecimal getPaymentAmt()
 	{
 		BigDecimal retValue = null;
-		String sql = "SELECT SUM(currencyConvert(p.PayAmt+p.DiscountAmt+p.WriteOffAmt, "
-				+ "p.C_Currency_ID, o.C_Currency_ID, p.DateTrx, COALESCE(o.C_ConversionType_ID,0), p.AD_Client_ID, p.AD_Org_ID) "
+		String sql = "SELECT SUM(currencyconvertpayment(p.c_payment_id, o.c_currency_id, p.PayAmt+p.DiscountAmt+p.WriteOffAmt, null) "
 				+ " * (CASE WHEN p.IsReceipt='Y' THEN 1 ELSE -1 END)) "
 				+ "FROM C_Payment p "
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/model/MOrder.java
```diff
@@ -3000,4 +3003,35 @@ public int getDocTypeID()
...
 	}
 
+	/**
+	 * 
+	 * @return prepayment amount for order
+	 */
+	public BigDecimal getPaymentAmt()
+	{
+		BigDecimal retValue = null;
+		String sql = "SELECT SUM(currencyConvert(p.PayAmt+p.DiscountAmt+p.WriteOffAmt, "
```
CarlosRuiz-globalqss, line 3013: Hi @hengsin - I have not tested yet, but maybe here we must use the new currencyconvertpayment function to make it compatible with IDEMPIERE-4083

org.adempiere.base/src/org/compiere/process/ServerProcessCtl.java
```diff
@@ -341,4 +342,21 @@ protected boolean startDBProcess (String ProcedureName)
...
+	public void setManagedTrxForJavaProcess(boolean managedTrx)
+	{
+		this.managedTrxForJavaProcess = managedTrx;
+	}
+	
+	/**
+	 * 
+	 * @return true if java process call will commit/rollback trx
+	 */
+	public boolean isManagedTrxForJavaProcess()
```
CarlosRuiz-globalqss, line 358: just a question - what is the purpose of this method?  and how to use it when needed?

hengsin: added for unit testing so that trx is not committed upon completion of running a process (for this case, the InOutGenerate process).

org.adempiere.base/src/org/compiere/model/MOrder.java
hengsin: ok, changed.

CarlosRuiz-globalqss (commented):
Tested, comments in ticket

CarlosRuiz-globalqss (approved):
Hi @hengsin, tested successfully.

I pushed a minor change to the migration script, please squash it if you agree with those changes.

Regards,

Carlos Ruiz

PR #538, opened 2021-01-20
IDEMPIERE-4658 The operator Dropdown entries for a foreign ID column is corrupted when a search preset is loaded in any window's Advanced Search tab.

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4658

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
@@ -1220,6 +1220,7 @@ public class FindWindow extends Window implements EventListener<Event>, ValueCha
 
             boolean selected = false;
             listColumn.appendItem("","");
+            ListItem liCol = null;
             for (int i = 0; i < cols.length; i++)
             {
             	ValueNamePair item = cols[i];
@@ -1228,7 +1229,7 @@ public class FindWindow extends Window implements EventListener<Event>, ValueCha
             	{
                 	listColumn.setSelectedItem(li);
             		selected = true;
-            		break;
+            		liCol = li;
             	}
             }
             if(!selected) listColumn.setSelectedIndex(0);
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
```diff
@@ -1228,7 +1229,7 @@ private void setValues(Listbox listColumn, Listbox listOperator, String[] fields
             	{
                 	listColumn.setSelectedItem(li);
             		selected = true;
-            		break;
```
hengsin, line 1231: why the need to remove the break ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
```diff
@@ -1242,10 +1243,12 @@ private void setValues(Listbox listColumn, Listbox listOperator, String[] fields
             	{
             		listOperator.setSelectedItem(li);
             		selected = true;
-            		break;
```
hengsin, line 1245: same here, why remove the break ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
CarlosRuiz-globalqss: The list of columns was being shown incomplete.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FindWindow.java
CarlosRuiz-globalqss: The list of operators was being filled incomplete.

hengsin (approved):
good to go

Soundbytes:
I just pulled the patch and tested it. It works as expected. 
Thanks for the quick response. :)

PR #539, opened 2021-01-20
IDEMPIERE-4653 Improve timeout handling of window tab

CarlosRuiz-globalqss:
* Fixed a problem with IDEMPIERE-4130 that was not using role max query records
* GridTable was throwing NPEs because the arrays were disposed before closing
* Lowered log level on MRole to avoid clogging the console (I was testing with INFO level)
* Added more points of control for the timeout exception to give meaningful errors to the user and allow refining the query without closing the Find dialog
* Implemented a control to avoid repeating a slow initial query every time the dialog is open
* Added dialog when the find criteria doesn't return rows and allows the user to refine the query
* When the initial number of records is unknown, show a question mark instead of the count

https://idempiere.atlassian.net/browse/IDEMPIERE-4653

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRole.java
+++ b/org.adempiere.base/src/org/compiere/model/MRole.java
@@ -124,7 +124,7 @@ public final class MRole extends X_AD_Role implements ImmutablePOSupport
 	 */
 	public synchronized static MRole get (Properties ctx, int AD_Role_ID, int AD_User_ID, boolean reload)
 	{
-		if (s_log.isLoggable(Level.INFO)) s_log.info("AD_Role_ID=" + AD_Role_ID + ", AD_User_ID=" + AD_User_ID + ", reload=" + reload);
+		if (s_log.isLoggable(Level.CONFIG)) s_log.info("AD_Role_ID=" + AD_Role_ID + ", AD_User_ID=" + AD_User_ID + ", reload=" + reload);
 		String key = AD_Role_ID + "_" + AD_User_ID;
 		MRole role = (MRole)s_roles.get (key, e -> new MRole(ctx, e));
 		if (role == null || reload)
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/model/MRole.java
```diff
@@ -124,7 +124,7 @@ private static MRole getDefaultRole() {
 	 */
 	public synchronized static MRole get (Properties ctx, int AD_Role_ID, int AD_User_ID, boolean reload)
 	{
-		if (s_log.isLoggable(Level.INFO)) s_log.info("AD_Role_ID=" + AD_Role_ID + ", AD_User_ID=" + AD_User_ID + ", reload=" + reload);
+		if (s_log.isLoggable(Level.CONFIG)) s_log.info("AD_Role_ID=" + AD_Role_ID + ", AD_User_ID=" + AD_User_ID + ", reload=" + reload);
```
hengsin, line 127: this doesn't looks correct, the s_log call is using info level.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADTabpanel.java
```diff
@@ -1172,7 +1172,7 @@ public void query (boolean onlyCurrentRows, int onlyCurrentDays, int maxRows)
     	{
     		if (DBException.isTimeout(e)) 
     		{
-    			FDialog.error(windowNo, GridTable.LOAD_TIMEOUT_ERROR_MESSAGE);
+    			throw e;
```
hengsin, line 1175: what's the need change this one here ?

org.adempiere.base/src/org/compiere/model/MRole.java
CarlosRuiz-globalqss: thanks - fixed

hengsin (approved):
good to go

PR #546, opened 2021-01-22
IDEMPIERE-4665 : Toolbar: Select Query cannot be hidden

nmicoud:


d-ruiz (approved):
Tested and it is working, there are some additional changes that should be added but I will create a new Pull Request for that

PR #549, opened 2021-01-22
IDEMPIERE-4668 update*.bat not included in windows installers

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4668


hengsin (approved):
trivial fix, good to go

PR #550, opened 2021-01-23
IDEMPIERE-4667 Project can not be printed using template print format

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4667


CarlosRuiz-globalqss (approved):
tested

PR #552, opened 2021-01-24
IDEMPIERE-4670 RUN_ImportIdempiere can call RUN_SyncDB and sign-database

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4670


hengsin (approved):
trivial fix, good to go

PR #553, opened 2021-01-25
Improve the StatusBar Menupopup to render HTML text

d-ruiz:
https://idempiere.atlassian.net/browse/IDEMPIERE-4672


CarlosRuiz-globalqss (approved):
tested

PR #554, opened 2021-01-26
IDEMPIERE-4673 PackOut fails to export attachments with File storage provider

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4673


hengsin (approved):
looks good.

PR #555, opened 2021-01-27
IDEMPIERE-4675 Add model factory base class backed by Map and Lambda …

hengsin:
…functional object

https://idempiere.atlassian.net/browse/IDEMPIERE-4675


CarlosRuiz-globalqss (approved):
unit test passed

PR #556, opened 2021-01-27
IDEMPIERE-4677 : Cross Tenant issues with AD_User/Role_OrgAccess

nmicoud:


CarlosRuiz-globalqss (approved):
Looks good

hieplq:
Calling Env.getAD_Client_ID(getCtx()) within the loadOrgAccessUser function can cause issues if the context is not yet logged in. To avoid this, it's recommended to use getAD_Client_ID() directly instead.

Is there a specific reason why we need to use Env.getAD_Client_ID(getCtx()) in this case?

CarlosRuiz-globalqss:
> Calling Env.getAD_Client_ID(getCtx()) within the loadOrgAccessUser function can cause issues if the context is not yet logged in. To avoid this, it's recommended to use getAD_Client_ID() directly instead.
> Is there a specific reason why we need to use Env.getAD_Client_ID(getCtx()) in this case?

Yes @hieplq - I think it makes sense - can you prepare a pull request with your suggestion?

PR #559, opened 2021-01-28
IDEMPIERE-4678 Save log informing Notice created when background job

CarlosRuiz-globalqss:
* Added also a refactor for the INSERT INTO AD_PInstance_Log to use safer binding variables and add method saveEx

https://idempiere.atlassian.net/browse/IDEMPIERE-4678

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
+++ b/org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
@@ -74,6 +74,8 @@ public class MPInstanceLog
 	private int m_P_ID;
 	private BigDecimal m_P_Number;
 	private String m_P_Msg;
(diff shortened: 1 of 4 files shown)


org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
```diff
@@ -97,53 +99,62 @@ public String toString ()
 	}	//	toString
 
 
+	final String insertSql = "INSERT INTO AD_PInstance_Log "
```
hengsin, line 102: this should be private final static

hengsin:
Probably should apply to ProcessInfoUtil.saveLogToDB as well.

hengsin (approved):
looks good.

PR #561, opened 2021-01-29
IDEMPIERE-4681 Can't use hidden column in display logic

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4681


CarlosRuiz-globalqss (changes requested):
comment on ticket

CarlosRuiz-globalqss:
Hi @hengsin,

About
> test case failed:
> Display Logic = @SO_CreditAvailable:0@=0

changing
`x.DisplayLogic Like '%@'||c.ColumnName||'@%'`
with
`x.DisplayLogic Like '%@'||c.ColumnName||'%'`
can do the trick.

But I'm not sure the consequences of getting extra columns that are unrelated, with this fix you could also get wrongly a column named @SO_CreditAvailable_New@ - does it have consequences getting an unrelated column?

If yes, then maybe an OR could be better - but that makes it dependent on more things that could be implemented in the display logic, like @C_BPartner_ID.Name@ or @DateAcct<yyyy-mm-dd>@

WDYT?

Regards,

Carlos Ruiz

hengsin:
Hi @CarlosRuiz-globalqss ,

I think it is safer to use the OR approach.

Regards,
Low

CarlosRuiz-globalqss (approved):

PR #567, opened 2021-02-02
IDEMPIERE-4690 Add column callout factory base class backed by Map an…

hengsin:
…d Lambda functional object

https://idempiere.atlassian.net/browse/IDEMPIERE-4690


CarlosRuiz-globalqss (approved):
unit tests successful

PR #574, opened 2021-02-09
IDEMPIERE-4676 Browser hangs after restarting iDempiere or refreshing…

hengsin:
… ui.zk, on macos sometimes during work

Immediately treat jqxhr.status=0 as server not available doesn't work
well on firefox. With Firefox both stop of server and navigate away to
another site provide same jqxhr.onerror event and jqxhr.status of 0.


CarlosRuiz-globalqss (approved):
tested successfully

PR #575, opened 2021-02-09
IDEMPIERE-1640 Can't create Credit Memo from Customer RMA when original invoice is multishipment

mschnbeck:
try to find the invoice via orderlines if not found via order


CarlosRuiz-globalqss (changes requested):
Hi @mschnbeck - can you please add a unit test case for this scenario that fails before the patch and passes after the patch.

mschnbeck:
Hi @CarlosRuiz-globalqss,
is there a wiki entry how to create a unit test? And perhaps a skeleton? I never did such unit test and I'm short of time at the moment, so anything what helps me to save time would be appreciated.

CarlosRuiz-globalqss:
You can check in the folder [org.idempiere.test/src/org/idempiere/test/model](https://github.com/idempiere/idempiere/tree/master/org.idempiere.test/src/org/idempiere/test/model).

We have there some test classes for invoices and orders, in this case I think you would add a new CustomerRMATest.java

mschnbeck:
Hi @CarlosRuiz-globalqss,

I must admit, that I have no idea how this test cases are intended to be started. Can you give me a hint?

hengsin:
> Hi @CarlosRuiz-globalqss,
> 
> I must admit, that I have no idea how this test cases are intended to be started. Can you give me a hint?

you means how to run it ? 

For Eclipse, Run/Debug Configurations ... > JUnit Plug-in Test > idempiere.unit.test (To run just your unit test case, duplicate idempiere.unit.test and change the "Test class" at the "Test" tab.

For maven, mvn verify -DskipTests=false

mschnbeck:
Hi @hengsin,

when preparing a test case for this I wasn't able to reproduce it again. And unfortunately I didn't remember exactly which way we ran into the problem. I tried several variants but couldn't reproduce it. So I rolled back the changes in the installation where it happened and since then I'm waiting for it to happen again. While I'm very sure regarding the problem, I think, for the moment we could reject the changes.

Regards,
Martin

hengsin:
close, not reproduceable anymore.

PR #576, opened 2021-02-10
IDEMPIERE-4695 Add document factory base class backed by Map and Lamb…

hengsin:
…da functional object

MappedDocumentFactoryTest fail when running idempiere.unit.test due to
document factory cache.


CarlosRuiz-globalqss (approved):
unit test passed

PR #578, opened 2021-02-10
IDEMPIERE-4700 PackOut is creating index and view data for Dictionary entries when export dictionary is disabled

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4700


hengsin (approved):
trivial fix, good to go

PR #580, opened 2021-02-11
IDEMPIERE-4606: Adding support for generic method to get model using …

dpansheriya:
…UUID

### DIFF
--- a/org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
+++ b/org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
@@ -322,4 +322,14 @@ public class DefaultModelFactory implements IModelFactory {
 			s_log.log(Level.SEVERE, "(rs) - Not found - Table=" + tableName);
 		return null;
 	}
+
+	@Override
+	public PO getPO(String tableName, String uuID, String trxName)
+	{
+		PO po = getPO(tableName, 0, trxName);
+		if (po != null)
+			po.loadByUU(uuID, trxName);
+		return po;
+	}
+
 }
--- a/org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
+++ b/org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
@@ -323,13 +323,4 @@ public class DefaultModelFactory implements IModelFactory {
 		return null;
 	}
 
-	@Override
-	public PO getPO(String tableName, String uuID, String trxName)
-	{
-		PO po = getPO(tableName, 0, trxName);
-		if (po != null)
-			po.loadByUU(uuID, trxName);
-		return po;
-	}
-
 }
--- a/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
+++ b/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
@@ -46,4 +46,14 @@ public interface IModelFactory {
 	 *	@return PO for Record or null
 	 */
 	public PO getPO (String tableName, ResultSet rs, String trxName);
+
+	/**
+	 * Get PO Class Instance
+	 * 
+	 * @param tableName
+	 * @param uuID      record
+	 * @param trxName   transaction
+	 * @return PO for Record or null
+	 */
+	public PO getPO(String tableName, String uuID, String trxName);
 }
--- a/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
+++ b/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
@@ -46,5 +46,4 @@ public interface IModelFactory {
 	 *	@return PO for Record or null
 	 */
 	public PO getPO (String tableName, ResultSet rs, String trxName);
-
 }
--- a/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
+++ b/org.adempiere.base/src/org/adempiere/base/IModelFactory.java
@@ -47,13 +47,4 @@ public interface IModelFactory {
 	 */
 	public PO getPO (String tableName, ResultSet rs, String trxName);
 
-	/**
-	 * Get PO Class Instance
-	 * 
-	 * @param tableName
-	 * @param uuID      record
-	 * @param trxName   transaction
-	 * @return PO for Record or null
-	 */
-	public PO getPO(String tableName, String uuID, String trxName);
 }
--- a/org.adempiere.base/src/org/adempiere/model/GenericPO.java
+++ b/org.adempiere.base/src/org/adempiere/model/GenericPO.java
@@ -82,16 +82,6 @@ public class GenericPO extends PO {
 	}
 
 	/**
-	 * @param tableName
-	 * @param ctx
-	 * @param uuID 
-	 * @param trxName
-	 */
-	public GenericPO(String tableName, Properties ctx, String uuID, String trxName) {
-		super(new PropertiesWrapper(ctx, tableName), 0, trxName, null, uuID);
-	}
-
-	/**
 	 * We must not use variable initializer here since the 2 variable below will be initialize inside
 	 * the initPO method called by the parent constructor.  
 	 */
--- a/org.adempiere.base/src/org/adempiere/model/GenericPO.java
+++ b/org.adempiere.base/src/org/adempiere/model/GenericPO.java
@@ -82,6 +82,16 @@ public class GenericPO extends PO {
 	}
 
 	/**
+	 * @param tableName
+	 * @param ctx
+	 * @param uuID 
+	 * @param trxName
+	 */
+	public GenericPO(String tableName, Properties ctx, String uuID, String trxName) {
+		super(new PropertiesWrapper(ctx, tableName), 0, trxName, null, uuID);
+	}
+
+	/**
 	 * We must not use variable initializer here since the 2 variable below will be initialize inside
 	 * the initPO method called by the parent constructor.  
 	 */
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -67,7 +67,7 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -4037560922339534982L;
+	private static final long serialVersionUID = 2951110137945905890L;
 
 	public final static int MAX_OFFICIAL_ID = 999999;
 
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -611,8 +611,7 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 	public PO getPOByUU (String uuID, String trxName)
 	{
 		PO po = getPO(0, trxName);
-		if (po != null)
-			po.loadByUU(uuID, trxName);
+		po.loadByUU(uuID, trxName);
 
 		return po;
 	} // getPOByUU
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -608,50 +608,14 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 	 * @param  trxName transaction
 	 * @return         PO for Record or null
 	 */
-	public PO getPOByUU(String uuID, String trxName)
+	public PO getPOByUU (String uuID, String trxName)
 	{
-		String tableName = getTableName();
-
-		PO po = null;
-		IServiceReferenceHolder<IModelFactory> cache = s_modelFactoryCache.get(tableName);
-		if (cache != null)
-		{
-			IModelFactory service = cache.getService();
-			if (service != null)
-			{
-				po = service.getPO(tableName, uuID, trxName);
-				if (po != null)
-					return po;
-			}
-			s_modelFactoryCache.remove(tableName);
-		}
-
-		List<IServiceReferenceHolder<IModelFactory>> factoryList = Service.locator().list(IModelFactory.class).getServiceReferences();
-		if (factoryList != null)
-		{
-			for (IServiceReferenceHolder<IModelFactory> factory : factoryList)
-			{
-				IModelFactory service = factory.getService();
-				if (service != null)
-				{
-					po = service.getPO(tableName, uuID, trxName);
-					if (po != null)
-					{
-						s_modelFactoryCache.put(tableName, factory);
-						break;
-					}
-				}
-			}
-		}
-
-		if (po == null)
-		{
-			// TODO have to check working.
-			po = new GenericPO(tableName, getCtx(), 0, trxName);
-		}
+		PO po = getPO(0, trxName);
+		if (po != null)
+			po.loadByUU(uuID, trxName);
 
 		return po;
-	} // getPOByUUID
+	} // getPOByUU
 
 	/**
 	 * 	Get PO Class Instance
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -606,7 +606,7 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 	 * 
 	 * @param  uuID    UUID
 	 * @param  trxName transaction
-	 * @return         PO for Record or null
+	 * @return         PO for Record
 	 */
 	public PO getPOByUU (String uuID, String trxName)
 	{
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -602,6 +602,58 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 	}	//	getPO
 
 	/**
+	 * Get PO Class Instance
+	 * 
+	 * @param  uuID    UUID
+	 * @param  trxName transaction
+	 * @return         PO for Record or null
+	 */
+	public PO getPOByUU(String uuID, String trxName)
+	{
+		String tableName = getTableName();
+
+		PO po = null;
+		IServiceReferenceHolder<IModelFactory> cache = s_modelFactoryCache.get(tableName);
+		if (cache != null)
+		{
+			IModelFactory service = cache.getService();
+			if (service != null)
+			{
+				po = service.getPO(tableName, uuID, trxName);
+				if (po != null)
+					return po;
+			}
+			s_modelFactoryCache.remove(tableName);
+		}
+
+		List<IServiceReferenceHolder<IModelFactory>> factoryList = Service.locator().list(IModelFactory.class).getServiceReferences();
+		if (factoryList != null)
+		{
+			for (IServiceReferenceHolder<IModelFactory> factory : factoryList)
+			{
+				IModelFactory service = factory.getService();
+				if (service != null)
+				{
+					po = service.getPO(tableName, uuID, trxName);
+					if (po != null)
+					{
+						s_modelFactoryCache.put(tableName, factory);
+						break;
+					}
+				}
+			}
+		}
+
+		if (po == null)
+		{
+			// TODO have to check working.
+			po = new GenericPO(tableName, getCtx(), 0, trxName);
+		}
+
+		return po;
+	} // getPOByUUID
+
+	/**
 	 * 	Get PO Class Instance
 	 *	@param whereClause where clause
 	 *	@param trxName transaction
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -320,9 +320,6 @@ public abstract class PO
 	/** Immutable flag **/
 	private boolean m_isImmutable = false;
 
-	/** Record UUID **/
-	private String m_uuid = null;
-
 	/** Access Level S__ 100	4	System info			*/
 	public static final int ACCESSLEVEL_SYSTEM = 4;
 	/** Access Level _C_ 010	2	Client info			*/
@@ -1385,14 +1382,15 @@ public abstract class PO
 	{
 		// reset new values
 		m_newValues = new Object[get_ColumnCount()];
-		m_uuid = uuID;
 		checkImmutable();
 
 		if (log.isLoggable(Level.FINEST))
 			log.finest("uuID=" + uuID);
-		if (!Util.isEmpty(m_uuid, true))
+		if (!Util.isEmpty(uuID, true))
 		{
-			load(trxName);
+			load(uuID,trxName);
+		}else {
+			load(0,trxName);
 		}
 	} // loadByUU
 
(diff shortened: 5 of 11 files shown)


org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
```diff
@@ -322,4 +322,14 @@ public PO getPO(String tableName, ResultSet rs, String trxName) {
 			s_log.log(Level.SEVERE, "(rs) - Not found - Table=" + tableName);
 		return null;
 	}
+
+	@Override
+	public PO getPO(String tableName, String uuID, String trxName)
```
hengsin, line 327: this should be moved to the interface as a default method. very unlikely that any model factory implementation will need to override this.

org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -601,6 +601,58 @@ public PO getPO (ResultSet rs, String trxName)
...
+
+		if (po == null)
+		{
+			// TODO have to check working.
+			po = new GenericPO(tableName, getCtx(), 0, trxName);
+		}
+
+		return po;
+	} // getPOByUUID
+
```
hengsin, line 655: this implementation plus the addition of new method to IModelFactory interface is actually pointless - it is just a unnecessary long way to reach the same necessary path of MTable.getPO(0, trxName) > po.loadByUU(uuid, trxName)

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -290,6 +319,10 @@ protected void copyPO(PO copy)
 	
 	/** Immutable flag **/
 	private boolean m_isImmutable = false;
+	
+	private String m_uuid = null;
+	/** UU flag **/
+	private boolean m_isUUID = false;
```
hengsin, line 325: should avoid adding another new flag here - increase potential for side effect and complicate po state management.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,6 +1376,35 @@ protected void load (int ID, String trxName)
...
+		{
+			setKeyInfo();
+			load(trxName);
+		}
+		else
+		{
+			loadDefaults();
+			m_createNew = true;
+			setKeyInfo();
+			loadComplete(true);
```
hengsin, line 1405: should refactor and share the code instead of just duplicate this else part from the "load (int ID, String trxName)" method.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1378,17 +1440,24 @@ public boolean load (String trxName)
 		try
 		{
 			pstmt = DB.prepareStatement(sql.toString(), m_trxName);	//	local trx only
-			for (int i = 0; i < m_IDs.length; i++)
+			if(m_isUUID)
```
hengsin, line 1443: don't need the flag here, just pass in the uuid parameter and check that for null is simpler and safer

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -3173,6 +3253,18 @@ protected void saveNew_afterSetID()
 	public String get_WhereClause (boolean withValues)
 	{
 		StringBuilder sb = new StringBuilder();
+		
+		if (m_isUUID) {
```
hengsin, line 3257: same here, instead of flag, pass in the uuid param is simpler and safer.

org.adempiere.base/src/org/idempiere/model/IMappedModelFactory.java
```diff
@@ -46,7 +46,8 @@
 	 */
 	void addMapping(String tableName, Supplier<Class<?>> classSupplier,
 			BiFunction<Integer, String, ? extends PO> recordIdFunction,
-			BiFunction<ResultSet, String, ? extends PO> resultSetFunction);
+			BiFunction<ResultSet, String, ? extends PO> resultSetFunction,
+			BiFunction<String, String, ? extends PO> recordUUFunction);
```
hengsin, line 50: the changes to IMappedModelFactory and the classes below is not needed as you have remove the uuid method from IModelFactory.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -291,6 +320,9 @@ protected void copyPO(PO copy)
 	/** Immutable flag **/
 	private boolean m_isImmutable = false;
 
+	/** Record UUID **/
+	private String m_uuid = null;
```
hengsin, line 324: I think it should be better if we remove this instance variable as well. the uuid can be parameter that's being pass around. If you are keeping this, you will have to ensure it always has the right uuid value, otherwise the PO state can be inconsistent at time.

dpansheriya:
@hengsin I fixed as per your suggestion. Please review and appreciate your feedback.

org.adempiere.base/src/org/idempiere/model/IMappedModelFactory.java
dpansheriya: @hengsin good catch, I missed it. Let me correct it.

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: @hengsin I tried to do this but looks get issue in get_WhereClause method as it invoked from many places and we have to pass null arguments on most of place. I think as currently the way it used without having getters and setters for m_uuid, we don't needs to maintain state of it. If we try to load value in it after reading record from DB or from RS, it may create more confusion as then after record always loaded by UU even ID present.

hengsin: I don't see why get_WhereClause is a problem, with get_WhereClause (boolean withValues, string uuid) and get_WhereClause (boolean withValues) { return get_WhereClause (boolean withValues, null);} should be sufficient maintain compatibility with existing code. Having instance variable in inconsistent state is always a potential source of unwanted side effect in future.

org.adempiere.base/src/org/idempiere/model/MappedModelFactory.java
```diff
@@ -44,6 +44,7 @@
 public class MappedModelFactory implements IModelFactory, IMappedModelFactory {
 
 	private final ConcurrentHashMap<String, Supplier<Class<?>>> classMap = new ConcurrentHashMap<>();
+	private final ConcurrentHashMap<String, BiFunction<String, String, ? extends PO>> recordUUMap = new ConcurrentHashMap<>();
```
hengsin, line 47: this should be removed

org.adempiere.base/src/org/idempiere/model/MappedModelFactory.java
```diff
@@ -73,16 +74,19 @@ public PO getPO(String tableName, ResultSet rs, String trxName) {
...
 	}
 	
 	@Override
 	public void removeMapping(String tableName) {
 		classMap.remove(tableName);
 		recordIdMap.remove(tableName);
 		resultSetMap.remove(tableName);
+		recordUUMap.remove(tableName);
 	}
+
```
hengsin, line 91: please revert the changes to this class

org.idempiere.test/src/org/idempiere/test/model/MappedModelFactoryTest.java
```diff
@@ -65,8 +65,10 @@ public MappedModelFactoryTest() {
 	@Order(1)
 	public void testDefaultMappedModelFactory() {
 		IMappedModelFactory mappedFactory = Core.getMappedModelFactory();
-		mappedFactory.addMapping(MyTest.Table_Name, () -> MyTest.class, (id, trxName) -> new MyTest(Env.getCtx(), id, trxName), 
-				(rs, trxName) -> new MyTest(Env.getCtx(), rs, trxName));		
+		mappedFactory.addMapping(MyTest.Table_Name, () -> MyTest.class, 
+		                         (id, trxName) -> new MyTest(Env.getCtx(), id, trxName), 
+		                         (rs, trxName) -> new MyTest(Env.getCtx(), rs, trxName), 
+		                         (uuID, trxName) -> new MyTest(Env.getCtx(), uuID, trxName));
```
hengsin, line 71: please revert the changes to this class

hengsin:
would be good if unit testing can be added to org.idempiere.test to verify the new method added to genericpo, po and mtable is working as expected.

org.adempiere.base/src/org/adempiere/base/IModelFactory.java
```diff
@@ -46,4 +46,5 @@
 	 *	@return PO for Record or null
 	 */
 	public PO getPO (String tableName, ResultSet rs, String trxName);
+
```
hengsin, line 49: would be good to remove this unnecessary white space only changes (same for the file above)

org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -601,6 +601,22 @@ public PO getPO (ResultSet rs, String trxName)
...
+	 * Get PO Class Instance
+	 * 
+	 * @param  uuID    UUID
+	 * @param  trxName transaction
+	 * @return         PO for Record or null
+	 */
+	public PO getPOByUU (String uuID, String trxName)
+	{
+		PO po = getPO(0, trxName);
+		if (po != null)
```
hengsin, line 614: don't think you can ever get po==null here.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,6 +1375,26 @@ protected void load (int ID, String trxName)
...
+		m_uuid = uuID;
+		checkImmutable();
+
+		if (log.isLoggable(Level.FINEST))
+			log.finest("uuID=" + uuID);
+		if (!Util.isEmpty(m_uuid, true))
+		{
+			load(trxName);
+		}
+	} // loadByUU
```
hengsin, line 1397: if the definition here is if uuid==null, it should create a new PO record, it should be simplify to just if (Util.isEmpty(uuid, true)) load(0, trxName); else load(uuid, trxName);

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1569,6 +1628,17 @@ else if (clazz == String.class)
...
 
+	/**
+	 * Load record with UUID
+	 * @param uuID - UUID
+	 * @param trxName - Transaction name
+	 */
+	protected void load(String uuID, String trxName)
+	{
+		load(0, trxName);
+		loadByUU(uuID, trxName);
```
hengsin, line 1639: would prefer the following instead, change the original load(trxName) method to load(String uuid, String trxName) and add load(String trxName) {load(null, trxName);}. Do the empty uuid check above in the rename load(uuid, trxname) method.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -3173,6 +3243,18 @@ protected void saveNew_afterSetID()
...
+		{
+			sb.append(getUUIDColumnName()).append("=");
+			if (withValues)
+				sb.append(m_uuid);
+			else
+				sb.append("?");
+
+			return sb.toString();
+		}
+
```
hengsin, line 3257: suggestion: rename this to get_whereclause(string uuid, boolean withValues). add get_whereclause(boolean withvalues) {return get_whereclause(null, withvalues);}

org.idempiere.test/src/org/idempiere/test/model/MappedModelFactoryTest.java
```diff
@@ -65,8 +65,9 @@ public MappedModelFactoryTest() {
 	@Order(1)
 	public void testDefaultMappedModelFactory() {
 		IMappedModelFactory mappedFactory = Core.getMappedModelFactory();
-		mappedFactory.addMapping(MyTest.Table_Name, () -> MyTest.class, (id, trxName) -> new MyTest(Env.getCtx(), id, trxName), 
-				(rs, trxName) -> new MyTest(Env.getCtx(), rs, trxName));		
+		mappedFactory.addMapping(MyTest.Table_Name, () -> MyTest.class, 
+		                         (id, trxName) -> new MyTest(Env.getCtx(), id, trxName), 
+		                         (rs, trxName) -> new MyTest(Env.getCtx(), rs, trxName));
```
hengsin, line 70: please revert changes to this class too

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: @hengsin ohk Let me do this.

org.idempiere.test/src/org/idempiere/test/model/MappedModelFactoryTest.java
dpansheriya: @hengsin I tried resetting but not getting any changes uncommitted.

org.adempiere.base/src/org/compiere/model/MTable.java
dpansheriya: Agree

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: After refactoring, this method no longer required.

org.idempiere.test/src/org/idempiere/test/model/MappedModelFactoryTest.java
dpansheriya: This is done in latest commit

dpansheriya: This is done in latest commit

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: In refactored code, this is handled.

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: This is done in latest commit

hengsin:
@CarlosRuiz-globalqss, do  you think is necessary to re-generate all model classes for this ticket ?

CarlosRuiz-globalqss:
> @CarlosRuiz-globalqss, do you think is necessary to re-generate all model classes for this ticket ?

Hi @hengsin - my comments in ticket

hengsin:
> > @CarlosRuiz-globalqss, do you think is necessary to re-generate all model classes for this ticket ?
> 
> Hi @hengsin - my comments in ticket

Sorry, can't find your comments there that's related to the question I ask above.

CarlosRuiz-globalqss:
redacting ... sorry for the delay

hengsin:
hi @dpansheriya , as per Carlos's comments in the ticket, please remove allthe uuid constructor related changes from PO and X_Test.

dpansheriya:
> hi @dpansheriya , as per Carlos's comments in the ticket, please remove all the uuid constructor related changes from PO and X_Test.

Sure, I am doing it. 
I am keeping getPOByUU as I am thinking to use return getPO(PO.getUUIDColumnName(getTableName())+"="+uuID, trxName); is not good as query is not parameterize.

org.adempiere.base/src/org/adempiere/model/GenericPO.java
```diff
@@ -81,6 +81,16 @@ public GenericPO(String tableName, Properties ctx, ResultSet rs, String trxName)
...
+	/**
+	 * @param tableName
+	 * @param ctx
+	 * @param uuID 
+	 * @param trxName
+	 */
+	public GenericPO(String tableName, Properties ctx, String uuID, String trxName) {
+		super(new PropertiesWrapper(ctx, tableName), 0, trxName, null, uuID);
+	}
+
```
hengsin, line 93: this constructor should be removed as well

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -190,6 +190,20 @@ public PO (Properties ctx, ResultSet rs, String trxName)
...
+	/**
+	 * Create & Load existing Persistent Object.
+	 * 
+	 * @param ctx     context
+	 * @param ID      the ID if 0, the record defaults are applied - ignored if re exists
+	 * @param trxName transaction name
+	 * @param rs      optional - load from current result set position (no navigation, not closed)
+	 * @param uuID    optional - load from uuID
+	 */
+	public PO(Properties ctx, int ID, String trxName, ResultSet rs, String uuID)
```
hengsin, line 206: this constructor should be removed as well

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,13 +1359,44 @@ protected void load (int ID, String trxName)
...
+
+		if (log.isLoggable(Level.FINEST))
+			log.finest("uuID=" + uuID);
+		if (!Util.isEmpty(uuID, true))
+		{
+			load(uuID,trxName);
+		}else {
+			load(0,trxName);
+		}
+	} // loadByUU
```
hengsin, line 1382: without the uuid constructor requirement, we just need to keep one of loadByUU(uuid, trxName) or load(uuid, trxName) (if we keep both, load(uuid, trxName) should be change to private or protected). and to keep things simple, for loadByUU(uuid, trxName), just return null if uuid is null or empty.

org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -601,6 +601,21 @@ public PO getPO (ResultSet rs, String trxName)
 		return po;
 	}	//	getPO
 
+	/**
+	 * Get PO Class Instance
+	 * 
+	 * @param  uuID    UUID
+	 * @param  trxName transaction
+	 * @return         PO for Record or null
```
hengsin, line 609: this javadoc is wrong, it will never return null

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Shouldn't we allow null UU to create new record?

hengsin: No, we don't need that  since the intention here is not to use uuid as primary key. The existing approach of id=0 should continue to be used to create new record.

org.adempiere.base/src/org/compiere/model/MTable.java
dpansheriya: @hengsin  you may referencing old commit as I did remove "or null" in commit 408fe45 (Latest one)

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: This also done in latest commit.

org.adempiere.base/src/org/adempiere/model/GenericPO.java
dpansheriya: Done

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Done

org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -601,6 +601,21 @@ public PO getPO (ResultSet rs, String trxName)
...
+	 * @return         PO for Record
+	 */
+	public PO getPOByUU (String uuID, String trxName)
+	{
+		PO po = getPO(0, trxName);
+		po.loadByUU(uuID, trxName);
+
+		return po;
+	} // getPOByUU
+
```
hengsin, line 618: Since we added new method to MTable, we need to update the serialVersionUID of MTable.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,13 +1343,44 @@ protected void load (int ID, String trxName)
...
+		m_newValues = new Object[get_ColumnCount()];
+		checkImmutable();
+
+		if (log.isLoggable(Level.FINEST))
+			log.finest("uuID=" + uuID);
+		if (!Util.isEmpty(uuID, true))
+		{
+			load(uuID,trxName);
+		}else {
+			throw new IllegalArgumentException("Invalid null UU - Must pass valid UU");
```
hengsin, line 1364: the null check should be at the top. since we are throwing illegalargumentexception here, the method shouldn't do anything if uuid is null/empty.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,13 +1343,44 @@ protected void load (int ID, String trxName)
...
 	 *  @param trxName transaction
 	 *  @return true if loaded
 	 */
-	public boolean load (String trxName)
+	public boolean load (String trxName) {
+		return load(null, trxName);
+	}
+	
+	/**
+	 *  (re)Load record with m_ID[*]
```
hengsin, line 1378: the javadoc should be updated to reflect the addition of uuID option.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -3164,15 +3202,37 @@ protected void saveNew_afterSetID()
...
+	 * 	@return where clause
+	 */
+	public String get_WhereClause (boolean withValues) {
+		return get_WhereClause(withValues,null);
+	}
 
 	/**
 	 * 	Create Single/Multi Key Where Clause
 	 * 	@param withValues if true uses actual values otherwise ?
+	 *  @param uuID RecordUU
```
hengsin, line 3218: javadoc should be updated to reflect the addtion of uuid option.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -3164,15 +3202,37 @@ protected void saveNew_afterSetID()
...
-	public String get_WhereClause (boolean withValues)
+	public String get_WhereClause (boolean withValues, String uuID)
 	{
 		StringBuilder sb = new StringBuilder();
+
+		if (!Util.isEmpty(uuID, true))
+		{
+			sb.append(getUUIDColumnName()).append("=");
+			if (withValues)
+				sb.append(uuID);
```
hengsin, line 3229: this is not correct - you need to  quote the uuID string value, i.e DB.TO_STRING(uuID)

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: It is already there. Please note added line is for uuID paramater.

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Done

org.adempiere.base/src/org/compiere/model/MTable.java
dpansheriya: Done

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Fixed

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1343,13 +1343,44 @@ protected void load (int ID, String trxName)
...
+	 */
+	public void loadByUU(String uuID, String trxName)
+	{
+		// reset new values
+		m_newValues = new Object[get_ColumnCount()];
+		checkImmutable();
+
+		if (log.isLoggable(Level.FINEST))
+			log.finest("uuID=" + uuID);
+		if (Util.isEmpty(uuID, true))
```
hengsin, line 1360: sorry, I means this should be at top, i.e comes before the "//reset new values" line.

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Done

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: Done

hengsin (approved):
looks good.

PR #581, opened 2021-02-12
IDEMPIERE-4694 Implement thread safe, annotation based osgi event han…

hengsin:
…dling

Fix testFactValidate unit test fail with seed db due to No Cost error
for Azalea Bush


CarlosRuiz-globalqss (approved):
looks good

PR #583, opened 2021-02-12
IDEMPIERE 4702 Messagebox enhancement: Esc key closes Dialog with cancel action

Soundbytes:
https://idempiere.atlassian.net/browse/IDEMPIERE-4702

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
@@ -29,6 +29,7 @@ import org.adempiere.webui.factory.ButtonFactory;
 import org.adempiere.webui.session.SessionManager;
 import org.adempiere.webui.theme.ThemeManager;
 import org.adempiere.webui.util.ZKUpdateUtil;
+import org.compiere.util.CLogger;
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
 import org.compiere.util.Util;
@@ -38,6 +39,7 @@ import org.zkoss.zk.ui.Page;
 import org.zkoss.zk.ui.event.Event;
 import org.zkoss.zk.ui.event.EventListener;
 import org.zkoss.zk.ui.event.Events;
+import org.zkoss.zk.ui.select.annotation.Listen;
 import org.zkoss.zul.Hbox;
 import org.zkoss.zul.Image;
 import org.zkoss.zul.Vbox;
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -55,6 +57,8 @@
 	 * 
 	 */
 	private static final long serialVersionUID = 8928526331932742124L;
+	
+	private static final CLogger log = CLogger.getCLogger(Messagebox.class);
```
CarlosRuiz-globalqss, line 61: This line is creating an eclipse warning - if not used is better to remove it

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -387,15 +399,19 @@ else if (event.getTarget() == btnIgnore)
 		{
 			returnValue = IGNORE;
 		}
-
+		close();
+	}
+	
+	private void close() {
+		SessionManager.getSessionApplication().getKeylistener().removeEventListener(Events.ON_CTRL_KEY, this);
```
CarlosRuiz-globalqss, line 406: Just a question - why is this line required?  Is zk adding the listener internally when you add the annotation?

CarlosRuiz-globalqss (changes requested):
Hi @Soundbytes, the pull request must be done against a branch with name IDEMPIERE-4702.

If you create the pull request over master it can become a problem very soon to resolve conflicts and merging.

I tested this successfully, can you please do the suggested change?

Regards,

Carlos Ruiz

CarlosRuiz-globalqss:
Also, in the commit message please add the link IDEMPIERE-4702  (with dash) so JIRA can connect the pull request.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
Soundbytes: That was left over from debugging - will be removed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
Soundbytes: Oops. I first tried a different approach using a the KeyListener. That worked equally well but was a bit more complicated so I removed it. Except for this line which I just had forgotten. - will be removed

Soundbytes:
> Hi @Soundbytes, the pull request must be done against a branch with name IDEMPIERE-4702.

It is not possible to change the source branch after the PR is publlished. I am replacing this request with a new one.

PR #585, opened 2021-02-14
IDEMPIERE-4704 Add base interface and class that uses generic type an…

hengsin:
…d lambda function for osgi factory that create new instance by class name or a string key

https://idempiere.atlassian.net/browse/IDEMPIERE-4704


CarlosRuiz-globalqss (approved):
unit tests passed

PR #586, opened 2021-02-14
IDEMPIERE-4702 Messagebox enhancement: Esc key closes Dialog with cancel

Soundbytes:
This replaces the retracted PR of the same name.  https://github.com/idempiere/idempiere/pull/583
(Original PR used the wrong source branch and can not be fixed)

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
@@ -38,6 +38,7 @@ import org.zkoss.zk.ui.Page;
 import org.zkoss.zk.ui.event.Event;
 import org.zkoss.zk.ui.event.EventListener;
 import org.zkoss.zk.ui.event.Events;
+import org.zkoss.zk.ui.select.annotation.Listen;
 import org.zkoss.zul.Hbox;
 import org.zkoss.zul.Image;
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
This test passes successfully:

- navigate to _Invoice (Customer)_ 200002 in GardenWorld
- click on _Posted_ button to open the _Posting_ window
- click on the _Re-Post_ button
- now this dialog can be closed with _Esc_ keystroke assuming the _Cancel_ button was pushed


Question for @hengsin - are you OK with the annotations approach?  Or is there a better way to implement it?

Regards,

Carlos Ruiz

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -353,6 +355,15 @@ public static int showDialog(String message, String title, int buttons, String i
 		Messagebox msg = new Messagebox();
 		return msg.show(message, title, buttons, icon, editor, callback, modal);
 	}
+	
+    // Andreas Sumerauer IDEMPIERE 4702
+	@Listen("onCancel = #form")
```
hengsin, line 360: Should be just @Listen("onCancel") here, MessageBox doesn't has an element with id="form".

hengsin (changes requested):

hengsin (approved):
tested

PR #587, opened 2021-02-15
IDEMPIERE-4690 Add column callout factory base class backed by Map an…

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4690

Refactoring. Add unit test for gridtab and callout.


CarlosRuiz-globalqss (approved):
test case passed

PR #590, opened 2021-02-16
IDEMPIERE-4358 Rename table issue

CarlosRuiz-globalqss:
Rename the table also in whereclause and orderbyclause in AD_Tab and AD_Ref_Table

https://idempiere.atlassian.net/browse/IDEMPIERE-4358


CarlosRuiz-globalqss:
Test case is:
- create a new table with columns (for example copying from M_FreightCategory)
- synchronize columns (to create the table)
- run the "Create window from table" process to create window, menu, etc
- then rename the table
- the window just created becomes unusable because the order by clause references the original name

d-ruiz (changes requested):
@CarlosRuiz-globalqss, I found a problem with the process, you can see the comments in the ticket.

d-ruiz (approved):
Tested succesfully

PR #593, opened 2021-02-17
IDEMPIERE-3101 implement OAuth2 for mail (gmail, outlook and other mail system)

CarlosRuiz-globalqss:
WIP - initial version working on zk with google apps mail using OAuth2 Authorization Code Flow and Client Type = web application

https://idempiere.atlassian.net/browse/IDEMPIERE-3101


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/AddAuthorizationMailForm.java
```diff
@@ -0,0 +1,305 @@
...
+import org.zkoss.zul.Grid;
+import org.zkoss.zul.Separator;
+import org.zkoss.zul.Space;
+import org.zkoss.zul.Timer;
+
+/**
+ * @author hengsin
+ *
+ */
+public class AddAuthorizationMailForm extends ADForm {
```
hengsin, line 57: this should be a generic AddAuthorizationForm (or at least has a base class of it) and scope pass in as parameter to the form. It will work the same for most (if not all) scope.

hengsin:
Looks fine but it will be much better if we can move most (if not all) of the new lib dependency to target platform instead (some of the lib is probably already part of the target platform).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/AddAuthorizationMailForm.java
CarlosRuiz-globalqss: Ready, I added code to use the scope as parameter +SCOPE passed from menu according to IDEMPIERE-4713

CarlosRuiz-globalqss:
> Looks fine but it will be much better if we can move most (if not all) of the new lib dependency to target platform instead (some of the lib is probably already part of the target platform).

OK, I can try that.
So, here I need to generate a new targetplatform version?

According to @hieplq instructions here?
https://wiki.idempiere.org/en/Building_iDempiere_by_tycho#Add_more_bundle_to_p2_maven_repository_of_idempiere

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/AddAuthorizationForm.java
```diff
@@ -0,0 +1,313 @@
...
+
+	/**
+	 * Trigger action when highlighted
+	 */
+	@Override
+	public void doHighlighted() {
+		super.doHighlighted();
+		parameter_scope = Env.getContext(Env.getCtx(), getWindowNo(), "+SCOPE");
+		if (Util.isEmpty(parameter_scope)) {
+			parameter_scope = MAuthorizationAccount.AD_AUTHORIZATIONSCOPE_EMail;
```
hengsin, line 308: I think the logical things to do here if +SCOPE is missing is to provide a dropdown list of scope available for user to pick.

hieplq:
> > Looks fine but it will be much better if we can move most (if not all) of the new lib dependency to target platform instead (some of the lib is probably already part of the target platform).
> 
> OK, I can try that.
> So, here I need to generate a new targetplatform version?
> 
> According to @hieplq instructions here?
> https://wiki.idempiere.org/en/Building_iDempiere_by_tycho#Add_more_bundle_to_p2_maven_repository_of_idempiere

hi @CarlosRuiz-globalqss I update guideline for [update target platform](https://wiki.idempiere.org/en/Building_iDempiere_by_tycho#update_.5Bidempiere-source.5D.2Forg.idempiere.p2.targetplatform.2Forg.idempiere.p2.targetplatform.target)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/AddAuthorizationForm.java
CarlosRuiz-globalqss: Thanks @hengsin - I did the suggested change, if the +SCOPE is not passed, then a list of valid scopes is shown.

org.adempiere.server/src/main/servlet/org/idempiere/web/OAuthCodeCallbackHandlerServlet.java
```diff
@@ -0,0 +1,108 @@
...
+					break;
+				}
+			}
+			if (credential == null || credential.get_ID() <= 0) {
+				errmsg = "Invalid \"state\" URL parameter, not found";
+			}
+		}
+
+		if (errmsg == null) {
+			Env.getCtx().setProperty(Env.AD_CLIENT_ID, String.valueOf(pinstance.getAD_Client_ID())); // To avoid Context Lost exception
```
hengsin, line 88: the standard approach to do this is create a new properties, use Env.setContext to populate the new properties instance with ad_client_id, etc and then 

try {
   ServerContext.setCurrentInstance(properties);
   //do your stuff
} finally {
   ServerContext.dispose();
}

CarlosRuiz-globalqss: Thanks @hengsin - changed and tested

d-ruiz:
Comments added to the ticket

org.adempiere.base/src/org/compiere/util/EMail.java
```diff
@@ -280,8 +290,15 @@ public String send ()
 			{
 				props.put("mail.smtp.starttls.enable", "true");
 			}
-
-			session = Session.getInstance(props, m_auth);
+			if (isOAuth2 && m_auth != null) {
```
d-ruiz, line 293: My 2 cents here, I think the validation `&& m_auth != null` is worthless at this point, as based on the code above isOAuth2 will never be true if m_auth is null. I understand it is defensive programming, but I think it could be abstracted more by creating a _userName_ variable (after all m_auth.getPasswordAuthentication().getUserName() is called twice in a scope of 20 lines. 

in that way, the m_auth != null would appear just once 
String userName = m_auth != null ? m_auth.getPasswordAuthentication().getUserName() : "";

I think both methods that receive the userName handle the null/empty value properly. 

But above the improvement of code cleanliness. I got the doubt if, in this way OAuth does not work with the SMTP flag off at all, AFAIK Microsoft can work without SMTP, but I am not sure. 

But by extracting the userName assignation, the userName could come from different sources in the future, (EmailUser, RequestUsedr, etc...), instead of making it depend on EMailAuthenticator. 

Anyway, just a thought I had by reviewing the code ... If you don't find it worthy, it is not a blocker.

CarlosRuiz-globalqss: Thanks @d-ruiz - I dropped that part of the if - I added it as defensive programming, but as you said is related to IDEMPIERE-4723 I cleaned that piece.

About the flag, the mechanism configured in AD_Client is all related to SMTP, there is no prevision in iDempiere to sent email with a different protocol, just SMTP.

The two flags there are for "SMTP Authentication" and "SMTP SSL/TLS" - the unchecked SMTP Authentication is when a server doesn't require a user/name to allow sending emails.  That's very rare I think, maybe in corporate servers that would be allowed (but is discouraged).  AFAIK none of the known mail provider allows that because it would be a big source of spam without control.

When no authentication is set (again, discouraged but possible), then no need to pass a username or password at all - OAuth or not OAuth.

PR #595, opened 2021-02-18
IDEMPIERE-4704 Add base interface and class that uses generic type an…

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4704

Fix unit test error when running through maven/jenkin (mvn verify
-DskipTests=false)


CarlosRuiz-globalqss (approved):
tested in jenkins server

PR #597, opened 2021-02-19
IDEMPIERE-4710 : NPE in ProcessInfo.getLastServerRebootDate()

nmicoud:


CarlosRuiz-globalqss:
As said in ticket:

"that sounds like a symptom that the system is not running correctly as creating that session record must be the first thing it happens on AdempiereServerMgr.startEnvironment"

hengsin (approved):
trivial change, good to go.

PR #598, opened 2021-02-19
IDEMPIERE-4712 : set MInvoice setReversal(boolean) method public

nmicoud:


hengsin (approved):
trivial change, good to go.

PR #600, opened 2021-02-19
IDEMPIERE-4713 Allow definition of context variables to inject on menu and window definition

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4713

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/Env.java
+++ b/org.adempiere.base/src/org/compiere/util/Env.java
@@ -2090,6 +2090,8 @@ public final class Env
 
 	/**	New Line 		 */
 	public static final String	NL = System.getProperty("line.separator");
+	/* Prefix for predefined context variables coming from menu or window definition */
+	private static final String PREFIX_PREDEFINED_VARIABLE = "+";
 
 
 	/**
@@ -2116,7 +2118,7 @@ public final class Env
 							) {
 							value = value.substring(1, value.length()-1);
 						}
-						Env.setContext(ctx, windowNo, var, value);
+						Env.setContext(ctx, windowNo, PREFIX_PREDEFINED_VARIABLE + var, value);
 					}
 				}
 			}
--- a/org.adempiere.base/src/org/compiere/util/Env.java
+++ b/org.adempiere.base/src/org/compiere/util/Env.java
@@ -2104,6 +2104,17 @@ public final class Env
 	}   //  static
 
 
+	/**
+	 * Add in context predefined variables with prefix +, coming from menu or window definition
+	 * Predefined variables must come separated by new lines in one of the formats:
+	 *   VAR=VALUE
+	 *   VAR="VALUE"
+	 *   VAR='VALUE'
+	 *  The + prefix is not required, is added here to the defined variables
+	 * @param ctx
+	 * @param windowNo
+	 * @param predefinedVariables
+	 */
 	public static void setPredefinedVariables(Properties ctx, int windowNo, String predefinedVariables) {
 		if (predefinedVariables != null) {
 			String[] lines = predefinedVariables.split("\n");
--- a/org.adempiere.base/src/org/compiere/util/Env.java
+++ b/org.adempiere.base/src/org/compiere/util/Env.java
@@ -2101,4 +2101,26 @@ public final class Env
 		getCtx().put(LANGUAGE, Language.getBaseAD_Language());
 	}   //  static
 
+
+	public static void setPredefinedVariables(Properties ctx, int windowNo, String predefinedVariables) {
+		if (predefinedVariables != null) {
+			String[] lines = predefinedVariables.split("\n");
+			for (String line : lines) {
+				int idxEq = line.indexOf("=");
+				if (idxEq > 0) {
+					String var = line.substring(0, idxEq).trim();
+					if (var.length() > 0) {
+						String value = line.substring(idxEq+1).trim();
+						if (   (value.startsWith("\"") && value.endsWith("\""))
+							|| (value.startsWith("'")  && value.endsWith("'") )
+							) {
+							value = value.substring(1, value.length()-1);
+						}
+						Env.setContext(ctx, windowNo, var, value);
+					}
+				}
+			}
+		}
+	}
+
 }   //  Env
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessDialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessDialog.java
@@ -116,6 +116,11 @@ public class ProcessDialog extends AbstractProcessDialog implements EventListene
 	private long prevKeyEventTime = 0;
 	private KeyEvent prevKeyEvent;
 
+	/**
+	 * Dialog to start a process/report
+	 * @param AD_Process_ID
+	 * @param isSOTrx
+	 */
 	public ProcessDialog (int AD_Process_ID, boolean isSOTrx)
 	{
 		this(AD_Process_ID, isSOTrx, null);
@@ -123,16 +128,9 @@ public class ProcessDialog extends AbstractProcessDialog implements EventListene
 	
 	/**
 	 * Dialog to start a process/report
-	 * @param ctx
-	 * @param parent
-	 * @param title
-	 * @param aProcess
-	 * @param WindowNo
 	 * @param AD_Process_ID
-	 * @param predefinedContextVariables 
-	 * @param tableId
-	 * @param recordId
-	 * @param autoStart
+	 * @param isSOTrx
+	 * @param predefinedContextVariables
 	 */
 	public ProcessDialog (int AD_Process_ID, boolean isSOTrx, String predefinedContextVariables)
 	{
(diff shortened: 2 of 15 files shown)


org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -2101,4 +2103,26 @@ public static int getZoomWindowID(int AD_Table_ID, int Record_ID, int windowNo)
 		getCtx().put(LANGUAGE, Language.getBaseAD_Language());
 	}   //  static
 
+
+	public static void setPredefinedVariables(Properties ctx, int windowNo, String predefinedVariables) {
+		if (predefinedVariables != null) {
+			String[] lines = predefinedVariables.split("\n");
```
hengsin, line 2109: this is split by newline ? I think comma or semi colon separated is easier for entry (would be nice if user can use either one of the split character).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/IDesktop.java
```diff
@@ -215,4 +215,7 @@
 	public void setTabTitle(String title, int windowNo);
 	
 	public void renderHomeTab();
+
+	default void setPredefinedContextVariables(String predefinedVariables) {
```
hengsin, line 219: don't think we have to add this to the interface. it is not a public api, just something use internally.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/AbstractDesktop.java
```diff
@@ -318,4 +329,13 @@ protected void showHighlighted(Window win, String position)
     	}
     }
 
+	@Override
+	public void setPredefinedContextVariables(String predefinedVariables) {
+		this.predefinedContextVariables = predefinedVariables;
+	}
+
+	public String getPredefinedContextVariables() {
```
hengsin, line 337: the get and set should be protected instead, it is just for internal use, i.e not public api

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
```diff
@@ -87,6 +88,7 @@ public ProcessDialog openProcessDialog(int processId, boolean soTrx) {
      */
 	public ADForm openForm(int formId) {
 		ADForm form = ADForm.openForm(formId);
+		Env.setPredefinedVariables(Env.getCtx(), form.getWindowNo(), getPredefinedContextVariables());
```
hengsin, line 91: shouldn't this happens before the openForm method and the form implementation class can get access to those variables in the openForm method ?

org.adempiere.base/src/org/compiere/util/Env.java
CarlosRuiz-globalqss: I thought about this but is very easy that comma or semi-colon is added inside a context variable, and to allow that we would need a parser that checks for quotation, and then a parser that checks for escape character (\\)
So, in principle with newline approach we just cannot add newlines within these context variables

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
CarlosRuiz-globalqss: I tried that, but not possible, we don't have the windowNo before that

CarlosRuiz-globalqss:
Thanks for peer review, pushed the requested changes, and answered the others.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
hengsin: perhaps we should just pass the context variables into the openform method ( and do the Env.setPredefinedVariables(Env.getCtx(), form.getWindowNo(), getPredefinedContextVariables()); in it ) ? I think it should be more straightforward to process the context variables in the form's initForm method.

CarlosRuiz-globalqss: Tried that [FailedPatch4713.txt](https://github.com/idempiere/idempiere/files/6021603/FailedPatch4713.txt)

Still is not possible, the form is open in Extensions.getForm - and to pass the parameters there we would need to change the IFormFactory to receive the parameter.

org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -2101,4 +2103,26 @@ public static int getZoomWindowID(int AD_Table_ID, int Record_ID, int windowNo)
 		getCtx().put(LANGUAGE, Language.getBaseAD_Language());
 	}   //  static
 
+
+	public static void setPredefinedVariables(Properties ctx, int windowNo, String predefinedVariables) {
```
hengsin, line 2107: please add javadoc

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessDialog.java
```diff
@@ -116,6 +116,11 @@
 	private long prevKeyEventTime = 0;
 	private KeyEvent prevKeyEvent;
 
+	public ProcessDialog (int AD_Process_ID, boolean isSOTrx)
```
hengsin, line 119: please add javadoc

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessDialog.java
```diff
@@ -124,16 +129,18 @@
 	 * @param aProcess
 	 * @param WindowNo
 	 * @param AD_Process_ID
+	 * @param predefinedContextVariables 
 	 * @param tableId
 	 * @param recordId
 	 * @param autoStart
 	 */
-	public ProcessDialog (int AD_Process_ID, boolean isSOTrx)
+	public ProcessDialog (int AD_Process_ID, boolean isSOTrx, String predefinedContextVariables)
```
hengsin, line 137: would be good if we can fix the javadoc above as well

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
hengsin: We don't have to change Extensions.getForm, we can add "Env.setPredefinedVariables(Env.getCtx(), form.getWindowNo(), predefinedContextVariables)"; prior to "form.init(adFormID, name);" (around line 180) and the context variables would be available when the ADForm.initForm() method is invoke.

CarlosRuiz-globalqss: OK, I'm doing the suggested change.

The problem was that I was testing with WResetPassword.

So, this approach will work fine for classes extending ADForm, but not for the forms that are developed implementing IFormController, those will require anyways something like the onPageAttached mechanism, but it works fine too, I tested it.

hengsin: It can works with WResetPassword too. Instead of form = new CustomForm(), you use form = new CustomForm() { protected void initForm() { ... } };

CarlosRuiz-globalqss:
Ready @hengsin - I also made some tests with Chuck's suggestions and found that clicking on New was not calling this, so I did some changes to implement that too.

hengsin (approved):
looks good

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/AbstractMenuPanel.java
```diff
@@ -313,6 +315,9 @@ private void onNewRecord(Treeitem selectedItem) {
         {
 			int menuId = Integer.parseInt((String)selectedItem.getValue());
 			MMenu menu = new MMenu(Env.getCtx(), menuId, null);
+			IDesktop desktop = SessionManager.getAppDesktop();
+			if (desktop instanceof AbstractDesktop)
+				((AbstractDesktop)desktop).setPredefinedContextVariables(menu.getPredefinedContextVariables());
```
hengsin, line 320: how does the PredefinedContextVariables at AD_Window and AD_Menu suppose to works together ?

CarlosRuiz-globalqss: The menu predefined variables are added to the context just when the window is opened from the menu, the window variables are added to the context always.
I have not specifically a use case for that - I made some tests for example making a window read-only when opened by zoom, but read-write with specific filters when opened from menu (as described in [the ticket](https://idempiere.atlassian.net/browse/IDEMPIERE-4713?focusedCommentId=46961))

PR #602, opened 2021-02-21
COF #11523 / IDEMPIERE-4715 - attribute set instance: new qty availab…

muriloht:
…le column and group results in select existing record dialog


hengsin (changes requested):
please remove all the unnecessary white space only changes.

CarlosRuiz-globalqss:
replaced with https://github.com/idempiere/idempiere/pull/670

PR #604, opened 2021-02-24
IDEMPIERE-4716 : set InfoPanel onDoubleClick method protected

nmicoud:


hengsin (approved):
trivial change, good to go

PR #605, opened 2021-02-26
IDEMPIERE-4718 Import Existing Maven Projects doesn't import the org.…

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4718

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
 		<module>org.idempiere.javadoc</module>
 		<module>org.idempiere.test-feature</module>
 		<module>org.idempiere.test</module>
-		<module>${idempiere.target}</module>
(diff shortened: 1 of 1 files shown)


hieplq:
purpose of that variable to help developer can separate maintain his target project without touch core

but i'm fine with this change because i never need to maintain separate target

CarlosRuiz-globalqss (approved):
looks OK.

The setting of this variable is done on org.idempiere.parent/pom.xml line 29, and there is another usage of the variable in same file line 538

Must that be changed too?

hengsin:
> looks OK.
> 
> The setting of this variable is done on org.idempiere.parent/pom.xml line 29, and there is another usage of the variable in same file line 538
> 
> Must that be changed too?

No, we don't have to change that.

PR #606, opened 2021-02-27
IDEMPIERE-4722 Remove constraint that idempiere source folder must be…

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4722

use project_loc of org.adempiere.base instead of workspace_loc


CarlosRuiz-globalqss (approved):
Hi @hengsin - I found a few pending entries, can you please verify also this patch:
[IDEMPIERE-4722_add.txt](https://github.com/idempiere/idempiere/files/6067698/IDEMPIERE-4722_add.txt)

And apply it if you think is OK.

PR #616, opened 2021-03-08
IDEMPIERE-4692 Mandatory range parameters : "To" style not updated after being filled manually

CarlosRuiz-globalqss:
* based on patch from Anozi Mada

https://idempiere.atlassian.net/browse/IDEMPIERE-4692


hengsin (approved):
looks good

PR #617, opened 2021-03-08
IDEMPIERE-4708 Context from ProcessParameterPanel is not cleared when closing the window

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4708


CarlosRuiz-globalqss:
mixed with another ticket - closing to reopen correct

PR #623, opened 2021-03-11
IDEMPIERE-4732 : WSearchEditor - make lookup, infoPanel and adwindow

nmicoud:


hengsin (changes requested):
Making this public is against the common practices of Java. If you need to access it in subclass, you should make it protected. If you need to access it from other places, add public getter method, for example getInfoPanel() {...}

nmicoud:
Oops, sorry, fixed

hengsin (approved):
trivial change, good to go.

PR #628, opened 2021-03-12
IDEMPIERE-4736 FHCA-2461 IDEMP - ad_column table - Framework query improvement

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4736

### DIFF
--- a/migration/i8.2/oracle/202103091322_PerformanceUpper.sql
+++ b/migration/i8.2/oracle/202103091322_PerformanceUpper.sql
@@ -1,0 +1,10 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Mar 9, 2021, 1:20:04 PM CET
+UPDATE AD_Ref_Table SET WhereClause='AD_Table.IsView=''N'' AND AD_Table.IsActive=''Y'' AND EXISTS (SELECT 1 FROM AD_Column c WHERE c.AD_Table_ID=AD_Table.AD_Table_ID AND UPPER(c.ColumnName)=UPPER(AD_Table.TableName)||''_ID'' AND c.IsActive=''Y'')',Updated=TO_DATE('2021-03-09 13:20:04','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Reference_ID=200165
+;
+
+SELECT register_migration_script('202103091322_PerformanceUpper.sql') FROM dual
+;
+
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/RolePanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/RolePanel.java
@@ -686,9 +686,9 @@ public class RolePanel extends Window implements EventListener<Event>, Deferrabl
         userPreference.setProperty(UserPreference.P_WAREHOUSE, lstItemWarehouse != null ? (String) lstItemWarehouse.getValue() : "0");
         userPreference.savePreference();
 
-        //force reload of default role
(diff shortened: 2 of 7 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/RolePanel.java
```diff
@@ -686,9 +686,9 @@ public void onCallback(Integer result) {
         userPreference.setProperty(UserPreference.P_WAREHOUSE, lstItemWarehouse != null ? (String) lstItemWarehouse.getValue() : "0");
         userPreference.savePreference();
 
-        //force reload of default role
-        MRole.getDefault(m_ctx, true);
-
+        //force reload of default role when more than 1 client
+        if (lstClient.getChildren().size() > 1)
+        	MRole.getDefault(m_ctx, true);
```
hengsin, line 691: @CarlosRuiz-globalqss , the 2 changes for MRole.getDefault is not mention at all in the ticket.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/RolePanel.java
CarlosRuiz-globalqss: Thanks @hengsin - the query on MRole is fixed, but beyond that I analyzed what was causing that query to be so frequent if there is a cache for that.

So, found that Login.loadPreferences is calling that forcing reload (no cache) and from my tests that's not necessary at all, the cache can be used without problem and in case of changes to the role a Cache Reset will solve the problem - which is the usual case.

On the other hand, in RolePanel.validateRoles I noticed that I added that call forcing reload (no cache) because of ticket [IDEMPIERE-4268](https://idempiere.atlassian.net/browse/IDEMPIERE-4268) on the commit https://github.com/idempiere/idempiere/commit/71a2ee3
This commit says "Fix problem changing client in role panel when more than 2 clients" - so instead of doing the database query always (which is expensive) I isolated that to the case that the fix mentions, when there are more than one client (one client is the most common case for most users)
Anyways, I repeated the tests on IDEMPIERE-4268 without that cache reload and I was not able to reproduce the issue, but I left that piece of code for safety - maybe my test cases were not enough.

migration/i8.2/oracle/202103091322_PerformanceUpper.sql
```diff
@@ -0,0 +1,10 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Mar 9, 2021, 1:20:04 PM CET
+UPDATE AD_Ref_Table SET WhereClause='AD_Table.IsView=''N'' AND AD_Table.IsActive=''Y'' AND EXISTS (SELECT 1 FROM AD_Column c WHERE c.AD_Table_ID=AD_Table.AD_Table_ID AND UPPER(c.ColumnName)=UPPER(AD_Table.TableName)||''_ID'' AND c.IsActive=''Y'')',Updated=TO_DATE('2021-03-09 13:20:04','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Reference_ID=200165
+;
+
+SELECT register_migration_script('202103091322_PerformanceUpper.sql') FROM dual
```
hengsin, line 8: perhaps this should be renamed to follow the usual naming convention for migration script ?

CarlosRuiz-globalqss: done

hengsin (approved):
looks good.

PR #631, opened 2021-03-15
IDEMPIERE-4731 No default salesrep_id in PO/SO,Invoice when using 'Change Role'

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4731


hengsin (approved):
looks good.

PR #634, opened 2021-03-16
IDEMPIERE-4727 Movement Window: a new line can be created for a movement that is completed.

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4727

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAllocationHdr.java
+++ b/org.adempiere.base/src/org/compiere/model/MAllocationHdr.java
@@ -947,17 +947,18 @@ public class MAllocationHdr extends X_C_AllocationHdr implements DocAction
 		} // for all lines
 		return true;
 	}	//	updateBP
-	
+
 	/**
-	 * 	Document Status is Complete or Closed
-	 *	@return true if CO, CL or RE
+	 * 	Document Status is Complete, Closed, Reversed or Voided
+	 *	@return true if CO, CL, RE or VO
 	 */
 	public boolean isComplete()
 	{
 		String ds = getDocStatus();
-		return DOCSTATUS_Completed.equals(ds) 
+		return DOCSTATUS_Completed.equals(ds)
 			|| DOCSTATUS_Closed.equals(ds)
-			|| DOCSTATUS_Reversed.equals(ds);
+			|| DOCSTATUS_Reversed.equals(ds)
+			|| DOCSTATUS_Voided.equals(ds);
 	}	//	isComplete
 
 	/**
--- a/org.adempiere.base/src/org/compiere/model/MAllocationHdr.java
+++ b/org.adempiere.base/src/org/compiere/model/MAllocationHdr.java
@@ -947,18 +947,17 @@ public class MAllocationHdr extends X_C_AllocationHdr implements DocAction
 		} // for all lines
 		return true;
 	}	//	updateBP
-
+	
 	/**
-	 * 	Document Status is Complete, Closed, Reversed or Voided
-	 *	@return true if CO, CL, RE or VO
+	 * 	Document Status is Complete or Closed
+	 *	@return true if CO, CL or RE
 	 */
 	public boolean isComplete()
 	{
 		String ds = getDocStatus();
-		return DOCSTATUS_Completed.equals(ds)
+		return DOCSTATUS_Completed.equals(ds) 
 			|| DOCSTATUS_Closed.equals(ds)
-			|| DOCSTATUS_Reversed.equals(ds)
-			|| DOCSTATUS_Voided.equals(ds);
+			|| DOCSTATUS_Reversed.equals(ds);
 	}	//	isComplete
 
 	/**
--- a/org.adempiere.base/src/org/compiere/model/MInOutConfirm.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOutConfirm.java
@@ -51,7 +51,7 @@ public class MInOutConfirm extends X_M_InOutConfirm implements DocAction
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 5270365186462536874L;
+	private static final long serialVersionUID = -1998947558580855224L;
 
 	/**
 	 * 	Create Confirmation or return existing one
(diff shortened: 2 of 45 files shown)


hengsin:
This feels like an inconsistency to me - edit and delete works without readonly logic but not for insert.

hengsin:
The inconsistency - AbstractADWindowContent, line after 1712, only enableDelete check processed.

CarlosRuiz-globalqss:
> The inconsistency - AbstractADWindowContent, line after 1712, only enableDelete check processed.

That sounds correct.  You cannot delete a processed record, but you can add new records.

hengsin:
> > The inconsistency - AbstractADWindowContent, line after 1712, only enableDelete check processed.
> 
> That sounds correct. You cannot delete a processed record, but you can add new records.

No, correct for level 0, wrong for level 1+

CarlosRuiz-globalqss:
> No, correct for tab level 0, wrong for tab level 1+

There are detail tabs that still allow edit/delete/insert in processed orders or invoices, for example Payment Schedule.

hengsin:
ok, with the readonly logic approach, we would have to announce 2 things here:
- if implementation has set custom readonly logic on those tab, will have to fix it.
- many implementation create duplicate of system window to apply customization (for e.g, a custom movement window), they will have to applied the readonly logic fix themselve.

CarlosRuiz-globalqss:
> ok, with the readonly logic approach, we would have to announce 2 things here:
> 
>     * if implementation has set custom readonly logic on those tab, they will have to use a different approach
> 
>     * many implementation create duplicate of system window to apply customization (for e.g, a custom movement window), they will have to applied the readonly logic fix themselve.

We can add a note in the [Migration_Notes](https://wiki.idempiere.org/en/Migration_Notes#Version_iDempiere_8.2), but anyways I think that's the normal thing to do when applying migration scripts.  Practically every migration script that contains update and delete are potentially doing something on top of a customization - the recommended practice I think is to have an inventory of dictionary changes.  And every insert can impact too if the implementor have custom windows, which must also have in inventory to control.

Regards,

Carlos Ruiz

CarlosRuiz-globalqss:
Added a new commit to do the validation also in model classes, to cover also the case when detail records are added/modified using webservices or other non-UI means.

org.adempiere.base/src/org/compiere/model/MAllocationHdr.java
```diff
@@ -947,17 +947,18 @@ private boolean updateBP()
...
 	 */
 	public boolean isComplete()
 	{
 		String ds = getDocStatus();
-		return DOCSTATUS_Completed.equals(ds) 
+		return DOCSTATUS_Completed.equals(ds)
 			|| DOCSTATUS_Closed.equals(ds)
-			|| DOCSTATUS_Reversed.equals(ds);
+			|| DOCSTATUS_Reversed.equals(ds)
+			|| DOCSTATUS_Voided.equals(ds);
```
hengsin, line 961: This is probably a change that might break some existing code/expectation.

hengsin: Found an example of that at InvoiceGenerate, line 267

CarlosRuiz-globalqss: Changed back all the changes to isComplete to let them as before, and change the logics in beforeSave to use isProcessed instead of isComplete.

The example at InvoiceGenerate:267 looks like a bug, in theory harmless as is "processing" void shipments - but in theory void shipments must have quantities in zero.

org.adempiere.base/src/org/compiere/model/MInOutConfirm.java
```diff
@@ -841,5 +841,18 @@ public int getC_Currency_ID()
...
+	 * 	Document Status is Complete, Closed, Reversed or Voided
+	 *	@return true if CO, CL, RE or VO
+	 */
+	public boolean isComplete()
+	{
+		String ds = getDocStatus();
+		return DOCSTATUS_Completed.equals(ds)
+			|| DOCSTATUS_Closed.equals(ds)
+			|| DOCSTATUS_Reversed.equals(ds)
+			|| DOCSTATUS_Voided.equals(ds);
```
hengsin, line 855: this is now inconsistent with isComplete() in other model class

org.adempiere.base/src/org/eevolution/model/MDDOrder.java
```diff
@@ -1274,9 +1274,10 @@ public int getC_Currency_ID() {
 	public boolean isComplete()
 	{
 		String ds = getDocStatus();
-		return DOCSTATUS_Completed.equals(ds) 
+		return DOCSTATUS_Completed.equals(ds)
 			|| DOCSTATUS_Closed.equals(ds)
-			|| DOCSTATUS_Reversed.equals(ds);
+			|| DOCSTATUS_Reversed.equals(ds)
+			|| DOCSTATUS_Voided.equals(ds);
```
hengsin, line 1280: this is now inconsistent with isComplete() in other model class

org.adempiere.base/src/org/compiere/model/MInOutLineConfirm.java
```diff
@@ -182,6 +182,11 @@ protected boolean beforeDelete ()
 	 */
 	protected boolean beforeSave (boolean newRecord)
 	{
+		MInOutConfirm parent = new MInOutConfirm(getCtx(), getM_InOutConfirm_ID(), get_TrxName());
+		if (newRecord && parent.isComplete()) {
```
hengsin, line 186: this is now inconsistent with isProcessed() check in other model class

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MInOutConfirm.java
CarlosRuiz-globalqss: solved

org.adempiere.base/src/org/eevolution/model/MDDOrder.java
CarlosRuiz-globalqss: solved

org.adempiere.base/src/org/compiere/model/MInOutLineConfirm.java
CarlosRuiz-globalqss: solved

hengsin (approved):
looks good.

PR #638, opened 2021-03-24
IDEMPIERE-4745 Fixes order by SQL generation error

Soundbytes:
Fixes order by SQL generation error in CreateWIndowFromTable process.
JIra: https://idempiere.atlassian.net/browse/IDEMPIERE-4745


CarlosRuiz-globalqss (approved):
Tested

Soundbytes:
Revisiting the issue I found that I had overlooked another error in the code I thought I had fixed.
Zero is a valid index value. So it should rather be: 
```
...
if (table.getColumnIndex("Value") >= 0 && ...
...
```
Also for plugin developers this is a trap that is too easy to fall in. 
Because of that I would propose to discourage the usage of getcolumnIndex() to check if a column exists and instead 
add a dedicated method to the API:
boolean MTable columnExists(int)
boolean MTable columnExists(String)

PR #640, opened 2021-03-26
IDEMPIERE-4746 Wrong usage of MTable.getColumnIndex

CarlosRuiz-globalqss:
* Implementing suggestion from Andreas Sumerauer
* Fix issue in IDEMPIERE-3916 (Move Client) and IDEMPIERE-1026 (Create Window from Table)

https://idempiere.atlassian.net/browse/IDEMPIERE-4746

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MTree_Base.java
+++ b/org.adempiere.base/src/org/compiere/model/MTree_Base.java
@@ -406,12 +406,12 @@ public class MTree_Base extends X_AD_Tree implements ImmutablePOSupport
 		
 		String tableName = getSourceTableName(true);
 		MTable table = MTable.get(getCtx(), tableName);
-		if (! table.columnExists("IsSummary")) {
+		if (! table.columnExistsNonVirtual("IsSummary")) {
 			// IsSummary is mandatory column to have a tree
 			log.saveError("Error", "IsSummary column required for tree tables"); 
 			return false;
 		}
-		if (! table.columnExists("Value")) {
+		if (! table.columnExistsNonVirtual("Value")) {
 			if (isTreeDrivenByValue()) {
 				// Value is mandatory column to have a tree driven by Value
 				setIsTreeDrivenByValue(false);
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/model/MTree_Base.java
```diff
@@ -406,12 +406,12 @@ protected boolean beforeSave (boolean newRecord)
...
 		String tableName = getSourceTableName(true);
 		MTable table = MTable.get(getCtx(), tableName);
-		if (table.getColumnIndex("IsSummary") < 0) {
+		if (! table.columnExists("IsSummary")) {
 			// IsSummary is mandatory column to have a tree
 			log.saveError("Error", "IsSummary column required for tree tables"); 
 			return false;
 		}
-		if (table.getColumnIndex("Value") < 0) {
+		if (! table.columnExists("Value")) {
```
Soundbytes, line 414: The Value column here needs to exist in the database. Complete if clause:
if (! table.columnExists("Value") || table.getColumn("Value").isVirtual()) ...

CarlosRuiz-globalqss: Thanks @Soundbytes - you're right - I tracked the 10 cases where that's used in core and all of them lead to a query using that column.

So, I decided is safer to add the method columnExistsNonVirtual - whoever wants to check the existence of virtual column can still use the getColumnIndex approach.

Soundbytes: @CarlosRuiz-globalqss Why not leave columnExists() as a check wether the column exists in the dictionary? After all all the code examples using getColumnIndex() for the purpose will be gone after this patch is merged.

CarlosRuiz-globalqss: What I didn't like leaving that method is that it can create again the confusion.  When a developer is looking for such method it will find first columnExists and probably will just use that one without looking if there is another.

Soundbytes: I see. Then maybe rename it to columnExistsInDictionary. I still find the getColumnIndex() approach error prone and now with no code examples available even more so. columnExistsNonVirtual could then be renamed to columnExistsInDB. Then it is alphabetically listed before the other method.

CarlosRuiz-globalqss: OK - done

hengsin (approved):
looks good

PR #641, opened 2021-03-26
IDEMPIERE-4747 Workflow Activity do not pass PO from Workflow Process

igorpojzl:
Issue: https://idempiere.atlassian.net/browse/IDEMPIERE-4747


hengsin:
Some unit test fail after application of this patch, still investigating what's the cause of it.

hengsin (approved):
tested ok. the unit test error is due to bug in the unit test case and a minor bug in document engine. will push after after the merging of this pr.

PR #642, opened 2021-03-30
IDEMPIERE-4749 : Allow to use OAuth2 libraries on plugins

nmicoud:


CarlosRuiz-globalqss:
Comments in ticket

hengsin:
will move the dependency to target platform instead.

PR #643, opened 2021-04-02
IDEMPIERE-4752 :  Cross tenant error fix while importing client data …

dpansheriya:
…2pack

To reproduce issue.
I exported a role from client and imported into another client where more then 1 client presents. It fails as pack in process try to run role access update for role belongs to other client too.


CarlosRuiz-globalqss (changes requested):
I think the change is OK when the PackIn is running on a non-System tenant.  But still is desirable to run role access update in all tenants when the PackIn is running on System.

CarlosRuiz-globalqss (approved):

PR #647, opened 2021-04-08
IDEMPIERE-4755 Change data access level of Report View to System+Client

ralexsander:
Report View are very useful to add WHERE clauses to VIEWs created on Table & Column window, so you can keep a single VIEW and use it with several "Report Views". 

The proposal of this ticket is to change the access level of Report View table to System+Client, so can create new Report Views on client instead of System. 

Also we have to make the field AD_ReportView_ID updatable on table AD_PrintFormat.

This feature is similar to Window Customization, but to be used with reports.


CarlosRuiz-globalqss (changes requested):
Hi @ralexsander - see comments in ticket

CarlosRuiz-globalqss (approved):

PR #649, opened 2021-04-08
IDEMPIERE-4757 Fix the Scheduler not working with Print Format

ralexsander:
Scheduler processes is not considering the Print Format defined in Scheduler itself.

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/ServerReportCtl.java
+++ b/org.adempiere.base/src/org/compiere/print/ServerReportCtl.java
@@ -289,9 +289,7 @@ public class ServerReportCtl {
 			MQuery query = MQuery.get (ctx, pi.getAD_PInstance_ID(), TableName);
 			PrintInfo info = new PrintInfo(pi);
 			re = new ReportEngine(ctx, format, query, info, pi.isSummary(), pi.getTransactionName());
-			if (pi.getSerializableObject() != null && pi.getSerializableObject() instanceof MPrintFormat) {
-				re.setPrintFormat((MPrintFormat)pi.getSerializableObject());
-			}
+			re.setPrintFormat((MPrintFormat) o);
 			if (pi.isPrintPreview() && pi.isBatch())
 			{
 				if ("HTML".equals(pi.getReportType())) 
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/print/ServerReportCtl.java
```diff
@@ -289,6 +289,9 @@ static public boolean startStandardReport (ProcessInfo pi)
 			MQuery query = MQuery.get (ctx, pi.getAD_PInstance_ID(), TableName);
 			PrintInfo info = new PrintInfo(pi);
 			re = new ReportEngine(ctx, format, query, info, pi.isSummary(), pi.getTransactionName());
+			if (pi.getSerializableObject() != null && pi.getSerializableObject() instanceof MPrintFormat) {
+				re.setPrintFormat((MPrintFormat)pi.getSerializableObject());
```
hengsin, line 293: why is this needed here ? the printformat from pi.getSerializableObject() is already pass into the ReportEngine constructor at line 291 above.

org.adempiere.base/src/org/compiere/print/ServerReportCtl.java
```diff
@@ -289,6 +289,7 @@ static public boolean startStandardReport (ProcessInfo pi)
 			MQuery query = MQuery.get (ctx, pi.getAD_PInstance_ID(), TableName);
 			PrintInfo info = new PrintInfo(pi);
 			re = new ReportEngine(ctx, format, query, info, pi.isSummary(), pi.getTransactionName());
+			re.setPrintFormat((MPrintFormat) o);
```
hengsin, line 292: I don't get it - isn't the format argument to ReportEngine constructor above is already doing that ( "MPrintFormat format = (MPrintFormat)o;" at line 287) ?

ralexsander: > I don't get it - isn't the format argument to ReportEngine constructor above is already doing that ( "MPrintFormat format = (MPrintFormat)o;" at line 287) ?
You are right, I'll check again, maybe the problem is in the ReportEngine itself

ralexsander: It was probably something bad in my environment. I was not able to reproduce using master branch. 

If you take a closer look at ReportEngine constructor, you see the PrintFormat is set directly to a variable. The method setPrintFormat does other things besides setting the field, but I spend all day and could not reproduce the random problem using the master branch. I will re-open this if It happens again and I figure it out how to reproduce properly.

ralexsander:
It was probably something bad in my environment. I was not able to reproduce using master branch.

PR #652, opened 2021-04-13
IDEMPIERE-4760 Credit check do not allow complete Customer Return

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4760

Disable Credit Check on Customer Returns.


hengsin (approved):
looks good.

PR #653, opened 2021-04-14
IDEMPIERE-2377: Add "Prepare" Document Action to all document type

dpansheriya:
As per requested by Chuck, Enabling Prepare action for all document type. If needs to disable, this can be controlled by document action access per role.


CarlosRuiz-globalqss (changes requested):
please check comment in ticket

CarlosRuiz-globalqss (approved):

PR #655, opened 2021-04-15
IDEMPIERE-4764 Hazelcast distributed cache can be an issue in some environments

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4764

- Change setting for all caches to be non-distributed


hengsin (approved):
looks good.

PR #657, opened 2021-04-16
IDEMPIERE-4763 Add IsDefault to AD_StorageProvider and use it when creating a new tenant (FHCA-1165)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4763

* Add column/field AD_StorageProvider.IsDefault
* Includes a minor refactoring of Login.loadDefault to add the ID column to the order by, to make predictable the ID returned when there are more than one record marked as default
* Context variable #AD_StorageProvider_ID is not created


hengsin (approved):
looks good.

PR #658, opened 2021-04-16
IDEMPIERE-4766 Record ID button would be more useful if it shows the ID

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4766


hengsin (approved):
trivial change, good to go

PR #659, opened 2021-04-19
IDEMPIERE-3062 Unable to complete MR with qty >1 with Attribute Set of Always Mandatory

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-3062

Fix wrong query on M_InoutLineMA


hengsin (approved):
trivial fix, good to go

PR #660, opened 2021-04-19
IDEMPIERE-4768 Cannot ship in some cases when multiple ASI with different material policy

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4768

WIP
Pending unit test
Pending review why this is not failing with sample GW data


CarlosRuiz-globalqss:
> Pending review why this is not failing with sample GW data

Effectively there was a bug on material receipt using the date of storage as material policy instead of the date of the document (or ASI).

> Pending unit test

Both unit tests added

hengsin (approved):
looks good.

PR #661, opened 2021-04-20
IDEMPIERE-4769 False result in validation for pending changes

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4769


hengsin (approved):
looks good

PR #664, opened 2021-04-23
IDEMPIERE-4709 Print Format Item: implement script column (expression)

CarlosRuiz-globalqss:
Fix issue when using spaces on the name of the print format item


CarlosRuiz-globalqss:
Forced push to fix bug

PR #669, opened 2021-04-28
IDEMPIERE-4774 Matching PO-Receipt-Invoice is still view Vendor RMA as unmatched receipt

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4774

Wrong number sign when matching Vendor Return and Credit Memo


CarlosRuiz-globalqss (approved):
tested

PR #670, opened 2021-04-28
COF #11523 / IDEMPIERE-4715 - attribute set instance: new qty available column and group results in select existing record dialog

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4715


CarlosRuiz-globalqss:
pushing tested and merged patch from Murilo

PR #675, opened 2021-05-04
IDEMPIERE-4777 New table whose name ends in "Line" and does not have a Header table

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4777


d-ruiz (approved):
Looks good

PR #689, opened 2021-05-17
IDEMPIERE-4790 : Deactivating ToolbarDetail Customize button throws N…

nmicoud:
…PE when using Toggle button


hengsin (approved):
looks good.

PR #690, opened 2021-05-17
IDEMPIERE-4779 : Dashboard / Report : parameters are wrongly displayed

nmicoud:


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -985,6 +988,17 @@ else if (value instanceof Integer)
 							 iPara.setInfo(info);
 						 }
 					 }
+					 else if (DisplayType.isID(iPara.getDisplayType())) {
```
hengsin, line 991: isID is not precise enough since that includes DisplayType.ID, DisplayType.Chart, etc. To handle just table, tabledirr and search, you need to use for just this 3 display type (i.e iPara.getDisplayType() == DisplayType.Table, etc). If the code works for all lookup displaytype, the check should be  DisplayType.isLookup.

nmicoud:
Hi,
Changes done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -1021,7 +1035,48 @@ else if (DisplayType.isDate(iPara.getDisplayType()))
...
+
+	private static final String displayColumns = "Name,Value,DocumentNo";
+	private String getDisplay(MPInstance i, MPInstancePara ip, Object id) {
+		try {
+			MProcessPara pp = MProcess.get(i.getAD_Process_ID()).getParameter(ip.getParameterName());
+			MTable table = null;
+			String column = "";
+
+			if (pp != null) {
+				if (pp.getAD_Reference_ID() == DisplayType.Search || pp.getAD_Reference_ID() == DisplayType.TableDir)
```
hengsin, line 1047: this checking is not correct, DisplayType.Search can have AD_Reference_Value_ID too.

nmicoud:
ok done
nb: just noted that you can use a AD_Reference_Value_ID for tabledir also

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -1021,7 +1035,50 @@ else if (DisplayType.isDate(iPara.getDisplayType()))
...
+
+					return DB.getSQLValueStringEx(null, sql.toString(), id);
+				}
+			}
+		}
+		catch (Exception e) {
+			logger.log(Level.WARNING, "Failed to retrieve data to display for embedded report " + MProcess.get(i.getAD_Process_ID()).getName() + " : " + ip.getParameterName(), e);
+		}
+
+		return id.toString();
```
hengsin, line 1079: Instead of coding the display column stuff here again, I think we should use the getlookupinfo methods in mlookupfactory and use the MLookupInfo.QueryDirect to get the display value. This reduce the duplicate of code and at the same give better consistency with other part of the application.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
nmicoud: Something like that ?

hengsin (approved):
looks good

PR #692, opened 2021-05-20
IDEMPIERE-4792 Bring back frozen column feature for desktop browser

hengsin:
Incorporate workaround from zk tracker and re-eanble grid frozen columns
for dekstop browser.


CarlosRuiz-globalqss (approved):
tested successfully

PR #694, opened 2021-05-20
IDempiere-4794

Soundbytes:
IQuickEntryFactory osgi service
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4794

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
@@ -34,7 +34,7 @@ import org.adempiere.webui.factory.IDashboardGadgetFactory;
 import org.adempiere.webui.factory.IFormFactory;
 import org.adempiere.webui.factory.IMappedFormFactory;
 import org.adempiere.webui.factory.IQuickEntryFactory;
-import org.adempiere.webui.grid.IQuickEntry;
+import org.adempiere.webui.grid.AbstractWQuickEntry;
 import org.adempiere.webui.panel.ADForm;
 import org.compiere.grid.ICreateFrom;
 import org.compiere.grid.ICreateFromFactory;
@@ -258,14 +258,14 @@ public class Extensions {
 		return formFactoryService;
 	}
 	
-	private final static CCache<String, IServiceReferenceHolder<IQuickEntryFactory>> s_quickEntryFactoryCache = new CCache<>(null, "IFormFactory", 100, false);
+	private final static CCache<Integer, IServiceReferenceHolder<IQuickEntryFactory>> s_quickEntryFactoryCache = new CCache<>(null, "IQuickEntryFactory", 100, false);
 	
 	/**
 	 *
 	 * @param AD_Window_ID 
 	 * @return IQuickEntryFactory instance or null if AD_Window_ID not found
 	 */
-	private static IQuickEntryFactory getQuickEntryService(String AdWindowID) {
+	private static IQuickEntryFactory getQuickEntryService(Integer AdWindowID) {
 		IServiceReferenceHolder<IQuickEntryFactory> cache = s_quickEntryFactoryCache.get(AdWindowID);
 		if (cache != null) {
 			IQuickEntryFactory service = cache.getService();
@@ -292,10 +292,10 @@ public class Extensions {
 	 * @param AD_Window_ID 
 	 * @return IQuickEntry instance or null if AD_Window_ID not found
 	 */
-	public static IQuickEntry getQuickEntry(int AD_Window_ID) {
-		IQuickEntryFactory service = getQuickEntryService(String.valueOf(AD_Window_ID));
+	public static AbstractWQuickEntry getQuickEntry(int AD_Window_ID) {
+		IQuickEntryFactory service = getQuickEntryService(AD_Window_ID);
 		if (service != null) {
-			IQuickEntry quickEntry = service.newQuickEntryInstance(AD_Window_ID);
+			AbstractWQuickEntry quickEntry = service.newQuickEntryInstance(AD_Window_ID);
 			if (quickEntry != null)
 				return quickEntry;
 		}
@@ -309,10 +309,10 @@ public class Extensions {
 	 * @param TabNo 
 	 * @return IQuickEntry instance or null if AD_Window_ID not found
 	 */
-	public static IQuickEntry getQuickEntry(int WindowNo, int TabNo, int AD_Window_ID) {
-		IQuickEntryFactory service = getQuickEntryService(String.valueOf(AD_Window_ID));
+	public static AbstractWQuickEntry getQuickEntry(int WindowNo, int TabNo, int AD_Window_ID) {
+		IQuickEntryFactory service = getQuickEntryService(AD_Window_ID);
 		if (service != null) {
-			IQuickEntry quickEntry = service.newQuickEntryInstance(WindowNo, TabNo, AD_Window_ID);
+			AbstractWQuickEntry quickEntry = service.newQuickEntryInstance(WindowNo, TabNo, AD_Window_ID);
 			if (quickEntry != null)
 				return quickEntry;
 		}
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
@@ -41,6 +41,7 @@ import org.compiere.grid.ICreateFromFactory;
 import org.compiere.grid.IPaymentForm;
 import org.compiere.grid.IPaymentFormFactory;
 import org.compiere.model.GridTab;
+import org.compiere.model.MDashboardContent;
 import org.compiere.util.CCache;
 import org.zkoss.zk.ui.Component;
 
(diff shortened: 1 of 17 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
```diff
@@ -254,5 +257,65 @@ public static IMappedFormFactory getMappedFormFactory(){
 		}
 		return formFactoryService;
 	}
-
+	
+	private final static CCache<String, IServiceReferenceHolder<IQuickEntryFactory>> s_quickEntryFactoryCache = new CCache<>(null, "IFormFactory", 100, false);
```
hengsin, line 261: wrong cache name

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
```diff
@@ -254,5 +257,65 @@ public static IMappedFormFactory getMappedFormFactory(){
...
+	
+	/**
+	 *
+	 * @param AD_Window_ID 
+	 * @return IQuickEntryFactory instance or null if AD_Window_ID not found
+	 */
+	private static IQuickEntryFactory getQuickEntryService(String AdWindowID) {
```
hengsin, line 268: ad_window_id should be int instead of string

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
```diff
@@ -141,15 +144,15 @@ public WQuickEntry(int AD_Window_ID)
...
+		log.info("R/O=" + isReadOnly());
 	}	//	WQuickEntry
 
 	/**
 	 *	Static Init
 	 * 	@throws Exception
 	 */
 
-	private void jbInit() throws Exception
+	protected void jbInit() throws Exception
```
hengsin, line 155: would be better to rename this to just init(), jbinit sounds weird.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/IQuickEntry.java
```diff
@@ -0,0 +1,67 @@
...
+import org.zkoss.zk.ui.Component;
+import org.zkoss.zk.ui.event.Event;
+import org.zkoss.zk.ui.event.EventListener;
+
+/**
+ * Quick Entry Window
+ * Interface based on class WQuickEntry by Carlos Ruiz
+ * @Author Andreas Sumerauer
+ */
+public interface IQuickEntry extends  Component, ISupportMask, EventListener<Event>, ValueChangeListener
```
hengsin, line 31: This design change make it necessary to do a window cast at WTableDirEditor. Unless we can find a way to get rid of that class cast, this probably not a good design change. It implies that you can create a IQuickEntry implementation that doesn't extends Window and will cause class cast exception.

Soundbytes: I am aware of that inconsistency and I don't like it either. 
I should be able to solve it using an abstract base class AbstractWQuickEntry instead of the Interface.
(Should have thought of that earlier)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
Soundbytes: :)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
Soundbytes: I tried to deduct the naming convention from the existing caches. If I was wrong what would be the correct name (and the convention)? 
Key type will be changed to Integer.

hengsin: not the variable name but "IFormFactory" which is obviously wrong here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
hengsin: if that's the case, perhaps there isn't a needs here to change it to protected ? otherwise, you can leave it as it is.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultQuickEntryFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class DefaultQuickEntryFactory implements IQuickEntryFactory {
+
+	@Override
+	public AbstractWQuickEntry newQuickEntryInstance(int WindowNo, int TabNo, int AD_Window_ID) {
+		return new WQuickEntry(WindowNo, TabNo, AD_Window_ID);
+	}
+
+	@Override
+	public AbstractWQuickEntry newQuickEntryInstance(int AD_Window_ID) {
```
hengsin, line 34: what's the use case for this ? don't see it being use anywhere.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IQuickEntryFactory.java
```diff
@@ -0,0 +1,50 @@
...
+	 */
+	public AbstractWQuickEntry newQuickEntryInstance(int WindowNo, int TabNo, int AD_Window_ID);
+	
+	/**
+	 * @param AD_Window_ID
+	 * @return new QuickEntry instance
+	 */
+	public AbstractWQuickEntry newQuickEntryInstance(int AD_Window_ID);
+	
+}
```
hengsin, line 50: what's the use case for the first and third method ? don't see it being use anywhere.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
Soundbytes: It is protected to allow for further reaching layout customization in the child class.
I would rather not make it private because of this.

hengsin: alright, better rename it to something that wouldn't cause compatibility issue, for e.g initQuickEntryLayout(). There might be subclass of WQuickEntry out there that have the same jbInit private method.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IQuickEntryFactory.java
Soundbytes: The original WQuickEdit class has public constructors with the same signature. The motivation for the additional factory methods was just to provide the same interface. 
If they are in fact redundant I'll be happy to remove them.

hengsin: alright, that looks ok. checking the usage of the WQuickEntry constructors, your changes miss the use of WQuickEntry at InfoWindow.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
hengsin: WPreference extends WQuickEntry and have its own protected jbInit method. Therefore, if you change WQuickEntry's jbInit method to protected, you MUST rename it to something else, otherwise it will break WPreference.

Soundbytes: You are right. I again forgot to consider the chance that some unpublished customization could break as a consequence. Will rename the method to initLayout() and make it protected.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IQuickEntryFactory.java
Soundbytes: You are right. Looking at the codebase I'd say we should in fact only expose the interface that is used by WTableDirEditor. I'll remove the other factory functions then.

hengsin: hmm .. shouldn't the interface be used in InfoWindow as well ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/grid/WQuickEntry.java
Soundbytes: Done.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IQuickEntryFactory.java
Soundbytes: Yes. I missed that completely. It's fixed now.

hengsin (approved):
tested

Soundbytes:
@hengsin Thanks for taking the time to get this mature so quickly.

PR #695, opened 2021-05-21
IDEMPIERE-434  Tab field group is not available in web

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-434

WTabEditor is not supported at this moment.

I did not know how to handle when we want to save Record and we have Mandatory Field on not selected Group Tab.


hengsin (changes requested):
there is a bug here - tabbox is always render even when there's no field with fieldgrouptype=tab.

igorpojzl:
Fixed Bug. tabbox is Rendered only when he has field with fieldgrouptype=tab and form is visible.

hengsin (approved):
tested

PR #696, opened 2021-05-24
IDEMPIERE-4796 Request EMail Processor  - Html email processing type parameter

igorpojzl:
Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4796

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
+++ b/org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
@@ -463,7 +463,7 @@ public class RequestEMailProcessor extends SvrProcess implements ProcessEmailHan
 		
 		if (log.isLoggable(Level.INFO)) log.info("created request " + req.getR_Request_ID() + " from msg -> " + emailContent.subject);
 		
-		if(p_HTMLAttachmentType.contains("H")) {
+		if("H".equals(p_HTMLAttachmentType)) {
 			String htmlContent = emailContent.getHtmlContent(true);
 			if (htmlContent != null){
 				MAttachment attach = req.createAttachment();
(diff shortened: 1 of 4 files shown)


org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
```diff
@@ -457,14 +463,26 @@ protected void createRequest(EmailContent emailContent, String trxName) throws E
 		
 		if (log.isLoggable(Level.INFO)) log.info("created request " + req.getR_Request_ID() + " from msg -> " + emailContent.subject);
 		
-		String htmlContent = emailContent.getHtmlContent(true);
-		if (htmlContent != null){
-			MAttachment attach = req.createAttachment();
-			
-			attach.addEntry(emailContent.subject + ".html", emailContent.getHtmlContent(true).getBytes(Charset.forName("UTF-8")));
-			attach.saveEx(trxName);
+		if(p_HTMLAttachmentType.contains("H")) {
```
hengsin, line 466: what's the expected value here ? if it is "H" or "I", this should be if ("H".equals(p_HTMLAttachmentType)), same goes for the comparison of "I" below (that should  be an else if instead of just if). of course, the current code is correct if your expected value here can be "HI" or "IH".

igorpojzl: I expected there could be variant with both values. We did not add this variant at the end. I can change it to equals.

igorpojzl:
Updated p_HTMLAttachmentType variable logic.

org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
igorpojzl: Changed to equals.

org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
```diff
@@ -457,14 +463,24 @@ protected void createRequest(EmailContent emailContent, String trxName) throws E
...
+			String htmlContent = emailContent.getHtmlContent(true);
+			if (htmlContent != null){
+				MAttachment attach = req.createAttachment();
+				
+				attach.addEntry(emailContent.subject + ".html", emailContent.getHtmlContent(true).getBytes(Charset.forName("UTF-8")));
+				attach.saveEx(trxName);
+			}
+		} else if("I".equals(p_HTMLAttachmentType)) {
+			ArrayList<BodyPart> imagesList = emailContent.getHTMLImageBodyParts();
+			for(BodyPart image: imagesList) {
```
hengsin, line 476: Looking at the code, a null check for imagesList is necessary here.

hengsin (changes requested):

org.adempiere.base.process/src/org/compiere/process/RequestEMailProcessor.java
igorpojzl: Commited

hengsin (approved):
looks good.

PR #699, opened 2021-05-26
IDEMPIERE-4800 Performance - avoid multiple for in GridTab.getParentTabNo

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4800


hengsin (approved):
looks good

PR #704, opened 2021-05-31
IDEMPIERE-4811 Custom Toolbar Buttons Dynamic Validation Fix

igorpojzl:
Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4811

Expanded ToolbarCustomButton class to set tabNo when rendered on Tab Detail Pane to have access to tab Context.
Expanded DetailPane on method updateToolbar to use ToolbarCustomButton.dynamicDisplay() when toolbarbutton is CustomToolbarButton.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
@@ -34,12 +34,7 @@ public class ToolbarCustomButton implements EventListener<Event>, Evaluatee {
 	private MToolBarButton mToolbarButton;
 
 	public ToolbarCustomButton(MToolBarButton mToolbarButton, Toolbarbutton btn, String actionId, int windowNo) {
-		toolbarButton = btn;
-		this.actionId = actionId;
-		this.windowNo = windowNo;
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
```diff
@@ -41,6 +42,16 @@ public ToolbarCustomButton(MToolBarButton mToolbarButton, Toolbarbutton btn, Str
 		toolbarButton.addEventListener(Events.ON_CLICK, this);
 	}
 	
+	public ToolbarCustomButton(MToolBarButton mToolbarButton, Toolbarbutton btn, String actionId, int windowNo, int tabNo) {
```
hengsin, line 45: to reduce duplication of code, the old constructor should call the new constructor with tabNo=-1

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
igorpojzl: Fixed

hengsin (approved):
looks good

PR #711, opened 2021-06-08
IDEMPIERE-3980

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-3980

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WImageDialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WImageDialog.java
@@ -140,6 +140,9 @@ public class WImageDialog extends Window implements EventListener<Event>
 		autoPreviewList.add("image/jpeg");
 		autoPreviewList.add("image/png");
 		autoPreviewList.add("image/gif");
+		autoPreviewList.add("image/tiff");
+		autoPreviewList.add("image/bmp");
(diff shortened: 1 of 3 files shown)


hengsin:
on firefox, "Right Click > This Frame > Open Frame in New Tab" will execute the embedded script.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WImageDialog.java
```diff
@@ -328,14 +350,25 @@ private void processUploadMedia(Media imageFile) {
 			return;
 
 		String fileName = imageFile.getName();
-		
+		String mimeType = MimeType.getMimeType(fileName);
+		if (! autoPreviewList.contains(mimeType))
+			throw new AdempiereException("File not allowed for uploading, just image types jpg/png/gif");
```
hengsin, line 355: should use ad_message here.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WImageDialog.java
CarlosRuiz-globalqss: ready - committed message

hengsin (approved):
looks good

PR #712, opened 2021-06-08
IDEMPIERE-4821 - Fix missing transaction in invoice creation from com…

ralexsander:
Missing transaction in invoice creation from commission. 
Also shows the generated invoice in result dialog with zoom option.


hengsin (approved):
looks good.

PR #715, opened 2021-06-09
IDEMPIERE-4823 RV_C_Invoice does not have a key column

d-ruiz:
https://idempiere.atlassian.net/browse/IDEMPIERE-4823


CarlosRuiz-globalqss (changes requested):
Hi @d-ruiz - can you please move the scripts to the folder i8.2

I think this is a bug better to be solved in stable.

d-ruiz:
Moved the files as requested

CarlosRuiz-globalqss (approved):

PR #723, opened 2021-06-16
IDEMPIERE-4834 : Import Formats are not ordered on Import File Loader

nmicoud:


hengsin (approved):
trivial fix, good to go

PR #730, opened 2021-06-18
IDEMPIERE-3836 jasper attachment: two report/process use jasper have …

hengsin:
…attachment file has duplicate name, one will win and other loss

https://idempiere.atlassian.net/browse/IDEMPIERE-3836


CarlosRuiz-globalqss (approved):
tested

PR #731, opened 2021-06-18
IDEMPIERE-4839 IQuickForm interface

Soundbytes:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4839

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/IQuickForm.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/IQuickForm.java
@@ -1,0 +1,45 @@
+/******************************************************************************
+ * Copyright (C) Contributors                                                 *
+ * This program is free software; you can redistribute it and/or modify it    *
+ * under the terms version 2 of the GNU General Public License as published   *
+ * by the Free Software Foundation. This program is distributed in the hope   *
+ * that it will be useful, but WITHOUT ANY WARRANTY; without even the implied *
+ * warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.           *
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program; if not, write to the Free Software Foundation, Inc.,    *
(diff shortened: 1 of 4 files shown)


hengsin (changes requested):
First, there should be javadoc for the methods at IQuickForm interface. Second, this seems incomplete as the onQuickForm() method at AbstractADWindowContent is not touch at all.

Soundbytes:
> First, there should be javadoc for the methods at IQuickForm interface. 
Will be added.

> Second, this seems incomplete as the onQuickForm() method at AbstractADWindowContent is not touch at all.
Great Idea! The initial idea that I have implemented here was to make the quickform dialog available as a custom form. Since the form is opened via toolbar button or menu entry I did not have to touch the onQuickForm() method.
I'll be happy to extend the patch and provide an IQuickEdit osgi service as an alternative way to access the functionality. 
I'd however prefer to add that as a separate commit.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/WQuickForm.java
```diff
@@ -55,15 +55,13 @@
  * @author Logilite Technologies
  * @since Nov 03, 2017
  */
-public class WQuickForm extends Window implements EventListener <Event>, DataStatusListener
+public class WQuickForm extends Window implements IQuickForm // EventListener <Event>, DataStatusListener
```
hengsin, line 58: please remove the //... part

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/IQuickForm.java
```diff
@@ -0,0 +1,45 @@
...
+
+	public void onIgnore();
+
+	public void onDelete();
+
+	public void onSave();
+
+	public void onRefresh();
+
+	public void dispose();
```
hengsin, line 44: please add javadoc for all the methods

hengsin (changes requested):
please address the requested changes.

hengsin (approved):
looks good.

PR #732, opened 2021-06-18
IDEMPIERE-4840: ConfirmPanel with Checkbox

Soundbytes:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4840

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
@@ -259,15 +259,8 @@ public final class ConfirmPanel extends Div
 
         // IDEMPIERE-1334 start
         pnlBtnCenter = new Hlayout();
-        pnlBtnCenter.setSclass("z-valign-middle confirm-panel-center");
-        pnlBtnCenter.setStyle("height: 100%");
+        pnlBtnCenter.setSclass("confirm-panel-center");
         // IDEMPIERE-1334 end
-
-        // Invisible Spacer. Lets Checkboxes appear vertically aligned to center.  
-        // TODO: Find a better solution that does not require the Div height to be hardcoded.
-        Div alignmentFix = new Div();
-        alignmentFix.setStyle("content: \"\"; display: inline-block; vertical-align: middle; height: 30px;");
-        pnlBtnCenter.appendChild(alignmentFix);
         
         this.appendChild(pnlBtnLeft);
         // IDEMPIERE-1334
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
@@ -259,8 +259,15 @@ public final class ConfirmPanel extends Div
 
         // IDEMPIERE-1334 start
         pnlBtnCenter = new Hlayout();
-        pnlBtnCenter.setSclass("confirm-panel-center");
+        pnlBtnCenter.setSclass("z-valign-middle confirm-panel-center");
+        pnlBtnCenter.setStyle("height: 100%");
         // IDEMPIERE-1334 end
+
+        // Invisible Spacer. Lets Checkboxes appear vertically aligned to center.  
+        // TODO: Find a better solution that does not require the Div height to be hardcoded.
+        Div alignmentFix = new Div();
+        alignmentFix.setStyle("content: \"\"; display: inline-block; vertical-align: middle; height: 30px;");
+        pnlBtnCenter.appendChild(alignmentFix);
         
         this.appendChild(pnlBtnLeft);
         // IDEMPIERE-1334
@@ -343,6 +350,16 @@ public final class ConfirmPanel extends Div
     public void addComponentsCenter(Combobox cbb){
     	pnlBtnCenter.appendChild(cbb);
     }
+    
+    /**
+     * Add checkbox to center panel
+     * @param cb
+     * @param labelText
+     */
+    public void addComponentsCenter(Checkbox cb){
+    	pnlBtnCenter.appendChild(cb);
+    	
+    }    
     
     /**
      * return button of the specified id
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
```diff
@@ -259,8 +259,15 @@ private void init()
...
-        pnlBtnCenter.setSclass("confirm-panel-center");
+        pnlBtnCenter.setSclass("z-valign-middle confirm-panel-center");
+        pnlBtnCenter.setStyle("height: 100%");
         // IDEMPIERE-1334 end
+
+        // Invisible Spacer. Lets Checkboxes appear vertically aligned to center.  
+        // TODO: Find a better solution that does not require the Div height to be hardcoded.
+        Div alignmentFix = new Div();
+        alignmentFix.setStyle("content: \"\"; display: inline-block; vertical-align: middle; height: 30px;");
+        pnlBtnCenter.appendChild(alignmentFix);
```
hengsin, line 270: I'm sorry but we can't accept this hack to core ...

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
```diff
@@ -259,8 +259,15 @@ private void init()
 
         // IDEMPIERE-1334 start
         pnlBtnCenter = new Hlayout();
-        pnlBtnCenter.setSclass("confirm-panel-center");
+        pnlBtnCenter.setSclass("z-valign-middle confirm-panel-center");
+        pnlBtnCenter.setStyle("height: 100%");
```
hengsin, line 263: setHeight("100%") is prefer over setstyle here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
```diff
@@ -529,8 +546,11 @@ public void addActionListener(String event, EventListener<?> listener)
         // IDEMPIERE-1334 start
         while (iter3.hasNext())
         {
-            Button button = (Button)iter3.next();
-            button.addEventListener(event, listener);
+        	Object element = iter3.next();
+        	if (element instanceof Button) 
+        	{
+	            ((Button)element).addEventListener(event, listener);
```
hengsin, line 552: I think it will be better to add the same checking for iter1 and iter2 as well.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
Soundbytes: I understand. I just added a comment to the jira ticket explaining why I still opened the pull request.

This is how it looks.
Bildschirmfoto 2021-06-18 um 15 53 15
Note: There is one layout problem though with the modification:
To achieve the proper vertical center align I had to add a spacer to the center div that assures it has the same height as the left and right divs. Problem is I had to hardcode the div height which surely cannot be accepted in Production code because it would lead to problems when a theme is used that has a different button height.
Despite some extensive research on the issue I could not come up with a cleaner solution that does not require setting the div height in code. I still opened the pull request in the hope that a solution can be found when more eyes look into the matter. 
I also find the added functionality of having a checkbox available in the ConfigPanel very valuable.
Without the Spacer the checkbox is always top aligned regardless of the selected css class/style:
Bildschirmfoto 2021-06-18 um 16 11 22
In the QuickEdit Dialog it still looks acceptable because of the 9px padding above the button row.
Without the padding however it looks really crappy to have the checkbox top aligned.

Maybe we could merge nonetheless without the spacer that ensures vertical middle alignment and add a fix for the alignment problem later when one has been found.

WDYT?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
```diff
@@ -344,6 +344,16 @@ public void addComponentsCenter(Combobox cbb){
     	pnlBtnCenter.appendChild(cbb);
     }
     
+    /**
+     * Add checkbox to center panel
+     * @param cb
+     * @param labelText
```
hengsin, line 350: wrong javadoc

Soundbytes: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
Soundbytes: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ConfirmPanel.java
Soundbytes: Hackish layout solution removed. 

I might add: 
The usage of a spacer element to enforce vertical middle alignment of elements of varying height is not a hack (when it can not be guaranteed that at least one of the actually used elemets has the height of the container). 
Instead it is the only solution to the problem I am aware of.  The hack IMO is the usage of a fixed height in code and I understand that this is not acceptable. 
The problem here is that (a) the height of the parent container is often dynamically calculated and because of that can not be retrieved in advance and that (b) zk uses nested divs in the implementation of vertical middle alignment. The intermediate div that is inserted by the renderer is not accessible for styling AFAIK. Because of this the layout will only work when a fixed height is assigned to the inner div. I am not an expert in zk. So I am still hopeful that the framework addresses the issue in a way that I am not aware of. I at least could not find the solution I am afraid.

hengsin (approved):
tested

Soundbytes:
@hengsin  I am afraid I only found your solution for the alignment problem today that you added to the jira ticket.

```
Checkbox cb = new Checkbox();
cb.setLabel("This is a checkbox");
cb.setStyle("display: inline-block; padding: 4px 20px; line-height: 20px;"); 
```
While this solution works it is not theming friendly IMO because it assumes a given height difference of 8px between button and checkbox element. Most probably I am overcomplicating things and I should just add it to the pull request. Can you reopen and amend the merge?

hengsin:
If you want to make any changes, just create a new PR for it.
It is theme friendly since you can use css instead and add it to your theme. It is just an example and doesn't means you must use setStyle to do it.

PR #734, opened 2021-06-21
IDEMPIERE-2937 Html Report - Suppress Null checkbox doesn't work

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-2937


CarlosRuiz-globalqss (approved):
tested - solves HTML

PR #739, opened 2021-06-23
IDEMPIERE-4837: Ignoring pricelist in Product Info window for Price L…

dpansheriya:
…ist, Inventory Move, Physical Inventory and Internal Use Inventory

### DIFF
--- a/migration/i8.2z/oracle/202106161301_IDEMPIERE-4837.sql
+++ b/migration/i8.2z/oracle/202106161301_IDEMPIERE-4837.sql
@@ -1,0 +1,22 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
+-- 16-Jun-2021, 10:47:48 PM IST
+UPDATE AD_Menu SET PredefinedContextVariables='IgnorePriceListInProductInfo=Y',Updated=TO_DATE('2021-06-16 22:47:48','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Menu_ID=132
+;
+
+-- 16-Jun-2021, 10:48:43 PM IST
+UPDATE AD_Menu SET PredefinedContextVariables='IgnorePriceListInProductInfo=Y',Updated=TO_DATE('2021-06-16 22:48:43','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Menu_ID=181
+;
+
+-- 16-Jun-2021, 10:49:37 PM IST
+UPDATE AD_Menu SET PredefinedContextVariables='IgnorePriceListInProductInfo=Y',Updated=TO_DATE('2021-06-16 22:49:37','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Menu_ID=503
+;
+
+-- 16-Jun-2021, 10:50:15 PM IST
+UPDATE AD_Menu SET PredefinedContextVariables='IgnorePriceListInProductInfo=Y',Updated=TO_DATE('2021-06-16 22:50:15','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Menu_ID=179
+;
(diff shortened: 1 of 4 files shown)


migration/i8.2z/oracle/202106161301_IDEMPIERE-4837.sql
```diff
@@ -0,0 +1,22 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- I forgot to set the DICTIONARY_ID_COMMENTS System Configurator
```
CarlosRuiz-globalqss, line 4: Please:
* move the scripts to folder i8.2
* in the case of windows the predefined context must be added to ad_window instead of ad_menu, otherwise the context will be added just when opened from menu, not when zooming
* set the comment for the script instead of "I forgot to ..."
* I think the variable must be the opposite "UsePriceListInProductInfo" (instead of Ignore) ... see my comment below, so, the list of windows to update will be different.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoProductWindow.java
```diff
@@ -443,9 +443,16 @@ protected int getSelectedWarehouseId() {
 	@Override
 	protected void initParameters() {
 		int M_Warehouse_ID = Env.getContextAsInt(Env.getCtx(), p_WindowNo, "M_Warehouse_ID");
-		int M_PriceList_ID = Env.getContextAsInt(Env.getCtx(), p_WindowNo, "M_PriceList_ID");
 		
-		int M_PriceList_Version_ID = findPLV (M_PriceList_ID);
+		String ignorePriceList = Env.getContext(Env.getCtx(), p_WindowNo,
```
CarlosRuiz-globalqss, line 447: Checking the occurrences of M_Product_ID field, I tend to think the majority of these fields are not related to the default price list.
I can see about 70 editable M_Product_ID fields like:
* Accounting Schema > Account Schema Dimensions
* Cash Plan > Cash Plan Line
* Commission > Commission Line
* Cost Adjustment > Cost Adjustment Line
* Customer RMA > RMA Line
* GL Distribution > Distribution
* GL Distribution > Line
* GL Journal > Line
* Import GL Journal > General Ledger
* and a long etcetera

Because there are more cases where price list must be ignored, I think the variable must be named the opposite UsePriceListInProductInfo (instead of Ignore) - and the code checks the opposite (that UsePriceListInProductInfo=Y)

In this case, instead of changing the 4 windows:
* Physical Inventory
* Price List
* Inventory Move
* Internal Use Inventory
as in the migration script

It would be the opposite, just windows that must USE the price list will have the predefined context variable, these would be:
* Sales Order
* Purchase Order
* Import Order
* Invoice Vendor
* Invoice Customer
* Import Invoice
* Requisition

I think there are not more than these 7, but if found any other that requires it, is easy to just add that.

CarlosRuiz-globalqss (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoProductWindow.java
hengsin: Hi @CarlosRuiz-globalqss ,
An minor issue that arise from this is it does break backward compatibility - I will have to add that to my custom sales order ad_window definition or it will break the product lookup that's working prior to this change.

CarlosRuiz-globalqss: that's correct - in such case the update must not be directly to the ad_window_id - but we can do the update to any window containing tabs for C_Order, C_Invoice, I_Order, I_Invoice and M_Requisition

dpansheriya: @CarlosRuiz-globalqss took notes, Let me change and submit new PR.

dpansheriya:
I am closing this and creating new PR.

PR #740, opened 2021-06-24
IDEMPIERE-4227 Unable to use 2 print format item with the same AD_Column_ID and different formatting

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4227


CarlosRuiz-globalqss (approved):
Tested - this PR solves IDEMPIERE-4227 and also IDEMPIERE-3402

PR #741, opened 2021-06-24
IDEMPIERE-4847 TableIndexValidate: include only active columns in table index

Soundbytes:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4847

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MTableIndex.java
+++ b/org.adempiere.base/src/org/compiere/model/MTableIndex.java
@@ -119,8 +119,7 @@ public class MTableIndex extends X_AD_TableIndex {
 	 */
 	public MIndexColumn[] getColumns(boolean reload)
 	{
-		String whereClause = MIndexColumn.COLUMNNAME_AD_TableIndex_ID + "=?";
-		return getColumns(reload, whereClause);
+		return getColumns(reload, false);
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/model/MTableIndex.java
```diff
@@ -117,10 +119,31 @@ public MTableIndex(MTable parent, String name)
 	 */
 	public MIndexColumn[] getColumns(boolean reload)
 	{
-		if (m_columns != null && !reload)
+		String whereClause = MIndexColumn.COLUMNNAME_AD_TableIndex_ID + "=?";
+		return getColumns(reload, whereClause);
```
hengsin, line 123: this should be simplify to just return getColumns(reload, false);

org.adempiere.base/src/org/compiere/model/MTableIndex.java
```diff
@@ -149,41 +172,21 @@ private String createDDL()
 		StringBuilder sql = null;
 		if (!isCreateConstraint())
 		{
-			sql = new StringBuilder("CREATE ");
+			sql = new StringBuilder("CREATE");
 			if (isUnique())
-				sql.append ("UNIQUE ");
-			sql.append("INDEX ").append (getName())
+				sql.append (" UNIQUE");
+			sql.append(" INDEX ").append (getName())
```
hengsin, line 178: the white space changes for CREATE, UNIQUE AND INDEX is unnecessary.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MTableIndex.java
Soundbytes: The sql string does not contain any double spaces. The only redundant space is before the opening parenthesis of the columnn list. I did not remove it because it enhances readability for when debugging.

hengsin: you confuse me here, where did I mention anything on "double spaces" here ? I'm just those changes above is not needed at all.

org.adempiere.base/src/org/compiere/model/MTableIndex.java
Soundbytes: done.

org.adempiere.base/src/org/compiere/model/MTableIndex.java
Soundbytes: IC.  done.

hengsin (approved):
looks good

PR #745, opened 2021-06-28
IDEMPIERE-4828-posting of Allocation with a AR Credit Memos with disc…

pmassing:
…ount, the tax-correction has the wrong sign

Control Sheet (IDEMPIERE-4828.ods) to the Jira-Ticket added.
Because it's not that easy to understand.. Tried to create some automatic tests, with documentation, about the accounting facts.
Tax-correction for Discount and write-off.

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
+++ b/org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
@@ -32,6 +32,7 @@ import org.compiere.model.MAllocationHdr;
 import org.compiere.model.MAllocationLine;
 import org.compiere.model.MCashLine;
 import org.compiere.model.MConversionRate;
+import org.compiere.model.MDocType;
 import org.compiere.model.MFactAcct;
 import org.compiere.model.MInvoice;
 import org.compiere.model.MInvoiceLine;
@@ -1856,6 +1857,11 @@ class Doc_AllocationTax
 				return false;
 			}
 
+			int doctypeid = -1; // ensure standard behavior
+			int doctypecm =  0;
+			Doc doc = DocManager.getDocument(as, MInvoice.Table_ID, factAcct.getRecord_ID(), line.getPO().get_TrxName());
+			if(doc!=null)
+				doctypeid = doc.getC_DocType_ID();
 
 //			Discount Amount
 			if (m_DiscountAmt.signum() != 0)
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
```diff
@@ -1869,16 +1875,33 @@ public boolean createEntries (MAcctSchema as, Fact fact, DocLine line)
 					{
 						//for sales actions
 						if (m_IsSOTrx) {
-							fact.createLine (line, m_DiscountAccount,
-								as.getC_Currency_ID(), amount, null);
-							fact.createLine (line, taxAcct,
-								as.getC_Currency_ID(), null, amount);
+							doctypecm = MDocType.getDocType(MDocType.DOCBASETYPE_ARCreditMemo);
```
CarlosRuiz-globalqss, line 1878: This is wrong here.
This line is getting the doctype for a credit memo, but there can be more than one and the method is returning the first.

Instead of comparing a doctype, it must check for the DocBaseType directly.

To make it fail I created a second "AR Credit Memo 2" document, simply copying it from "AR Credit Memo" - then the code here fails to realize it is a credit memo.

There is similar code below to fix.

CarlosRuiz-globalqss (changes requested):
Thanks @pmassing - nice work with the unit test, I checked and it fails before the patch, and pass successfully after the patch.

I made a request for change on a failing case.

org.adempiere.base/src/org/compiere/acct/Doc_AllocationHdr.java
pmassing: Hi @CarlosRuiz-globalqss 
yes, that can cause an error.
I will correct the part of code.
Many Thanks!
Best regards,
patric

pmassing: Changed to verify DocBaseType in DocAllocationHdr

Manual tested with different DockTypes
- DocType=AR Invoice 2 - DocBaseType=AR Invoice
- DocType=AR CreditMemo 2 - DocBaseType=AR CreditMemo

- DocType=AP Invoice 2 - DocBaseType=AP Invoice
- DocType=AP CreditMemo 2 - DocBaseType=AP CreditMemo

hengsin (approved):
pass unit test

CarlosRuiz-globalqss (approved):

PR #746, opened 2021-06-28
IDEMPIERE-3732 : MRefList.getList cannot allow to choose order by

nmicoud:


CarlosRuiz-globalqss:
Hi @nmicoud - I changed the getList method to use the same query constructed in MLookupFactory including too the client ASP filter if required.

Also changed MLookupFactory and here to use the MReference cache (better for performance)

Can you please check and test if this patch works for you?
[PR746_add.txt](https://github.com/idempiere/idempiere/files/6765636/PR746_add.txt)

Regards,

Carlos Ruiz

nmicoud:
Hi @CarlosRuiz-globalqss,

I conduct some tests and found issues in your patch

```
+		if (orderByValue)
+			sql.append(" ORDER BY 2");
+		else
+			sql.append(" ORDER BY 3"); // sort by name/translated name - teo_sarca, [ 1672820 ]
```

should be

```
+		if (orderByValue)
+			sql.append(" ORDER BY 1");
+		else
+			sql.append(" ORDER BY 2"); // sort by name/translated name - teo_sarca, [ 1672820 ]
```


What's the interest of
`ArrayList<Object> params = new ArrayList<Object>();`
as there is no value inside ; can't just pass null in DB.getValueNamePairs ?

Regards,

Nicolas

CarlosRuiz-globalqss:
Thanks @nmicoud - of course you're right, updated patch:
[PR746_add_2.txt](https://github.com/idempiere/idempiere/files/6775556/PR746_add_2.txt)

nmicoud:
Hi @CarlosRuiz-globalqss ,
PR updated
Thanks
Nicolas

CarlosRuiz-globalqss (approved):

PR #750, opened 2021-06-30
IDEMPIERE-4855 Records of Tab level 2 disappear in particular case.

HideakiHagiwara:
I think It seems Good.
But I am not sure whether this improvement best or not.
Please check it.


hengsin:
please check whether pull/751 works for you. your patch activate level 2 tab when it is not yet visible which I think it is not good.

CarlosRuiz-globalqss:
Replaced with https://github.com/idempiere/idempiere/pull/751

PR #751, opened 2021-06-30
IDEMPIERE-4855 Records of Tab level 2 disappear in particular case.

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4855

### DIFF
--- a/migration/i8.2/oracle/202106291641_IDEMPIERE-4851.sql
+++ b/migration/i8.2/oracle/202106291641_IDEMPIERE-4851.sql
@@ -1,0 +1,15 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-4851 Tenant Admin can deactivate his own client.
+-- Jun 29, 2021, 4:39:25 PM CEST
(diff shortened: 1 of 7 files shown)


CarlosRuiz-globalqss (approved):
tested - solved the issue reported

HideakiHagiwara:
Thanks @hengsin for my pull request review and created new pull request.

I checked your pull request and tested some case.
Your pull request seems fine. 

So, I would like to confirm bad point of my pull request.
I would be happy if you could tell me.

I reviewed bad point of my pull request.

I think most bad point of my pull request is to activate All detail tab.
It's a waste of system resources and bad to performance.
I should activate active detail tab only.
Right?

I would be grateful if you could let me know if there are any other bad points or adverse effect as far as you can think of it.

I would like to get a better understanding of the window engine.
:-D

Best Regards,

hengsin:
hi @HideakiHagiwara ,

Yes, that's the issue here. Where possible, we try to activate a detail tab when it is visible (for e.g when level 0 tab is header, level 2 tab is definitely invisible).

Regards,
Low

HideakiHagiwara:
Hi, @hengsin 
Thanks quick response.
And thank you for your teaching.

I would like to keep learning Window Engine and other iDempiere system.
Please let me know if you have another chance.

In appreciation I'll show you around Japan, If you come again Japan.
I knew you came to HAKONE in Japan.
:-D

hengsin:
thanks for the invitation (I've been to Japan twice - Hokkaido and Tokyo). As for now, I can't even travel to my parent's house that's just 30km away :(

PR #752, opened 2021-06-30
IDEMPIERE-4856 Setting an attribute with type reference doesn't work in german / IDEMPIERE-2999

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4856


mschnbeck:
Tested on postgresql. The update did the trick. I'm not sure whether I'm unable to find where to approve this pull request or I'm not allowed to do so.

PR #753, opened 2021-06-30
IDEMPIERE-4854 Bugfixes in InOutGenerate

Soundbytes:
InOutGenerate.createLine(): The Process would sometimes generate
multiple inout lines from a single product lot/asi.
InOutGenerate.prepare(): moved date setting code out of the loop.


hengsin (changes requested):
need test case or unit test

hengsin (approved):
looks good

PR #754, opened 2021-06-30
IDEMPIERE-4837:on Product Info, only defaulting price list when windo…

dpansheriya:
…w has UserPriceListInProductInfo ctx variable set.

As per review comments reimplemented price list defaulting logic.


CarlosRuiz-globalqss (approved):
tested successfully

PR #758, opened 2021-07-01
IDEMPIERE-4860 : MRole.updateAccessRecords insert active records base…

nmicoud:
…d on deactivated objects


hengsin (approved):
looks good

PR #769, opened 2021-07-06
IDEMPIERE-4864 Fix shortcut handling for combobox and ad window

hengsin:
Fix toolbar shortcut is active when it is block by process log dialog

https://idempiere.atlassian.net/browse/IDEMPIERE-4864


CarlosRuiz-globalqss (approved):
tested

PR #771, opened 2021-07-07
IDEMPIERE-3101 implement OAuth2 for mail (gmail, outlook and other mail system)

CarlosRuiz-globalqss:
Split processToken in two public methods to allow calling from external interfaces

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAuthorizationCredential.java
+++ b/org.adempiere.base/src/org/compiere/model/MAuthorizationCredential.java
@@ -55,7 +55,7 @@ public class MAuthorizationCredential extends X_AD_AuthorizationCredential {
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -2433704480923324032L;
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/MAuthorizationCredential.java
```diff
@@ -106,10 +106,33 @@ public String processToken(String code, MPInstance pinstance) {
 				msg = "Process instance parameter for Scopes not found";
 				return msg;
 			}
+			MPInstanceLog pilog = pinstance.addLog(null, 0, null, null, MAuthorizationAccount.Table_ID, 0);
+			msg = processToken(Env.getCtx(), code, paramScope.getP_String(), pilog);
+			pilog.saveEx();
```
d-ruiz: This line should be moved inside processToken "if  (pilog != null)", here it is not clear why you're saving if nothing is changed. Looks also as a side effect of processToken, better to keep it there

d-ruiz (approved):
Changes ok

PR #774, opened 2021-07-08
IDEMPIERE-4858 : CreateTable process

nmicoud:


CarlosRuiz-globalqss (changes requested):
Hi @nmicoud - please check comments in ticket

CarlosRuiz-globalqss:
Hi @nmicoud - I did some tests and still found some things failing

* Failed creating Description column: ** Fill mandatory fields: : Length
* Value and DocumentNo are not usually mandatory

I did those fixes and some minor reordering and renaming of variables in this patch:
[PR774_peer_review.txt](https://github.com/idempiere/idempiere/files/6856873/PR774_peer_review.txt)

Can you please check it - and if satisfied please integrate it into this pull request.

nmicoud:
Hi @CarlosRuiz-globalqss ,

I integrate your changes in the pull request.

Anyway I have some doubts for :
 - Selecting columns by default (but it can easily be overwritten with default value)
 - DocumentNo as non mandatory ; in my mind, it must be mandatory, but I see several places where is optional (and here again is easy to tick it or override the default value)
 
Thanks for your feedback

Regards,

Nicolas

nmicoud:
Hi @CarlosRuiz-globalqss ,
I've tested your changes and I'll integrate them.
Just a doubt for tablename/name/description/data access level / create key column parameters : maybe we should hide them (instead of making them readonly) when process is used to complete an existing table.
wdyt?

CarlosRuiz-globalqss:
Hi @nmicoud,

>  Just a doubt for tablename/name/description/data access level / create key column parameters : maybe we should hide them (instead of making them readonly) when process is used to complete an existing table.

I didn't try that option, if you think it's better for usability please go ahead and add that.

Regards,

Carlos Ruiz

nmicoud:
Hi @CarlosRuiz-globalqss, 

PR udpated with display logic instead of ReadOnly.
I also change the name of the process in the class and added you as author (but without an organization, don't know what to write :-/))

Regards,

Nicolas

CarlosRuiz-globalqss:
Hi @nmicoud 

> PR udpated with display logic instead of ReadOnly.

It didn't work - the parameters are not being filled if not displayed, the process failed with error 
org.adempiere.exceptions.AdempiereException: Fill mandatory fields: : Data Access Level - AccessLevel
because it doesn't get the tablename to know that already exists

nmicoud:
Fixed, in my local DB, there was the default value (from your script) and I didn't notice that when updating them.

CarlosRuiz-globalqss:
Ah I see - I added default value to all those parameters - can you please restore that - maybe one of those parameters is needed and we don't notice in the tests.

nmicoud:
I restore for Tablename it with the last commit ; is working from here.
I think is enough, no?

CarlosRuiz-globalqss:
better safe than sorry - I would prefer to restore the defaults for all those parameters that are coming from the table.

nmicoud:
Indeed

CarlosRuiz-globalqss (approved):
Ready

PR #776, opened 2021-07-09
IDEMPIERE-4871 Added Data Type for Script Column

igorpojzl:
Jira: https://idempiere.atlassian.net/browse/IDEMPIERE-4871

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -1079,7 +1079,8 @@ public class DataEngine
 											value = parseVariable(display, pdc, pd);
 											Interpreter bsh = new Interpreter ();
 											try {
-												display = bsh.eval(display).toString();
+												value = bsh.eval(value.toString());
+												
 											} catch (EvalError e) {
 												log.severe(e.getMessage());
 											}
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1042,26 +1044,51 @@ else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
...
+											display = display.replace("@SCRIPT", "");
+											value = parseVariable(display, pdc, pd);
+											Interpreter bsh = new Interpreter ();
+											try {
+												display = bsh.eval(display).toString();
+											} catch (EvalError e) {
+												log.severe(e.getMessage());
+											}
+										}
+										pde = new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(), (Serializable) value, displayType, pdc.getFormatPattern());
```
hengsin, line 1087: This is incomplete (and hence inconsistent), should handle the code path for "@SCRIPT" too. Similar to the db code path, a couple of instanceof check should be sufficient to determine the return type from bsh.eval(display).

hengsin (changes requested):

org.adempiere.base/src/org/compiere/print/DataEngine.java
igorpojzl: Comited fix for eval function, missed change of variable. Path for @SCRIPT is limited by method parseVariable which Can get only use BigDecimal Columns atm and return only numeric String to evaluation so it will be always numeric.

hengsin (approved):
looks good

PR #779, opened 2021-07-13
IDEMPIERE-4873 Jasper Report Starter Refactoring

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4873


CarlosRuiz-globalqss (approved):
Tested attachment approach with three jasper basic reports for testing.

PR #783, opened 2021-07-20
IDEMPIERE-4866 : Multiple smtp per tenant

nmicoud:
1st version

### DIFF
--- a/migration/i8.2z/oracle/202107200830_IDEMPIERE-4866.sql
+++ b/migration/i8.2z/oracle/202107200830_IDEMPIERE-4866.sql
@@ -1,0 +1,483 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Jul 20, 2021, 8:18:16 AM CEST
+INSERT INTO AD_Table (AD_Table_ID,Name,TableName,LoadSeq,AccessLevel,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,IsSecurityEnabled,IsDeleteable,IsHighVolume,IsView,EntityType,ImportTable,IsChangeLog,ReplicationType,CopyColumnsFromTable,IsCentrallyMaintained,AD_Table_UU,Processing,DatabaseViewDrop,CopyComponentsFromView,CreateWindowFromTable) VALUES (200280,'AD_SMTP','AD_SMTP',0,'3',0,0,'Y',TO_DATE('2021-07-20 08:18:15','YYYY-MM-DD HH24:MI:SS'),0,TO_DATE('2021-07-20 08:18:15','YYYY-MM-DD HH24:MI:SS'),0,'N','Y','N','N','D','N','Y','L','N','Y','dd5e80b1-9206-4f13-9d9a-a225ed3d7b78','N','N','N','N')
+;
+
+-- Jul 20, 2021, 8:18:16 AM CEST
+INSERT INTO AD_Sequence (Name,CurrentNext,IsAudited,StartNewYear,Description,IsActive,IsTableID,AD_Client_ID,AD_Org_ID,Created,CreatedBy,Updated,UpdatedBy,AD_Sequence_ID,IsAutoSequence,StartNo,IncrementNo,CurrentNextSys,AD_Sequence_UU) VALUES ('AD_SMTP',1000000,'N','N','Table AD_SMTP','Y','Y',0,0,TO_DATE('2021-07-20 08:18:16','YYYY-MM-DD HH24:MI:SS'),0,TO_DATE('2021-07-20 08:18:16','YYYY-MM-DD HH24:MI:SS'),0,200349,'Y',1000000,1,200000,'1ec73ebd-f1cc-4563-9789-b2656dd6df87')
+;
+
+-- Jul 20, 2021, 8:18:27 AM CEST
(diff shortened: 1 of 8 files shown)


migration/i8.2z/oracle/202107200830_IDEMPIERE-4866.sql
```diff
@@ -0,0 +1,483 @@
...
+;
+
+-- Jul 20, 2021, 8:22:35 AM CEST
+INSERT INTO AD_Window (AD_Window_ID,Name,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,WindowType,Processing,EntityType,IsSOTrx,IsDefault,WinHeight,WinWidth,IsBetaFunctionality,AD_Window_UU) VALUES (200118,'SMTP Server',0,0,'Y',TO_DATE('2021-07-20 08:22:34','YYYY-MM-DD HH24:MI:SS'),0,TO_DATE('2021-07-20 08:22:34','YYYY-MM-DD HH24:MI:SS'),0,'M','N','D','Y','N',0,0,'N','5e08444e-edc9-40e0-8f12-b71bdd4111b2')
+;
+
+-- Jul 20, 2021, 8:23:15 AM CEST
+INSERT INTO AD_Tab (AD_Tab_ID,Name,AD_Window_ID,SeqNo,IsSingleRow,AD_Table_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,HasTree,IsInfoTab,IsTranslationTab,IsReadOnly,Processing,ImportFields,TabLevel,IsSortTab,EntityType,IsInsertRecord,IsAdvancedTab,AD_Tab_UU,TreeDisplayedOn,IsLookupOnlySelection,IsAllowAdvancedLookup,MaxQueryRecords) VALUES (200298,'SMTP Server',200118,10,'N',200280,0,0,'Y',TO_DATE('2021-07-20 08:23:15','YYYY-MM-DD HH24:MI:SS'),0,TO_DATE('2021-07-20 08:23:15','YYYY-MM-DD HH24:MI:SS'),0,'N','N','N','N','N','N',0,'N','D','Y','N','bdf27ca5-26af-456d-a647-6bad3d5cad38','B','N','Y',0)
```
CarlosRuiz-globalqss, line 144: for consistency please all new windows are better created with IsSingleRow=Y

org.adempiere.base.process/src/org/compiere/process/EMailTest.java
```diff
@@ -51,28 +58,80 @@ protected void prepare ()
 	 */
 	protected String doIt () throws Exception
 	{
-		MClient client = MClient.get (getCtx(), p_AD_Client_ID);
-		if (log.isLoggable(Level.INFO)) log.info(client.toString());
+		int tableID = getProcessInfo().getTable_ID();
+		System.out.println(tableID);
```
CarlosRuiz-globalqss, line 62: no System.out - better log....

migration/i8.2z/oracle/202107200830_IDEMPIERE-4866.sql
```diff
@@ -0,0 +1,483 @@
...
+;
+
+-- Jul 20, 2021, 8:21:06 AM CEST
+INSERT INTO AD_Element (AD_Element_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,ColumnName,Name,Description,Help,PrintName,EntityType,AD_Element_UU) VALUES (203527,0,0,'Y',TO_DATE('2021-07-20 08:20:53','YYYY-MM-DD HH24:MI:SS'),0,TO_DATE('2021-07-20 08:20:53','YYYY-MM-DD HH24:MI:SS'),0,'UsedByEmailAndOrDomain','Used by emails and/or domains',NULL,NULL,'Used by emails and/or domains','D','a1d45340-7b37-4a20-aecb-7520ed02e32c')
```
CarlosRuiz-globalqss, line 120: Please add a description to this new Element, what is expected to be there? is it comma separated? can email and domains be combined?

See also comment below in MSMTP, maybe this must be single-valued

org.adempiere.base/src/org/compiere/model/MSMTP.java
```diff
@@ -0,0 +1,88 @@
...
+	 * @param ctx     context
+	 * @param rs      ResultSet
+	 * @param trxName transaction
+	 */
+	public MSMTP(Properties ctx, ResultSet rs, String trxName) {
+		super(ctx, rs, trxName);
+	} // MSMTP
+	
+	/** Search a SMTP server for the sender (1st try with the full address, 2nd try with domain only) */
+	public static int getSmtpID(int clientID, String from, String trxName) {
```
CarlosRuiz-globalqss, line 64: - use log instead of System.out

- as you mentioned in the ticket this must be cached

Now, thinking about this approach, if I understand you right you can combine emails and also different domains in the same record - probably you want that separated by something (commas)

But I think that's wrong, if you register carlos@idempiere.com - then that record will be used also for nicolas@idempiere.com because of the domain discovery.

Maybe a better way - and more straight to use with cache - is that the field "Used by emails and/or domains" must be renamed to "Used by email or domain".  And register single record for complete email or for a full domain, that way I think the query must be searched with equal (=) operator instead of LIKE

If several users are intended to send emails with different SMTP then you can register several records.

The method must check first for the EMail (cached) and then check for the domain part (cached too).

nmicoud: Hi,
I'm trying to work on this, but i got lots of errors in Eclipse when working on branch 4866 (master is ok).
 (main on is "Failure to find org.eclipse.tycho:tycho-source-plugin:jar:2.4.0-SNAPSHOT"...) ; so I'll try to fix that but if not possible, I think I'll close this PR and create a new one based on latest sources

nmicoud:
Closing this PR and opening https://github.com/idempiere/idempiere/pull/884

PR #788, opened 2021-07-21
IDEMPIERE-4886 Synchronize multiple columns on ALTER

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4886

Created a process parameter "Date From" - when empty or when the table is new it behaves the same as before.

When filled and the table already exists in DB - then all the columns modified after that "Date From" are synchronized.


CarlosRuiz-globalqss:
Pushing as tested successfully by @nmicoud on oracle

PR #797, opened 2021-07-26
IDEMPIERE-4808 : Payment Allocation should be IsSOTrx neutral - fix D…

nmicoud:
…efault SQL variable parse failed


CarlosRuiz-globalqss (approved):
tested

PR #798, opened 2021-07-26
IDEMPIERE-4889 Performance - Attachment management

CarlosRuiz-globalqss:
The approach is backward compatible, the actual plugins will keep working without changes.

For plugins that want to use the new delayed load implementation is simple:
* do not load the entries data in loadLOBData - this can be done using the new constructor MAttachmentEntry(MAttachment, String, int, Object)
* implement the load of the entry data in loadLOBDataEntry

The new approach was implemented and tested in AttachmentFileSystem, for attachments with many files, or with very big files the performance and memory footprint is reduced because the load of the data from files is done just when needed.

The unchanged AttachmentDBSystem was tested too to ensure everything keeps working without changes there.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/IAttachmentLazyDataSource.java
+++ b/org.adempiere.base/src/org/compiere/model/IAttachmentLazyDataSource.java
@@ -1,0 +1,38 @@
+/***********************************************************************
+ * This file is part of iDempiere ERP Open Source                      *
+ * http://www.idempiere.org                                            *
+ *                                                                     *
+ * Copyright (C) Contributors                                          *
+ *                                                                     *
+ * This program is free software; you can redistribute it and/or       *
+ * modify it under the terms of the GNU General Public License         *
+ * as published by the Free Software Foundation; either version 2      *
+ * of the License, or (at your option) any later version.              *
+ *                                                                     *
+ * This program is distributed in the hope that it will be useful,     *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of      *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the        *
+ * GNU General Public License for more details.                        *
+ *                                                                     *
+ * You should have received a copy of the GNU General Public License   *
+ * along with this program; if not, write to the Free Software         *
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,          *
+ * MA 02110-1301, USA.                                                 *
+ *                                                                     *
+ * Sponsor:                                                            *
+ * - FH                                                                *
+ * Contributors:                                                       *
+ * - Carlos Ruiz                                                       *
+ **********************************************************************/
+
+package org.compiere.model;
+
+/**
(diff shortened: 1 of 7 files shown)


org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
```diff
@@ -71,21 +60,41 @@ public MAttachmentEntry (String name, byte[] data)
...
+
+	/**
+	 * Constructor for delayed load
+	 * If you use this constructor, it is required that your storage provider implements the method loadLOBDataEntry for delayed load
+	 * @param parent
+	 * @param name
+	 * @param index
+	 * @param extraObj
+	 */
+	public MAttachmentEntry (MAttachment parent, String name, int index, Object extraObj) {
```
hengsin, line 72: Hi @CarlosRuiz-globalqss , changing extraObj to an interface would make this easier to understand and implement. For e.g, ILazyDataSource {public byte[] getData(); public boolean isLoader();} and you can then have FileLazyDataSource implementation that internally holds an File object and lazy load it when getData is call.

hengsin (changes requested):

CarlosRuiz-globalqss:
Hi @hengsin - ready - thanks for your suggestion - it works fine.

org.adempiere.base/src/org/compiere/model/IAttachmentLazyDataSource.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.compiere.model;
+
+/**
+ *	IDEMPIERE-4889
+ * 	@author Carlos Ruiz - globalqss
+ */
+public interface IAttachmentLazyDataSource {
+
+	public byte[] getData();
```
hengsin, line 36: should have javadoc here

org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
```diff
@@ -157,13 +170,13 @@ public String toString ()
 	public String toStringX ()
 	{
 		StringBuilder sb = new StringBuilder (m_name);
-		if (m_data != null)
+		if (getData() != null)
```
hengsin, line 173: this make toStringX() call rather expensive. perhaps not an issue if this is rarely use.

org.adempiere.base/src/org/compiere/model/IAttachmentLazyDataSource.java
CarlosRuiz-globalqss: ready

org.adempiere.base/src/org/compiere/model/MAttachmentEntry.java
CarlosRuiz-globalqss: Checked - is just used in two points in core:
MAttachment.addEntry when log level is FINE
WAttachment.displaySelected when log level is CONFIG

So, I think is safe.

hengsin (approved):
looks good.

PR #799, opened 2021-07-27
IDEMPIERE-4059 Print format editor

hengsin:
- add maven for eclipse configs
- remove maximize/restore button

https://idempiere.atlassian.net/browse/IDEMPIERE-4059

### DIFF
--- a/org.idempiere.printformat.editor/.project
+++ b/org.idempiere.printformat.editor/.project
@@ -26,14 +26,14 @@
 			</arguments>
 		</buildCommand>
 		<buildCommand>
-			<name>org.eclipse.babel.editor.rbeBuilder</name>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
 			<arguments>
 			</arguments>
 		</buildCommand>
 	</buildSpec>
 	<natures>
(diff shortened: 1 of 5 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I tested this successfully, however I think a better approach could be with something like this:
		setSizable(true);
		setClosable(true);
		setMaximizable(true);
		setMaximized(true);
		appendChild(mainLayout);
		LayoutUtils.addSclass("tab-editor-form-content", mainLayout);
		setBorder("normal");
		setHeight("80%");
		setWidth("80%");

That makes the window resizable and clicking the restore button makes the window just a bit smaller.

Another issue was reported by @nmicoud - I tested and effectively changing Name and Print Name is not affecting the item.

And an idea - maybe hard to implement - would be great if the button could be in the report viewer directly and behave as the report wizard, refreshing the preview when closed.

hengsin:
Hi @CarlosRuiz-globalqss , ready for testing (I didn't implement the report viewer suggestion).

CarlosRuiz-globalqss (approved):
tested

PR #800, opened 2021-07-27
IDEMPIERE-1026 Button on table to create automatically the window

CarlosRuiz-globalqss:
Set readonly for the common document fields - useful for IDEMPIERE-4858


d-ruiz (approved):
Looks good

PR #808, opened 2021-07-30
IDEMPIERE-4898

Dorithur:
https://idempiere.atlassian.net/browse/IDEMPIERE-4898

Fix for "AND" Join Element nodes on Workflow.


hengsin (changes requested):
see comments in ticket.

hengsin:
closing this as we can't get a response from the submitter of this PR

PR #810, opened 2021-07-31
IDEMPIERE-4867 Updating moving node by PO

ArmenVH:
Allow to handle change events in the AD_TreeNodeXX tables.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
@@ -202,7 +202,7 @@ public class ADTreeOnDropListener implements EventListener<Event> {
 		PO mnPO = new Query(Env.getCtx(), mTree.getNodeTableName(), whereClause.toString(), trxName).first();
 		mnPO.set_ValueNoCheck("Parent_ID", mtnParentNode.getNode_ID());
 		mnPO.set_ValueNoCheck("SeqNo", NodeIndex);
-		mnPO.save(trxName);
+		mnPO.saveEx(trxName);
 	}
 	
 	class MenuListener implements EventListener<Event> {
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
@@ -202,7 +202,8 @@ public class ADTreeOnDropListener implements EventListener<Event> {
 		PO mnPO = new Query(Env.getCtx(), mTree.getNodeTableName(), whereClause.toString(), trxName).first();
 		mnPO.set_ValueNoCheck("Parent_ID", mtnParentNode.getNode_ID());
 		mnPO.set_ValueNoCheck("SeqNo", NodeIndex);
-		mnPO.saveEx(trxName);
+		if (mnPO.is_Changed()) 
+			mnPO.saveEx(trxName);
 	}
 	
 	class MenuListener implements EventListener<Event> {
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
@@ -18,6 +18,8 @@ import org.adempiere.webui.util.TreeUtils;
 import org.adempiere.webui.window.FDialog;
 import org.compiere.model.MTree;
 import org.compiere.model.MTreeNode;
+import org.compiere.model.PO;
+import org.compiere.model.Query;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
@@ -147,6 +149,9 @@ public class ADTreeOnDropListener implements EventListener<Event> {
 		
 		//  insert
 		newParent = treeModel.addNode(newParent, movingNode, index);
+
+		MTreeNode mtnMovingNode = (MTreeNode) movingNode.getData();
+
 		
 		int path[] = treeModel.getPath(movingNode);
 		if (TreeUtils.isOnInitRenderPosted(tree) || tree.getTreechildren() == null)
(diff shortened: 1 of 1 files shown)


hengsin (changes requested):
this is not correct, if you are using PO to perform the update then you should remove the sql update script.

ArmenVH:
>then you should remove the sql update script.

You mean that the sql update script should NOT be run for the node being moved (movingNode), am I right ?

hengsin:
> > then you should remove the sql update script.
> 
> You mean that the sql update script should NOT be run for the node being moved (movingNode), am I right ?

Yes, if you have already perform the update using PO, why should you update it again with SQL script ?

ArmenVH:
> > > then you should remove the sql update script.
> > 
> > 
> > You mean that the sql update script should NOT be run for the node being moved (movingNode), am I right ?
> 
> Yes, if you have already perform the update using PO, why should you update it again with SQL script ?

I completely agree with your idea - the changes are presented.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
```diff
@@ -170,15 +175,20 @@ private void moveNode(DefaultTreeNode<Object> movingNode, DefaultTreeNode<Object
...
+					StringBuilder sql = new StringBuilder("UPDATE ");
+					sql.append(mTree.getNodeTableName())
+						.append(" SET Parent_ID=").append(oldMParent.getNode_ID())
+						.append(", SeqNo=").append(i)
+						.append(", Updated=getDate()")
+						.append(" WHERE AD_Tree_ID=").append(mTree.getAD_Tree_ID())
+						.append(" AND Node_ID=").append(md.getNode_ID());
+					if (log.isLoggable(Level.FINE)) log.fine(sql.toString());
+					no = DB.executeUpdate(sql.toString(),trx.getTrxName());
+				}
```
hengsin, line 191: this is both incomplete and confusing. incomplete as you still have a mix of po and sql update. confusing as why the need to have mtnMovingNode and the added condition? shouldn't it be just a direct replacement of sql script update with PO update without the added variables and condition ?

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
ArmenVH: > this is both incomplete and confusing. incomplete as you still have a mix of po and sql update. confusing as why the need to have mtnMovingNode and the added condition? shouldn't it be just a direct replacement of sql script update with PO update without the added variables and condition ?

The update is performed in a loop - that is, for all nodes of the parent to which the movingNode belongs. If I replace SQL update with PO then change events will be generated for all nodes - but my goal is to generate the change event only for the node being moved, therefore I added a condition to check that the current node that will be updated is equal to the moving node and only in this case I use the PO.
Please correct me if I'm going in the wrong direction.
updatePO

hengsin: I think it is fine to use PO update for all node that's being updated here. It is perhaps slower but it is more consistent in that we have event and change log for all node that have been updated here.

ArmenVH: Agree, but how we can provide event generation ONLY for movingNode ?

hengsin: You don't have to. In the event handler, you can determine whether the PO is the movingNode by checking which value of the PO have been changed.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
```diff
@@ -213,6 +196,15 @@ private void moveNode(DefaultTreeNode<Object> movingNode, DefaultTreeNode<Object
 		}
 	}
 	
+	private void updateNodePO(MTreeNode mtnParentNode, MTreeNode mtnMovingNode, int NodeIndex, String trxName) {
+		StringBuilder whereClause = new StringBuilder("AD_Tree_ID=").append(mTree.getAD_Tree_ID())
+				.append(" AND Node_ID=").append(mtnMovingNode.getNode_ID());
+		PO mnPO = new Query(Env.getCtx(), mTree.getNodeTableName(), whereClause.toString(), trxName).first();
+		mnPO.set_ValueNoCheck("Parent_ID", mtnParentNode.getNode_ID());
+		mnPO.set_ValueNoCheck("SeqNo", NodeIndex);
+		mnPO.save(trxName);
```
hengsin, line 205: Please use saveEx(trxName) instead since save(trxName) can fail silently.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
ArmenVH: Changes have been made.
Thank you for your help and advice.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
```diff
@@ -213,6 +196,15 @@ private void moveNode(DefaultTreeNode<Object> movingNode, DefaultTreeNode<Object
 		}
 	}
 	
+	private void updateNodePO(MTreeNode mtnParentNode, MTreeNode mtnMovingNode, int NodeIndex, String trxName) {
+		StringBuilder whereClause = new StringBuilder("AD_Tree_ID=").append(mTree.getAD_Tree_ID())
+				.append(" AND Node_ID=").append(mtnMovingNode.getNode_ID());
+		PO mnPO = new Query(Env.getCtx(), mTree.getNodeTableName(), whereClause.toString(), trxName).first();
+		mnPO.set_ValueNoCheck("Parent_ID", mtnParentNode.getNode_ID());
+		mnPO.set_ValueNoCheck("SeqNo", NodeIndex);
+		mnPO.saveEx(trxName);
```
hengsin, line 205: tested working fine, just need one small change, please replace mnPO.saveEx(trxName) with:

if (mnPO.is_Changed())
       mnPO.saveEx(trxName);

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
ArmenVH: Changes have been made.
Thanks.

hengsin (approved):
tested

PR #811, opened 2021-08-02

Dorithur:
SQL statement added as an option to define WF Next Node Condition


hengsin (changes requested):
see comments in ticket

PR #813, opened 2021-08-03

Dorithur:
SQL statement added as an option to define WF Next Node Condition.

New Pull Request due to mistake in previous one.

### DIFF
--- a/migration/i8.2z/oracle/202108021638_IDEMPIERE-4900.sql
+++ b/migration/i8.2z/oracle/202108021638_IDEMPIERE-4900.sql
@@ -1,0 +1,41 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Aug 2, 2021, 3:42:37 PM MSK
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,IsHtml) VALUES (214567,0,'SQLStatement',706,'SQLStatement',4000,'N','N','N','N','N',0,'N',10,0,0,'Y',TO_DATE('2021-08-02 15:42:36','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2021-08-02 15:42:36','YYYY-MM-DD HH24:MI:SS'),100,50028,'Y','N','D','N','N','N','Y','a3885450-50d5-41b6-a7f4-8cd4bcc9e465','Y',0,'N','N','N')
+;
+
+-- Aug 2, 2021, 3:42:44 PM MSK
+ALTER TABLE AD_WF_NextCondition ADD SQLStatement VARCHAR2(4000 CHAR) DEFAULT NULL 
+;
+
+-- Aug 2, 2021, 3:42:59 PM MSK
+INSERT INTO AD_Field (AD_Field_ID,Name,AD_Tab_ID,AD_Column_ID,IsDisplayed,DisplayLength,SeqNo,IsSameLine,IsHeading,IsFieldOnly,IsEncrypted,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,IsReadOnly,IsCentrallyMaintained,EntityType,AD_Field_UU,IsDisplayedGrid,SeqNoGrid,ColumnSpan) VALUES (206737,'SQLStatement',644,214567,'Y',4000,130,'N','N','N','N',0,0,'Y',TO_DATE('2021-08-02 15:42:58','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2021-08-02 15:42:58','YYYY-MM-DD HH24:MI:SS'),100,'N','Y','D','f78eaea8-558a-4d74-8c50-26c59468c3a2','Y',130,5)
+;
+
+-- Aug 2, 2021, 3:43:09 PM MSK
+UPDATE AD_Field SET IsDisplayed='Y', SeqNo=110, AD_Reference_Value_ID=NULL, AD_Val_Rule_ID=NULL, XPosition=1, IsToolbarButton=NULL,Updated=TO_DATE('2021-08-02 15:43:09','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Field_ID=206737
+;
+
+-- Aug 2, 2021, 3:43:35 PM MSK
+UPDATE AD_Field SET DisplayLogic='@SQLStatement@ = ''''', AD_Reference_Value_ID=NULL, AD_Val_Rule_ID=NULL, IsToolbarButton=NULL,Updated=TO_DATE('2021-08-02 15:43:35','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Field_ID=10101
+;
+
+-- Aug 2, 2021, 3:44:18 PM MSK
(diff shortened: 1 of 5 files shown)


org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -114,15 +116,35 @@ public boolean isOr()
...
+	{
+		return evaluate(activity.getPO());
+	}	//	evaluate
+	
+	/**
+	 * 	Evaluate Condition
+	 * 	@param po PO
+	 *	@return true if true
+	 */
+	public boolean evaluate (PO po)
```
CarlosRuiz-globalqss, line 128: Is there any advantage adding this method?  I think is not necessary.

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -342,5 +364,40 @@ public MWFNextCondition markImmutable() {
...
+	
+	/**
+	 * Executes SQLStatement of Record and returns result as Object
+	 * @param Record_ID
+	 * @return SQL statement execution result
+	 * @throws Exception
+	 */
+	private Object getSqlResult(int Record_ID) throws Exception {
+		String sqlStatement = (String) get_Value("SQLStatement");
+		String tableName = getAD_Column().getAD_Table().getTableName();
```
CarlosRuiz-globalqss, line 376: This construction is very bad for performance
It is always better to use the cached objects from table and column to obtain the information.

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -342,5 +364,40 @@ public MWFNextCondition markImmutable() {
...
+		String pkName = tableName + "_ID"; // primary key name
```
CarlosRuiz-globalqss, line 377: not every table has primary key - so maybe would be better to check for this, also some tables have _ID different than table+_ID

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -342,5 +364,40 @@ public MWFNextCondition markImmutable() {
...
+
+		Object result = null;
+
+		String resultSql = String.format("SELECT (%s) FROM %s WHERE %s = %d", sqlStatement, tableName, pkName, Record_ID);
```
CarlosRuiz-globalqss, line 381: this looks strange - I thought it was just to run the SQL statement straight there
I haven't tested but I think the construction maybe is strange and different to the other SQLStatement approach on the application.
Some examples maybe?

migration/i8.2z/oracle/202108021638_IDEMPIERE-4900.sql
```diff
@@ -0,0 +1,41 @@
...
+;
+
+-- Aug 2, 2021, 3:42:44 PM MSK
+ALTER TABLE AD_WF_NextCondition ADD SQLStatement VARCHAR2(4000 CHAR) DEFAULT NULL 
+;
+
+-- Aug 2, 2021, 3:42:59 PM MSK
+INSERT INTO AD_Field (AD_Field_ID,Name,AD_Tab_ID,AD_Column_ID,IsDisplayed,DisplayLength,SeqNo,IsSameLine,IsHeading,IsFieldOnly,IsEncrypted,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,IsReadOnly,IsCentrallyMaintained,EntityType,AD_Field_UU,IsDisplayedGrid,SeqNoGrid,ColumnSpan) VALUES (206737,'SQLStatement',644,214567,'Y',4000,130,'N','N','N','N',0,0,'Y',TO_DATE('2021-08-02 15:42:58','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2021-08-02 15:42:58','YYYY-MM-DD HH24:MI:SS'),100,'N','Y','D','f78eaea8-558a-4d74-8c50-26c59468c3a2','Y',130,5)
```
CarlosRuiz-globalqss, line 13: Any field involving SQL Statement or classnames must be made available JUST for Advanced Roles (this is, SuperUser in Tenant)
So AD_Field.IsAdvancedField must be Y
Or alternatively the ReadOnlyLogic must be @#ShowAdvanced@=N

CarlosRuiz-globalqss (changes requested):
Code peer review - however I think the implementation would be better different - please check comments in ticket

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
Dorithur: It's used in my new plugin, which is based on this modifications.
Maybe I should commit this method separately.

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
Dorithur: You mean that I need to use M-class for that?

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
Dorithur: I've seen this construction many times in iDempiere core. I thought that it's a simple and relatively reliable method.

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
Dorithur: This approach is similar to Virtual Columns. Code is in org.compiere.model.GridTable.createSelectSql() method.
It uses query from SQLStatement as a subquery in SELECT statement to give user an ability to refer to current table.

For example if you need to get some Name from C_BPartner table while WF setted to work in AD_User, you can type:
`SELECT Name FROM C_BPartner WHERE C_BPartner_ID = AD_User.C_BPartner_ID`

And it will be transformed to:
`SELECT **(SELECT Name FROM C_BPartner WHERE C_BPartner_ID = AD_User.C_BPartner_ID)** FROM AD_User WHERE AD_User_ID = 1000000`

Otherwise user couldn't refer to columns of table from WF in this query.

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
CarlosRuiz-globalqss: no problem - if you need it is OK, I was just wondering if it was necessary

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
CarlosRuiz-globalqss: There is a method to check that in MTable.isSingleKey() - and I think (not tested) you can obtain the key with MTable.getKeyColumns()[0]

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -114,15 +116,35 @@ public boolean isOr()
...
 			throw new IllegalStateException("No Column defined - " + this);
 			
-		PO po = activity.getPO();
 		if (po == null || po.get_ID() == 0)
 			throw new IllegalStateException("Could not evaluate " + po + " - " + this);
 		//
-		Object valueObj = po.get_ValueOfColumn(getAD_Column_ID());
+		Object valueObj = null;
+
+		if (get_Value("SQLStatement") != null && !get_Value("SQLStatement").equals("")) {
```
hengsin, line 138: this should be changed to getSQLStatement()

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -342,5 +364,40 @@ public MWFNextCondition markImmutable() {
...
 	}
+	
+	/**
+	 * Executes SQLStatement of Record and returns result as Object
+	 * @param Record_ID
+	 * @return SQL statement execution result
+	 * @throws Exception
+	 */
+	private Object getSqlResult(int Record_ID) throws Exception {
+		String sqlStatement = (String) get_Value("SQLStatement");
```
hengsin, line 375: same as above, please change it to getSQLStatement()

org.adempiere.base/src/org/compiere/wf/MWFNextCondition.java
```diff
@@ -342,5 +364,40 @@ public MWFNextCondition markImmutable() {
...
+
+		Object result = null;
+
+		String resultSql = String.format("SELECT (%s) FROM %s WHERE %s = %d", sqlStatement, tableName, pkName, Record_ID);
+
+		PreparedStatement pstmt = null;
+		ResultSet rs = null;
+
+		try {
+			pstmt = DB.prepareStatement(resultSql, null);
```
hengsin, line 387: for better security, call the DB.prepareStatement method that make sure the resultset is readonly.

hengsin:
replace by pull request 981

PR #816, opened 2021-08-05
IDEMPIERE-4902 : Record migration scripts with Dictionnary EntityType

nmicoud:


CarlosRuiz-globalqss (changes requested):
Hi @nmicoud - test didn't work when the AdempiereSys variable is not set.

Can you please integrate the following changes:
[pull816_PR.txt](https://github.com/idempiere/idempiere/files/6938250/pull816_PR.txt)

And after applied, please move the scripts to folder i8.2

The change also implement the suggestion from  @hieplq at IDEMPIERE-2523

CarlosRuiz-globalqss (approved):

PR #825, opened 2021-08-12
IDEMPIERE-4061 isTranslated is not set Yes on new Record default

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4061


hengsin (changes requested):
please resolve conflict with master.

PR #826, opened 2021-08-12
IDEMPIERE-4910 Performance - Discover of UUID column not cached in PO

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4910


hengsin (approved):
looks good

PR #828, opened 2021-08-13
IDEMPIERE-701 Implement solving indirect context

hengsin:
- implement for tab display logic and readonly logic

https://idempiere.atlassian.net/browse/IDEMPIERE-701


CarlosRuiz-globalqss:
Hi @hengsin - can you please advice about test cases for these changes?

I tried these:

* I set the Title as Manager for user GardenAdmin
* In window Sales Order, tab Payment Schedule, set DisplayLogic as `@SalesRep_ID.Title@=Manager`
* in my debug cases the GridTab.get_ValueAsString was never called
* then I set the ReadOnlyLogic with the same value
* the GridTab.get_ValueAsString was called, but it returned null because was trying to look for a column SalesRep_ID in the Payment Schedule tab (that would mean this works just with fields from the same tab and not the parent tab?)

hengsin:
Hi @CarlosRuiz-globalqss ,

For display logic, gridtab.get_valueAsString is not call, that's using ADTabpane.get_ValueAsString

Only same tab ? that shouldn't be the case, will test again to confirm that.

hengsin:
Hi @CarlosRuiz-globalqss ,

1. Tested Sales Order, tab Payment Schedule, set DisplayLogic as @SalesRep_ID.Title@=Manager and it is working as expected.
2. For ReadOnly logic, I've committed a fix for the use of parent grid tab for indirect context resolution.

CarlosRuiz-globalqss (approved):

CarlosRuiz-globalqss:
Hi @hengsin - I did some tests on this and it's working fine now, so it's OK if you want to push that and leave the following suggestions for later:

The problem I can see is that we have a performance issue when using the indirect context, for example:

Setting `@SalesRep_ID.Title@=Manager` in ReadOnlyLogic of the Payment Schedule made this query:
`SELECT Title FROM AD_User WHERE AD_User_ID = ?`
to execute 32 times on every field change

When I did the same test in DisplayLogic it worked better, it was called just like 3 times.

Indeed, I think we have this problem with some actual fields in the database, like "Window, Tab & Field>Field>Included Tab" the DisplayLogic says:
`@AD_Column_ID.AD_Reference_ID@=200127|@AD_Column_ID.AD_Reference_ID@=200128|@AD_Reference_ID@=200127|@AD_Reference_ID@=200128`
so, this query
`SELECT AD_Reference_ID FROM AD_Column WHERE AD_Column_ID = ?`
is executed 14 times when saving a record

maybe is a minor stuff on dictionary, but on heavy usage windows that could make the performance bad.

Not sure about this, but I think we have (had?) a mechanism somewhere to cache those query values when executed the same query during a short time, but don't remember where that was implemented.

hengsin:
hi @CarlosRuiz-globalqss 

Created https://idempiere.atlassian.net/browse/IDEMPIERE-4924 for the performance bug

PR #829, opened 2021-08-13
IDEMPIERE-1351 Quick Entry when Field can not zoom to 2 different win…

hengsin:
…dows

https://idempiere.atlassian.net/browse/IDEMPIERE-1351


CarlosRuiz-globalqss (approved):
tested

PR #830, opened 2021-08-13
IDEMPIERE-4885 Possible bug with uom conversion on invoice when create from shipment

alara123:
We correct the issue keeping the unit from the in/out record and we are only converting the price of the item.

[https://idempiere.atlassian.net/browse/IDEMPIERE-4885](https://idempiere.atlassian.net/browse/IDEMPIERE-4885)

### DIFF
--- a/org.adempiere.ui/src/org/compiere/grid/CreateFromInvoice.java
+++ b/org.adempiere.ui/src/org/compiere/grid/CreateFromInvoice.java
@@ -547,6 +547,14 @@ public abstract class CreateFromInvoice extends CreateFrom
 				if (inoutLine != null)
 				{
 					invoiceLine.setShipLine(inoutLine);		//	overwrites
+					if(invoiceLine.getC_UOM_ID()!=inoutLine.getC_UOM_ID()) {
+						invoiceLine.setC_UOM_ID(inoutLine.getC_UOM_ID());						
+						BigDecimal PriceEntered = MUOMConversion.convertProductFrom (invoice.getCtx(), M_Product_ID, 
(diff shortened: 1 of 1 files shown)


org.adempiere.ui/src/org/compiere/grid/CreateFromInvoice.java
```diff
@@ -547,6 +547,14 @@ else if (M_RMALine_ID != 0)
 				if (inoutLine != null)
 				{
 					invoiceLine.setShipLine(inoutLine);		//	overwrites
+					if(invoiceLine.getC_UOM_ID()!=inoutLine.getC_UOM_ID()) {
+						invoiceLine.setC_UOM_ID(inoutLine.getC_UOM_ID());						
+						BigDecimal PriceEntered = MUOMConversion.convertProductFrom (invoice.getCtx(), M_Product_ID, 
+								inoutLine.getC_UOM_ID(), invoiceLine.getPriceEntered());
+							if (PriceEntered == null)
+								PriceEntered = invoiceLine.getPriceEntered();
```
hengsin, line 555: this isn't a safe fallback, better to raise error here if conversion is not possible.

hengsin (changes requested):
please fix the price conversion fallback

org.adempiere.ui/src/org/compiere/grid/CreateFromInvoice.java
alara123: code updated for throw error when no conversion found

alara123:
code updated for throw error when no conversion found

hengsin (approved):
looks good.

PR #831, opened 2021-08-13
Idempiere 4914

alara123:
Added a client check when retrieving the default bank account


alara123:
Sorry accidentally send a commit from another correction

PR #833, opened 2021-08-14
IDEMPIERE-2963 Error creating report after zoom across

hengsin:
- generate fully qualified where clause for generic zoom across

https://idempiere.atlassian.net/browse/IDEMPIERE-2963


CarlosRuiz-globalqss (approved):
tested

PR #834, opened 2021-08-15
IDEMPIERE-3646 statusupdate should only be shown on its own tab

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-3646


CarlosRuiz-globalqss (approved):
tested with process to Create Periods in Calendar>Year

PR #835, opened 2021-08-15
IDEMPIERE-4915 Improvements to Preference Popup

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4915


CarlosRuiz-globalqss (approved):
Hi @hengsin - I tested this successfully.

I think it would be better to move the line 58
`private static final int MY_PROFILE_MENU_ID = 53246;`
to declare that in the class org.compiere.model.SystemIDs

Please do the change if you consider it correct and push this ahead

PR #837, opened 2021-08-16
IDEMPIERE-4880 Replace iText PDF with OpenPDF

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4880


CarlosRuiz-globalqss (approved):
Basic tests generating PDF on Sales Order and diverse reports

PR #838, opened 2021-08-17
IDEMPIERE-4250 Implement optimistic locking support

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4250


CarlosRuiz-globalqss:
Hi @hengsin - I tested this successfully with the test case.

Side note: I was not able to use it in the UI - the UI has its own optimist locking implemented checking for Updated in GridTable.hasChanged - when I disabled that check it was strange as it seems the PO is read from the database just before being saved, I mean, the record was saved even when it was changed in another window, because the record was read before saving, when the same column is updated there is another mechanism to check for the old value that worked in that case.

OK, about the code, I didn't do tests, but I think the line 2903 in PO is wrong, or can lead to problems, if the value in database is null that must be caught by the line 2880, if it arrives to line 2903 that would mean the String has an empty value (not null).

My suggestion (not tested) would be to apply this patch
[empty1.txt](https://github.com/idempiere/idempiere/files/7014061/empty1.txt) checking for empty string

Or maybe easier just to drop the special code from String as in patch 
[empty2.txt](https://github.com/idempiere/idempiere/files/7014066/empty2.txt)

hengsin:
Hi @CarlosRuiz-globalqss , yes, atm, this is only for process and services (for e.g, web services) use only. Will check the patch attached.

hengsin:
Hi @CarlosRuiz-globalqss ,

1. Empty String and Null is not the same.
2. That's actually base on  the same approach from doUpdate (around line 2690). If it is not good here, it shouldn't be good there either.

Regards,
Low

PR #841, opened 2021-08-18
IDEMPIERE-4917:Inject Context Variables from Menu to Info Window

HideakiHagiwara:
IDEMPIERE-4917

I added Info window to menu and tested below case.
I could perse Predefined Context Variables at Info Window.
*Sql WHERE field of info window.
*Display Logic field of info window column.
*Default Logic field of info window column.
*Dynamic Validation of info window column.
*Fields of process parameter that assigned info window.

I added a new method to IInfoFactory,so someone who developed a plugin of info window needs to implement the method.
And I confirmed to impact of source code change.
*I could open info window and use at field of window.
*I could open info window and use at field of find window.
*I could open info window and use at field of process parameter.
*I could open info window and use at Views of gadget.

This change is normal InfoWindow class only, 
Below special Infowindow classes can not perse Predefined Context Variables yet.
*InfoBPartnerWindow
*InfoProductWindow
*InfoInvoiceWindow
*InfoOrderWindow
*InfoAssetWindow
*InfoInOutWindow
*InfoPaymentWindow
*InfoCashLinePanel
*InfoAssignmentWindow

Because I think that these special Info windows does not open from menu usually.
I can't judgement if I should change up to these classes.

best regards.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IInfoFactory.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IInfoFactory.java
@@ -34,4 +34,6 @@ public interface IInfoFactory {
             boolean multiSelection, String whereClause, int AD_InfoWindow_ID);
 	
 	public InfoWindow create (int AD_InfoWindow_ID); 
+	
+	public InfoWindow create (int AD_InfoWindow_ID, String predefinedContextVariables); 
 }
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
@@ -111,10 +111,9 @@ public abstract class TabbedDesktop extends AbstractDesktop {
 	 */
 	@Override
 	public void openInfo(int infoId) {
-		InfoPanel infoPanel = InfoManager.create(infoId);
+		InfoPanel infoPanel = InfoManager.create(infoId, getPredefinedContextVariables());
 		
 		if (infoPanel != null) {
-			Env.setPredefinedVariables(Env.getCtx(), infoPanel.getWindowNo(), getPredefinedContextVariables());
 			DesktopTabpanel tabPanel = new DesktopTabpanel();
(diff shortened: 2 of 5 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IInfoFactory.java
```diff
@@ -34,4 +34,6 @@ public InfoPanel create (Lookup lookup, GridField field,
             boolean multiSelection, String whereClause, int AD_InfoWindow_ID);
 	
 	public InfoWindow create (int AD_InfoWindow_ID); 
+	
+	public InfoWindow create (int AD_InfoWindow_ID, String predefinedContextVariables);
```
hengsin, line 38: please use default method to maintain backward compatibility.

hengsin:
Hi @HideakiHagiwara ,

All info window support use from field and menu so we should change those special info window class as well.

To make thing more consistent, it would be nice to add the same inject context variable field to ad_column, ad_field, etc when reference/displaytype is search.

HideakiHagiwara:
Hi, @hengsin

Thank you for your reply.

Yes, I think backward compatibility is very important too.
But I am sorry, I did not know how to implement except adding new create method definition to IInfoFactory class .

Please tell me If you have another idea of implement.

After doing so, I would like to modify the other special info window classes.
:-D

best regards.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/IInfoFactory.java
hengsin: public default InfoWindow create (int AD_InfoWindow_ID, String predefinedContextVariables) {
       return create (AD_InfoWindow_ID);
}

HideakiHagiwara: Hi, @hengsin

Thank you very much!

I'm very ashamed, I didn't know How to use "default" at interface class. 

It became very study.

I will try to rewrite and modify special info windows.

Thanks again!
:-D

HideakiHagiwara:
I forgot to create a branch. 
So,Close once. 
I will create the pull request again later.

PR #842, opened 2021-08-18
IDEMPIERE-4918 Cannot use simple auth after used OAuth (FHCA-2014)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4918


d-ruiz (approved):
Looks good

PR #843, opened 2021-08-18
IDEMPIERE-4917:Inject Context Variables from Menu to Info Window

HideakiHagiwara:
This improvement is applied normal Info Window and special Info Windows below.
*InfoBPartnerWindow
*InfoProductWindow
*InfoInvoiceWindow
*InfoOrderWindow
*InfoAssetWindow
*InfoInOutWindow
*InfoPaymentWindow
*InfoAssignmentWindow

I added above Info Windows to menu and tested below case.
I could perse Predefined Context Variables at Info Windows.
*Sql WHERE field of info window.
*Display Logic field of info window column.
*Default Logic field of info window column.
*Dynamic Validation of info window column.
*Fields of process parameter that assigned info window.

I added a new method to IInfoFactory as Default.
And I confirmed there are not bad impact of source code change.
*I could open info window and use at search field of window.
*I could open info window and use at search field of find window.
*I could open info window and use at search & Multiple Selection Search of process parameter.
*I could open info window and use at Views of gadget.

I seem there are not bad impact of source code change.

best regards.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
@@ -111,10 +111,9 @@ public abstract class TabbedDesktop extends AbstractDesktop {
 	 */
 	@Override
 	public void openInfo(int infoId) {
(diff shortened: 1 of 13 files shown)


hengsin (approved):
looks good.

PR #844, opened 2021-08-18
IDEMPIERE-4911 Silent fail when translated column is shorter than original column (FHCA-2888)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4911

Changes:
* Saving a value that doesn't fit in the translation now fails with a proper user understandable message
* Synchronize Doc Translation now fails with a user understandable message in case it fails because a value doesn't fit
* When updating a column on a translation table, check for the size of the original column and update the size accordingly
* When updating a translated column, show warnings to user about creating the table, or creating the column, or updating the column size

### DIFF
--- a/migration/i8.2/oracle/202108182002_IDEMPIERE-4911.sql
+++ b/migration/i8.2/oracle/202108182002_IDEMPIERE-4911.sql
@@ -7,7 +7,7 @@ INSERT INTO AD_Message (MsgType,MsgText,AD_Client_ID,AD_Org_ID,IsActive,Created,
 ;
 
 -- Aug 18, 2021, 8:13:56 PM CEST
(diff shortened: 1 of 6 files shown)


d-ruiz:
Tested successfully in PostgreSQL, can't test with Oracle.

Just one small suggestion. 

When you try to create a record where the name is longer than the TRL column's name, for example, the user sees this message:
"_Error: Error synchronizing translation, string too long - There is a mismatch on the size of a translated column, the translated column is shorter than the original column_"

This error is shown to the end-user, who probably won't understand what is happening, and won't have any idea how to solve the issue. That sounds more like an error to write in the log, with the name of the table failing maybe.

IMHO, a friendlier message could be shown to the end-user, something like: 
"Cannot save the record. The translation value is longer than allowed, please select a shorter name or contact the system administrator."

It's just a suggestion, if you think this is well enough, please go ahead and push the changes.

Best regards,
Diego Ruiz

CarlosRuiz-globalqss:
Ready @d-ruiz - I improved the redaction of the message partially as you suggested.

d-ruiz (approved):
Looks good

PR #846, opened 2021-08-20
IDEMPIERE-4723 - Clean up code -> remaining org.adempiere.base/src/or…

d-ruiz:
…g/compiere/model/ classes


hengsin (approved):
looks good

PR #852, opened 2021-08-30
Idempiere 1575

lHeidbreder:
An addendum to ticket 1575 with regards to ticket 4724 concerning import loader formats. To accomodate the use of multiple callouts, the size of the callout field has been increased to 255, which should be sufficient for most users.


CarlosRuiz-globalqss (changes requested):
Hi @mschnbeck - the oracle script has TO_TIMESTAMP where it must be TO_DATE, and both scripts miss the line to register_migration_script

Please add this patch that solves both problems:
[pr852_PR.txt](https://github.com/idempiere/idempiere/files/7077524/pr852_PR.txt)

Regards,

Carlos Ruiz

lHeidbreder:
> 
> 
> Hi @mschnbeck - the oracle script has TO_TIMESTAMP where it must be TO_DATE, and both scripts miss the line to register_migration_script
> 
> Please add this patch that solves both problems:
> [pr852_PR.txt](https://github.com/idempiere/idempiere/files/7077524/pr852_PR.txt)
> 
> Regards,
> 
> Carlos Ruiz

Hi @CarlosRuiz-globalqss,
in fear of something being broken, I went to regenerate the scripts and triple checked that everything seemed fine this time. I also took the opportunity to increase the size even further as to safely allow for three or four callouts.

I hope everything is in order now.
Yours,
Lukas Heidbreder

CarlosRuiz-globalqss (approved):

PR #855, opened 2021-08-30
IDEMPIERE-4911 Silent fail when translated column is shorter than original column (FHCA-2888)

CarlosRuiz-globalqss:
fix issue with 2Pack

https://idempiere.atlassian.net/browse/IDEMPIERE-4911


d-ruiz (approved):
Looks good

PR #858, opened 2021-08-31
IDEMPIERE-4937 Refreshing the GridTab after moving a node in the Tree

ArmenVH:
Immediately displays the updated data on the GridTab if changes were made to the database when moving a node in the Tree.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ADTreeOnDropListener.java
@@ -14,6 +14,9 @@ package org.adempiere.webui.component;
 
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


hengsin:
Hi @ArmenVH ,

This is working fine but unfortunately introduce a noticeable delay even when there's no event listener that's making any changes after the move.

For better UI experience, I've created https://github.com/idempiere/idempiere/pull/860. The main difference (beside some UI stuff) is the code doesn't refresh the grid tab unless it is told by the Event Listener(s) that some additional db changes have been made. Your event listener should use the env context to flag that - Env.setContext(Env.getCtx(), trx.getEventChangesIndicatorKey(), Boolean.TRUE).

Please test https://github.com/idempiere/idempiere/pull/860 and approved it if it is working good for your use cases, thanks.

Regards,
Low

hengsin:
replace by https://github.com/idempiere/idempiere/pull/860

PR #859, opened 2021-09-01
Idempiere 4845

lHeidbreder:
These changes allow the production model to add attribute set instances manually if desired. 
Noteworthy: the movement quantity to be added on the production's components needs to be negative ( just as it would be if it were generated automatically), otherwise it might result in rather strange consequences.

I also added an 'isautogenerated' column to M_ProductionLineMA which, for now, only serves as documentation.


migration/i8.2/oracle/202106291050_IDEMPIERE-4845.sql
```diff
@@ -0,0 +1,17 @@
...
+;
+-- AMEND-IDEMPIERE-4845
+-- 28.06.2021, 08:43:38 MESZ
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Help,AD_Table_ID,ColumnName,DefaultValue,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,FKConstraintType,IsHtml) VALUES (214520,0,'Auto Generated','Record is Auto Generated by System.',765,'IsAutoGenerated','N',1,'N','N','N','N','N',0,'N',20,0,0,'Y',TO_DATE('2021-06-28 08:43:33','YYYY-MM-DD HH24:MI:SS'),100,TO_DATE('2021-06-28 08:43:33','YYYY-MM-DD HH24:MI:SS'),100,202701,'N','N','D','N','N','N','Y','442e9fd0-9587-4ea3-9f77-2126731561bf','Y',0,'N','N','N','N')
+;
+
+-- 28.06.2021, 08:43:45 MESZ
+ALTER TABLE M_ProductionLineMA ADD IsAutoGenerated CHAR(1) DEFAULT 'N' CHECK (IsAutoGenerated IN ('Y','N'))
```
hengsin, line 14: this default is wrong for existing data, should be Y instead.

migration/i8.2/postgresql/202106291050_IDEMPIERE-4845.sql
```diff
@@ -0,0 +1,14 @@
...
+;
+-- AMEND-IDEMPIERE-4845
+-- 28.06.2021, 08:43:38 MESZ
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Help,AD_Table_ID,ColumnName,DefaultValue,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,FKConstraintType,IsHtml) VALUES (214520,0,'Auto Generated','Record is Auto Generated by System.',765,'IsAutoGenerated','N',1,'N','N','N','N','N',0,'N',20,0,0,'Y',TO_TIMESTAMP('2021-06-28 08:43:33','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2021-06-28 08:43:33','YYYY-MM-DD HH24:MI:SS'),100,202701,'N','N','D','N','N','N','Y','442e9fd0-9587-4ea3-9f77-2126731561bf','Y',0,'N','N','N','N')
+;
+
+-- 28.06.2021, 08:43:45 MESZ
+ALTER TABLE M_ProductionLineMA ADD COLUMN IsAutoGenerated CHAR(1) DEFAULT 'N' CHECK (IsAutoGenerated IN ('Y','N'))
```
hengsin, line 11: wrong default, same as above

org.adempiere.base/src/org/compiere/model/MProductionLine.java
```diff
@@ -1,457 +1,519 @@
...
+			DB.close(rs, pstmt);
+			rs = null;
+			pstmt = null;
+		}
+		
+		MProductionLineMA[] retValue = new MProductionLineMA[list.size()];
+		list.toArray(retValue);
+		return retValue;
+	}
+}
```
hengsin, line 519: I'm sure you didn't change the whole file. We can't review when changes/PR is not submitted as a proper diff.

hengsin (changes requested):

migration/i8.2/oracle/202106291050_IDEMPIERE-4845.sql
lHeidbreder: While that is true, there is no mechanical difference when creating attribute set instances manually as opposed to automatically, so setting the default to 'Y' would make it impossible to mark them as manually created at all. The option I see to mark all old entries as automatically generated would be to add
`UPDATE M_ProductionLineMA SET IsAutoGenerated = 'Y';`
Since this script is only run once, that should solve the issue.

CarlosRuiz-globalqss: > UPDATE M_ProductionLineMA SET IsAutoGenerated = 'Y';
>  Since this script is only run once, that should solve the issue.

I think is safer to change the default to 'Y' - is better not to assume the script won't be re-run.

hengsin: You will break existing integration and customization code that writes directly to m_productionlinema.

CarlosRuiz-globalqss: Hi @hengsin - don't know if the comment is for me or for @lHeidbreder 

I have not analyzed the code to see backward compatibility.  So, if you think the default must be 'N' is OK, but in such case I think to be on the safe side that must be done in three steps:

1 - add the column with default null and then:
 UPDATE M_ProductionLineMA SET IsAutoGenerated = 'Y' WHERE IsAutoGenerated IS NULL;
(which is the same as adding the column with default='Y')

2 - and then change the default to 'N' for the new records

My point is, it must not be assumed that people will not run the script again, there must be a where clause to protect against that case.

hengsin: Hi @CarlosRuiz-globalqss ,

My comments above is for @lHeidbreder. I'm the first to raise above that the default should be 'Y' instead of 'N'.

Regards,
Low

hengsin:
Hi @lHeidbreder,

The MProductionLine changes is missing. Please always check the content at the "Files changed" tab to verify the pull request is as what you wants it to be.

Regards,
Low

PR #860, opened 2021-09-03
IDEMPIERE-4937 Refreshing the GridTab after moving a node in the Tree

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4937


ArmenVH (approved):
Tested - it works perfectly.
Thanks.
Armen

ArmenVH (commented):
In my humble opinion your last commit **[Use Trx object instead of context for event changes indicator**] is more preferable.
Thanks
Armen

PR #861, opened 2021-09-03
IDEMPIERE-4943 Create a Material Receipt from Invoice with UOM conversion is leaving corrupt data

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4943


hengsin (approved):
trivial fix, good to go.

d-ruiz (commented):
Tested with material receipts with UOM conversion and with default UOM. Both cases are working fine

PR #862, opened 2021-09-03
IDEMPIERE-2853: Fixing Tab Editor process do not shows

dpansheriya:


dpansheriya:
This patch contain fix for regression issue due to dropping isSort field.

hengsin (changes requested):
I believe we should add back the issorttab grid field (it can be hidden), this is potentially breaks lots of stuff out there.

CarlosRuiz-globalqss:
Ah yes - this looks bad, I see references in core code about IsSortTab for example in:
GenericZoomProvider
lots of calls to GridTabVO.isSortTab() which has not been moved to the new approach.

What @hengsin suggest is better - restore the old field in a read-only mode and fill it properly via beforeSave.

I'm closing this one - in favour of the suggested approach.

CarlosRuiz-globalqss:
replaced with https://github.com/idempiere/idempiere/pull/885

PR #863, opened 2021-09-03
IDEMPIERE-4924 Performance issue with indirect context

hengsin:
- added cache
https://idempiere.atlassian.net/browse/IDEMPIERE-4924


CarlosRuiz-globalqss:
Hi @hengsin - I tested this and it works when the changes are done using the model, but if the change of the foreign record is done using direct SQL then the query is not checked again until a cache reset is executed.

Would it be better to have a short life cache for this case?

I would prefer a cache that last just seconds (f.e. 5 seconds), but the cache just admit minutes as the parameter and the min to set is 1 minute, so, maybe this patch is worthy initially?
[oneminutecache.txt](https://github.com/idempiere/idempiere/files/7161257/oneminutecache.txt)

CarlosRuiz-globalqss (approved):

PR #865, opened 2021-09-04
IDEMPIERE-4945 Selected row is changed after table sort in Swing UI

dram:
At the end of table sorting, trigger data status change event before UI
event, so that GridTab data is changed before UI refresh, which is
needed for Swing UI.

I have tested the patch in Swing and ZK.

Not sure why ZK UI do not have this same issue.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -1013,10 +1013,10 @@ public class GridTable extends AbstractTableModel
 		
 		if (!isSameSortEntries)
 		{
-			//	update UI
-			fireTableDataChanged();
 			//  Info detected by MTab.dataStatusChanged and current row set to 0
 			fireDataStatusIEvent(SORTED_DSE_EVENT, "#" + m_sort.size());
+			//	update UI
+			fireTableDataChanged();
 		}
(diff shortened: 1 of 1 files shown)


hengsin:
Hi @dram ,

I've not tested this but the code is not an exact reflection of what's describe above.

GridTab doesn't holds any data, it get its data from the internal grid table object and when fireTableDataChanged() is invoke, the data at GridTable have been sorted/changed (hence the container of it, the GridTab).

Regards,
Low

dram:
Hi @hengsin ,

Sorry for the bad description. What I mean for "GridTab data" is its `m_currentRow` property. That property is updated after `fireDataStatusIEvent` call by GridTable, the relevant code is:

https://github.com/idempiere/idempiere/blob/ae5a665dd8c9cb6a7ecf911cb35690a40542f2a3/org.adempiere.base/src/org/compiere/model/GridTab.java#L2354-L2361

CarlosRuiz-globalqss:
Hi @dram 

> Not sure why ZK UI do not have this same issue.

I think we had this issue in zk before (must be a JIRA ticket about that).

I don't remember exactly how was the solution - I'll try to search for the ticket later.

hengsin (approved):
looks good

PR #871, opened 2021-09-10
IDEMPIERE-4942 Performance Measurement - NPE by opening the Performan…

hengsin:
…ce Detail Window

add null check

https://idempiere.atlassian.net/browse/IDEMPIERE-4942


CarlosRuiz-globalqss (approved):
trivial fix

PR #877, opened 2021-09-12

saulojg:
https://idempiere.atlassian.net/browse/IDEMPIERE-4842


org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		if(candidate!=null)
+			// inject association into DefaultModelFactory's cache
+			s_classCache.put(tableName, candidate);
+
+		return candidate;
+	}
+
+}
```
hengsin, line 65: I don't see a need to extend DefaultModelFactory here. Since we are not removing DefaultModelFactory, we don't have to do an extend here - if AnnotationBasedModelFactory return null, the DefaultModelFactory will just do its part as it is now. Also, if we drop DefaultModelFactory in future, this is just unnecessary legacy baggage.

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
```diff
@@ -0,0 +1,65 @@
...
+		// scan annotations
+		BundleWiring wiring = usingBundle.adapt(BundleWiring.class);
+		for(Class<?> clazz : ClassIndex.getAnnotated(Model.class, wiring.getClassLoader()))
+		{
+			Model ma = clazz.getAnnotation(Model.class);
+			if(ma.table().equalsIgnoreCase(tableName)) 
+			{
+				candidate = clazz;
+				// M* classes are preferred over X* classes
+				if(!ma.intermediate())
```
hengsin, line 53: Don't fancy we take the intermediate class here. With this, if we don't manually annotate each and every existing M* classes, the X* will bed used instead. As I've posted below, I think we need to have facility to generate or modify M* classes instead.

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
```diff
@@ -168,6 +168,7 @@ private String createHeader (int AD_Table_ID, StringBuilder sb, StringBuilder ma
 		start.append("/** Generated Model for ").append(tableName).append(NL)
 			 .append(" *  @author iDempiere (generated) ").append(NL)
 			 .append(" *  @version ").append(Adempiere.MAIN_VERSION).append(" - $Id$ */").append(NL)
+			 .append("@org.adempiere.base.Model(table=\"").append(tableName).append("\", intermediate=true)").append(NL)
 			 .append("public class ").append(className)
 			 	.append(" extends PO")
 			 	.append(" implements I_").append(tableName)
```
hengsin, line 174: I think we need to have the facility to generate the M class here (or add annotation to existing M class). Without this, we are leave with generated class that can't be used by DefaultModelFactory or accidentally take precedence over existing M* class (if we take the intermediate class).

org.adempiere.base/src/org/compiere/model/MProduct.java
```diff
@@ -48,6 +48,7 @@
  * @author Mark Ostermann (mark_o), metas consult GmbH
  * 			<li>BF [ 2814628 ] Wrong evaluation of Product inactive in beforeSave()
  */
+@org.adempiere.base.Model(table="M_Product")
 public class MProduct extends X_M_Product implements ImmutablePOSupport
 {
 	/**
```
hengsin, line 54: don't need an example here unless it is an modification for all M* classes

org.adempiere.base/src/org/compiere/model/X_M_Product.java
```diff
@@ -27,6 +27,7 @@
 /** Generated Model for M_Product
  *  @author iDempiere (generated) 
  *  @version Release 8.2 - $Id$ */
+@org.adempiere.base.Model(table="M_Product", intermediate = true)
```
hengsin, line 30: if needed, we should run model generator for all existing classes, not just 1 single class here. I understand that this served as an example but this is not appropriate for a PR.

hengsin (changes requested):

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
hengsin: Just come across this http://www.atteo.org/static/classindex/apidocs/org/atteo/classindex/ClassIndex.html#getSubclasses-java.lang.Class- and I'm wondering whether we can use this to get rid of the need to use intermediate and the need to annotate M* classes ?

org.adempiere.base/OSGI-INF/org.adempiere.base.AnnotationBasedModelFactory.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.3.0" activate="activate" name="org.adempiere.base.AnnotationBasedModelFactory">
+   <service scope="bundle">
+      <provide interface="org.adempiere.base.IModelFactory"/>
+   </service>
+   <implementation class="org.adempiere.base.AnnotationBasedModelFactory"/>
+</scr:component>
```
hengsin, line 7: You have duplicate component definition here. I believe one is manually created and the other auto generated from "@Component" annotation.

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
saulojg: The main reason behind extending ```DefaultModelFactory``` is to avoid rewriting ```org.adempiere.base.DefaultModelFactory.getPO(String, int, String)```. It might not be ideal to have every other model factory implement ```PO``` instantiation via reflection on it's own, duplicating the same code each time (e.g.: https://github.com/ingeint/idempiere-plugin-scaffold/blob/master/template/plugin.symbolic.name/src/plugin.root/base/CustomModelFactory.java#L78 ).
If there's an intention to deprecate ```DefaultModelFactory``` I'd suggest to move this kind of model instantiation somewhere else, so that it can be reused by model factories (especially coming from plugins). Maybe we could also cache instances of these standard constructors and speed up model class instantiation a little bit.

org.adempiere.base/OSGI-INF/org.adempiere.base.AnnotationBasedModelFactory.xml
saulojg: Good catch. Fixed.

org.adempiere.base/src/org/compiere/model/X_M_Product.java
saulojg: We didn't worry much about this, as accepting this PR might imply that all X classes might get annotated. Though it isn't required, we'd like that core model classes get detected and loaded by using this method instead of  the current ```Class.forName``` based approach. However, we've reverted this commit for now.

org.adempiere.base/src/org/compiere/model/MProduct.java
saulojg: Please see below.

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
```diff
@@ -168,6 +169,8 @@ private String createHeader (int AD_Table_ID, StringBuilder sb, StringBuilder ma
 		start.append("/** Generated Model for ").append(tableName).append(NL)
 			 .append(" *  @author iDempiere (generated) ").append(NL)
 			 .append(" *  @version ").append(Adempiere.MAIN_VERSION).append(" - $Id$ */").append(NL)
+			 .append("@IndexSubclasses").append(NL)
+			 .append("@org.adempiere.base.Model(table=\"").append(tableName).append("\", intermediate=true)").append(NL)
```
hengsin, line 173: with the new getSubclasses approach, isn't the intermediate attribute not needed anymore ?

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
saulojg: Excellent idea, implemented at 9d9a22eb749f5075bdfebd1aecd20ea4158daf4b.

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
hengsin: alright, that's a good point. I guess we can do the refactoring and performance optimization later (the straightforward approach is to extract that to an abstractmodelfactory class and have the class cache as instance variable instead. The class cache doesn't have to be static since osgi service is already a singleton).

org.adempiere.base/src/org/adempiere/base/Model.java
```diff
@@ -0,0 +1,24 @@
+package org.adempiere.base;
```
hengsin, line 1: please add license header

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
```diff
@@ -0,0 +1,68 @@
+package org.adempiere.base;
```
hengsin, line 1: please add license header

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
saulojg: OK I'll try that approach on a separate PR.

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
saulojg: Sorted

org.adempiere.base/src/org/adempiere/base/Model.java
saulojg: Sorted

org.adempiere.base/src/org/adempiere/base/AnnotationBasedModelFactory.java
saulojg: Sorted

saulojg:
This branch contains changes for tickets 4842 and 4965.

hengsin (changes requested):
Hi @saulojg ,

Please merge the changes from https://github.com/hengsin/idempiere/commit/7157900ae4876b55bd34be1ab6ccb1eb1fdd17bb

Note that you have to either remove the OSGI-INF/annotationbasemodelfactory.xml file or remove the "@Component" annotation from AnnotationBasedModelFactory. Having both, you will end up having 2 component definition file for AnnotationBasedModelFactory (the manually created OSGI-INF/annotationbasemodelfactory.xml and OSGI-INF/org.adempiere.base.AnnotationBasedModelFactory.xml generated by the Eclipse PDE's DS Annotations processor).

Also, we will have to regenerate all core model classes after this get merge into core. 

Hi @CarlosRuiz-globalqss , do we have a wiki page for the steps to regenerate all core model classes ?

Regards,
Low

saulojg:
> Hi @saulojg ,
> 
> Please merge the changes from [hengsin@7157900](https://github.com/hengsin/idempiere/commit/7157900ae4876b55bd34be1ab6ccb1eb1fdd17bb)

Done
> 
> Note that you have to either remove the OSGI-INF/annotationbasemodelfactory.xml file or remove the "@component" annotation from AnnotationBasedModelFactory. Having both, you will end up having 2 component definition file for AnnotationBasedModelFactory (the manually created OSGI-INF/annotationbasemodelfactory.xml and OSGI-INF/org.adempiere.base.AnnotationBasedModelFactory.xml generated by the Eclipse PDE's DS Annotations processor).

Apparently this XML file got renamed (according to git) on your commit, so I think we should be good. I've read your comments before and was (incorrectly) sure the file got removed, thanks for taking a second look. 
> 
> Also, we will have to regenerate all core model classes after this get merge into core.

Yes, either that or come up with a quick shell script.

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss , do we have a wiki page for the steps to regenerate all core model classes ?

Hi @hengsin - this is what I have in my release notes:
________________________
generate complete model for new version (core and e-evolution) (be careful not to include customizations):
Run model generator with the following parameters:

[1]
"/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
"org.compiere.model"
"D,EE05"
"%"

[2]
"/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
"org.compiere.model"
"D"
"AD_UserMail"
"D"

[3]
"/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
"org.eevolution.model"
"EE01,EE04,EE02"
"%"

[4]
"/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
"org.compiere.model"
"D"
"'AD_AllUsers_V','AD_AllClients_V'"
________________________

saulojg:
Thank you Carlos I'll give it a try

saulojg:
Squashing idea, not fully tested:

```
# IDEMPIERE-4842 - Easier model registration (before IDEMPIERE-4965)
git cherry-pick -n d31eff507b6f02d2faeac70b867fd8d216dc93de b83531507a7b1bb9f63e7725d78cf0a43333ef61 1788165faefdaaa6287c261ba486c92561bcc78c a19cd5acd28155e90dabc9712747ed9fc6d7a695 b1e604f229a3578d927b199bfed629a84bc922e2 9d9a22eb749f5075bdfebd1aecd20ea4158daf4b 6880da3bdb4a888654c7c4da40379a338193add1 09aab97d37a4e4fadfbfa9c8b84aa2e93fb839f6
git commit

# IDEMPIERE-4965 Code reutilization for PO instantiation
git cherry-pick -n 058e93aae0ef791a91a554df74b477c3f00ba335 4aead673c53968c576909d77f57270a29753b0c0
git checkout --theirs org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
git add org.adempiere.base/src/org/adempiere/base/DefaultModelFactory.java
git cherry-pick --continue

# IDEMPIERE-4842 After IDEMPIERE-4965
git cherry-pick -n 633a5ad5ef389ca2242ebc97a6641a6cc39293a4 b254efb9c9a3688085624e887def20e2e8299b96 a1a14ffe211deff109414c00c35bb8c195567ed5
git commit

# Update model classes
git cherry-pick d0e9f3cf05f704dc048548fc91da30f9d1d3a95a

```

hengsin:
> > Hi @CarlosRuiz-globalqss , do we have a wiki page for the steps to regenerate all core model classes ?
> 
> Hi @hengsin - this is what I have in my release notes:
> 
> generate complete model for new version (core and e-evolution) (be careful not to include customizations):
> Run model generator with the following parameters:
> 
> [1]
> "/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
> "org.compiere.model"
> "D,EE05"
> "%"
> 
> [2]
> "/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
> "org.compiere.model"
> "D"
> "AD_UserMail"
> "D"
> 
> [3]
> "/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
> "org.eevolution.model"
> "EE01,EE04,EE02"
> "%"
> 
> [4]
> "/home/carlos/gitIdempiere/localdev/org.adempiere.base/src"
> "org.compiere.model"
> "D"
> "'AD_AllUsers_V','AD_AllClients_V'"

Thanks, added to https://wiki.idempiere.org/en/Generate_Model_For_New_Version

hengsin (approved):
tested changes prior to generation of model classes.

PR #878, opened 2021-09-13
IDEMPIERE-4961 Minor performance enhancement for AD Window

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-4961


org.adempiere.base/src/org/compiere/model/MUserQuery.java
```diff
@@ -300,7 +300,7 @@ public MUserQuery(Properties ctx, ResultSet rs, String trxName)
 	protected boolean beforeSave(boolean newRecord) {
 		if (getAD_Tab_ID() > 0) {
 			if (newRecord || is_ValueChanged(COLUMNNAME_AD_Tab_ID)) {
-				MTab tab = new MTab(getCtx(), getAD_Tab_ID(), get_TrxName());
+				MTab tab = MTab.get(getAD_Tab_ID());
```
CarlosRuiz-globalqss, line 303: Hi @hengsin - I have not tested, but usually reading cache in beforeSave or afterSave always bring problems for 2Pack.

In this case the test would be a 2Pack containing the tab and then the user query, when that happens, the tab is not discovered because the MTab.get is read out of the 2pack transaction.

CarlosRuiz-globalqss (approved):

PR #880, opened 2021-09-14
IDEMPIERE-4963 Performance: avoid loop in X_ class interface getters

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4963


hengsin (approved):
looks good.

PR #885, opened 2021-09-17
IDEMPIERE-2853 Panel tab as a factory - preserve AD_Tab.IsSortTab for backward compatibility

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-2853


hengsin (approved):
tested

PR #887, opened 2021-09-18

hengsin:
- Temporary disable AnnotationBasedModelFactory until getSubclasses()
doesn't work for subclass from different package is resolve.


hengsin:
push through as a temporary emergency fix.

PR #888, opened 2021-09-19

hengsin:
turn off annotation processing in Eclipse editor which causes slow editor performance


hengsin:
minor Eclipse only fix.

PR #889, opened 2021-09-19
IDEMPIERE-4962 Performance: change PostgreSQL native dialect mode def…

hengsin:
…ault to true

- minor refinement and fix Convert_PostreSQLTest

https://idempiere.atlassian.net/browse/IDEMPIERE-4962


CarlosRuiz-globalqss (approved):
basic tests passed (POS order with and without native dialect)

PR #890, opened 2021-09-20
IDEMPIERE-4972 : Translate icon is missing when using font icons

nmicoud:


CarlosRuiz-globalqss:
Hi @nmicoud - I think for translate the most known icon is f1ab
https://fontawesome.com/v5.15/icons/language?style=solid
WDYT?

nmicoud:
hi @CarlosRuiz-globalqss,
I like your suggestion, indeed, is better for illustrating "translation".
But is quite different from the original one (flag).
Anyway, this "disruption" seems acceptable so +1 for me :)

PR #891, opened 2021-09-20
IDEMPIERE-4968 Enable 2Pack to run java class or script (FHCA-3112 FHCA-3113) / Shell script

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4968

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/X_AD_Package_Exp_Detail.java
+++ b/org.adempiere.base/src/org/compiere/model/X_AD_Package_Exp_Detail.java
@@ -31,7 +31,7 @@ public class X_AD_Package_Exp_Detail extends PO implements I_AD_Package_Exp_Deta
 	/**
 	 *
 	 */
-	private static final long serialVersionUID = 20210917L;
+	private static final long serialVersionUID = 20210920L;
 
     /** Standard Constructor */
     public X_AD_Package_Exp_Detail (Properties ctx, int AD_Package_Exp_Detail_ID, String trxName)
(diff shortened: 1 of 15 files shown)


org.adempiere.base/src/org/compiere/model/X_AD_Package_Exp_Detail.java
```diff
@@ -934,6 +950,8 @@ public String getTarget_Directory ()
 	public static final String TYPE_ImportFormat = "IMP";
 	/** Info Window = IW */
 	public static final String TYPE_InfoWindow = "IW";
+	/** Java Shell Script = JSH */
+	public static final String TYPE_JavaShellScript = "JSH";
```
hengsin, line 954: This naming choice sounds a little confusing to me since it got nothings to do with "shell". I would prefer to call it just "Script" instead.

org.adempiere.base/src/org/compiere/model/X_AD_Package_Exp_Detail.java
```diff
@@ -950,6 +968,10 @@ public String getTarget_Directory ()
 	public static final String TYPE_Reference = "REF";
 	/** Role = S */
 	public static final String TYPE_Role = "S";
+	/** Shell Script = SH */
+	public static final String TYPE_ShellScript = "SH";
+	/** Code Snippet = SNI */
+	public static final String TYPE_CodeSnippet = "SNI";
```
hengsin, line 974: It seems this and TYPE_File_CodeOrOther is not use anywhere ?

CarlosRuiz-globalqss: OK, please check https://github.com/idempiere/idempiere/pull/898

org.adempiere.base/src/org/compiere/model/X_AD_Package_Exp_Detail.java
CarlosRuiz-globalqss: renamed to ScriptJSR223

hengsin (approved):
tested

PR #893, opened 2021-09-22

saulojg:
https://idempiere.atlassian.net/browse/IDEMPIERE-4842

We've replaced the [ClassIndex](https://github.com/atteo/classindex) library with [ClassGraph](https://github.com/classgraph/classgraph), since the former caused problems with Eclipse's incremental builds.

Co-authored with @hengsin


hengsin (approved):
tested

PR #895, opened 2021-09-23

hengsin:
- Needs to enable jar scanning to support felix/console installation of plugin jar.
- add rejectJars call to reduce startup delay after enable of jar scanning. keep the list short to reduce future maintenance hazard.
- remove the not supported use of acceptPackagesNonRecursive and acceptClasses together.
- change getAcceptClassesPatterns() default to X_* and M*. Withouut acceptPackages, we need to scan both X and M classes.


hengsin:
hi @saulojg ,

Please help review this, thanks.

Regards,
Low

saulojg (approved):
Changes already reviewed at Mattermost.

saulojg (approved):

PR #896, opened 2021-09-23
IDEMPIERE-4977 IDempiere Homepage using custom image path failed to l…

uthadehikaru:
…oad the image

this is a bug fix for this issue. method to encode should be encodeToString instead encode.

### DIFF
--- a/org.adempiere.server/src/main/home/org/bmlaurus/home/Prop.java
+++ b/org.adempiere.server/src/main/home/org/bmlaurus/home/Prop.java
@@ -193,7 +193,7 @@ public final class Prop implements Serializable {
 			try {
 				fis = new URL(getProperty(TEMPLATE_PATH)+getProperty(TEMPLATE_NAME)+"/styles/template.css");
 				byte data[]=read(fis);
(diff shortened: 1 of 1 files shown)


hengsin (changes requested):
According to https://stackoverflow.com/questions/5641303/base64url-in-java, the proper fix should be Base64.getUrlEncoder().withoutPadding().encodeToString(data) instead.

CarlosRuiz-globalqss (changes requested):
tested

CarlosRuiz-globalqss:
> According to https://stackoverflow.com/questions/5641303/base64url-in-java, the proper fix should be Base64.getUrlEncoder().withoutPadding().encodeToString(data) instead.

Hi @hengsin - I tested the solution from @uthadehikaru and it worked fine.

The change suggested in stackoverflow didn't work - it shows blank images.  The getUrlEncoder works just for URLs and filenames and here it is encoding a byte image.

So, I think is OK to push this as it is.

hengsin:
alright, my mistake, the patch is good.

hengsin (approved):

PR #897, opened 2021-09-23
IDEMPIERE-4978 Fix error in setting product resource type

kriegfrj:
I think this fixes things.


CarlosRuiz-globalqss (approved):
looks correct

PR #898, opened 2021-09-23
IDEMPIERE-4979 ModelGenerator class generates entries for inactive ref lists

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4979


hengsin (approved):
trivial fix, good to go

PR #900, opened 2021-09-24
IDEMPIERE-4981 Make it easier to find Process Audit record for each S…

hengsin:
…cheduler Run

https://idempiere.atlassian.net/browse/IDEMPIERE-4981


CarlosRuiz-globalqss (approved):
Hi @hengsin,

Tested and it works adding the string "AD_PInstance_ID=1000017" to the Reference field.

If you want please push it ahead as is seems trivial.

However, I think a better solution would be to add AD_SchedulerLog.AD_PInstance_ID - that way the reference will be zoomable.

Regards,

Carlos Ruiz

PR #901, opened 2021-09-27
IDEMPIERE-4982 : Methods to retrieve value from multiple process para…

nmicoud:
…meters

### DIFF
--- a/org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
+++ b/org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
@@ -371,7 +371,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter as an array of int. Validate every value is an integer and throws NumberFormatException if one of the value is not a valid integer.
-	 * @return String
+	 * @return array of int
 	 */
 	public int[] getParameterAsIntArray() {
 		return getParameterAsIntArray(getParameterAsString());
@@ -379,7 +379,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter To as an array of int. Validate every value is an integer and throws NumberFormatException if one of the value is not a valid integer.
-	 * @return String
+	 * @return array of int
 	 */
 	public int[] getParameterToAsIntArray() {
 		return getParameterAsIntArray(getParameter_ToAsString());
@@ -387,7 +387,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter as an array of int. Validate every value is an integer and throws NumberFormatException if one of the value is not a valid integer.
-	 * @return String
+	 * @return array of int
 	 */
 	private int[] getParameterAsIntArray(String param) {
 
@@ -399,7 +399,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter as an array of String.
-	 * @return String
+	 * @return array of String
 	 */
 	public String[] getParameterAsStringArray() {
 		return getParameterAsStringArray(getParameterAsString());
@@ -407,7 +407,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter To as an array of String.
-	 * @return String
+	 * @return array of String
 	 */
 	public String[] getParameterToAsStringArray() {
 		return getParameterAsStringArray(getParameter_ToAsString());
@@ -415,7 +415,7 @@ public class ProcessInfoParameter implements Serializable
 
 	/**
 	 * Return the value of the parameter as an array of String.
-	 * @return String
+	 * @return array of String
 	 */
 	private String[] getParameterAsStringArray(String param) {
 
--- a/org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
+++ b/org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
@@ -20,6 +20,8 @@ import java.io.Serializable;
 import java.math.BigDecimal;
 import java.sql.Timestamp;
 
+import org.compiere.util.DB;
+
 
 /**
  * 	Process Parameter
@@ -32,10 +34,11 @@ import java.sql.Timestamp;
  */
 public class ProcessInfoParameter implements Serializable
 {
+
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -8571973325856537109L;
+	private static final long serialVersionUID = -5396796617617359891L;
 
 	/**
 	 *  Construct Parameter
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
For me the patch is OK with two comments (asking also opinion from @hengsin please)

1 - it sounds harmless to implement the getParameter_ToAsMultiple* - but I really doubt they must be used, a range on arrays doesn't make too much sense to me in this context

2 - line 317 can throw a NumberFormatException - is it OK this way - or better to make it a controlled exception and throw a more readable message for the user?   Note that this error is not expected to happen when iDempiere is configured correctly.

WDYT?

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +285,78 @@ public void setParameterName (String ParameterName)
...
+
+		StringBuilder whereValidated = new StringBuilder();
+		for (String par : strarr) {
+			int val = Integer.valueOf(par);
+			if (whereValidated.length() > 0)
+				whereValidated.append(",");
+			whereValidated.append(val);
+		}
+
+		return whereValidated.toString();
```
hengsin, line 323: The whereValidated is actually redundant here. If every value is an integer, whereValidated must be equal to param. If one of them is not an integer, the method will throw exception and whereValidated is of no use either.

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +285,78 @@ public void setParameterName (String ParameterName)
...
+	public String getParameter_ToAsMultipleString() {
+		return getParameterMultipleString(m_Parameter_To);
+	}
+
+	/**
+	 * Return the value of the parameter as a validated String (all values between commas are surrounded by '')
+	 * @return String
+	 */
+	private String getParameterMultipleString(Object param) {
+		if (param == null || !(param instanceof String))
```
hengsin, line 347: The method works fine but the javadoc is misleading - there's no validation happens here, you just quoted it.

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +285,78 @@ public void setParameterName (String ParameterName)
...
+	 */
+	public String getParameter_ToAsMultipleInt() {
+		return getParameterMultipleInt(m_Parameter_To);
+	}
+
+	/**
+	 * Return the value of the parameter as a validated String (all values between commas must be integer)
+	 * @return String
+	 */
+	private String getParameterMultipleInt(Object param) {
```
hengsin, line 308: should just use string here and getParameterAsMultipleInt(m_Parameter) can be changed to getParameterAsMultipleInt(getParameterAsString())

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +285,78 @@ public void setParameterName (String ParameterName)
 		m_ParameterName = ParameterName;
 	}
 
+	/**
+	 * Return the value of the parameter as a validated String (all values between commas must be integer)
+	 * @return String
```
hengsin, line 290: I think this can be changed to "Return the value of the parameter as a comma separated integer string. Validate every value is an integer and throws NumberFormatException if one of the value is not a valid integer.".

Also, I think there might be instance where you would want an int[] instead of a csv string.

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +285,78 @@ public void setParameterName (String ParameterName)
...
+		StringBuilder whereValidated = new StringBuilder();
+		for (String par : strarr) {
+			if (whereValidated.length() > 0)
+				whereValidated.append(",");
+			whereValidated.append(DB.TO_STRING(par));
+		}
+
+		return whereValidated.toString();
+	}
+
```
hengsin, line 361: Similar to the multiple int above, I think there are circumstances where you would want to have string[] instead of a csv string.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: done

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: javadocs updated

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: done ; I also add a getParameter_ToAsString() method

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: done

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: done

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +300,134 @@ public void setParameterName (String ParameterName)
...
+	 */
+	public String[] getParameterToAsMultipleStringArray() {
+		return getParameterAsMultipleStringArray(getParameter_ToAsString());
+	}
+
+	/**
+	 * Return the value of the parameter as an array of String. Validate every value is an integer and throws NumberFormatException if one of the value is not a valid integer.
+	 * @return String
+	 */
+	private String[] getParameterAsMultipleStringArray(String param) {
```
hengsin, line 420: Hi @nmicoud , you have wrongly copy over the integer javadoc to the string version. For the array method, the method name can be simplified to getParameterAsStringArray and getParameterAsIntArray.

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +300,134 @@ public void setParameterName (String ParameterName)
...
+	 */
+	public String getParameterAsMultipleString() {
+		return getParameterMultipleString(getParameterAsString());
+	}
+
+	/**
+	 * Return the value of the parameter as a String with all values between commas surrounded by quotes
+	 * @return String
+	 */
+	public String getParameter_ToAsMultipleString() {
```
hengsin, line 348: I think a more direct and obvious naming is getParameterAsCSVString and getParameterAsCSVInt

hengsin (changes requested):

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: done

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: hi
done

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
```diff
@@ -282,4 +300,134 @@ public void setParameterName (String ParameterName)
...
+	public String[] getParameterToAsStringArray() {
+		return getParameterAsStringArray(getParameter_ToAsString());
+	}
+
+	/**
+	 * Return the value of the parameter as an array of String.
+	 * @return String
+	 */
+	private String[] getParameterAsStringArray(String param) {
+
```
hengsin, line 421: sorry to be picky here but there are still several method above with wrong "@return" type

hengsin (changes requested):

org.adempiere.base/src/org/compiere/process/ProcessInfoParameter.java
nmicoud: should be ok now :)

hengsin (approved):
looks good.

PR #903, opened 2021-09-27
IDEMPIERE-4983 : Fix Russian Ruble currency

nmicoud:


CarlosRuiz-globalqss:
The patch looks simple - however I would ask first Eugene about the potential impact on their accounting schema.

CarlosRuiz-globalqss (changes requested):
Hi @nmicoud

This answer 
https://idempiere.atlassian.net/browse/IDEMPIERE-4983?focusedCommentId=47538
is the scenario I think most of the times have happened.

People must be simply using the current record for Ruble even if the ISO code was wrong.

And the problem is not just for the russian implementations, but any implementation in the world recording payments/invoices in this foreign currency.

So, I think the solution is simpler, just update the ISO Code to RUB on c_currency_id = 155

IMO that's the less disruptive thing to do, some people probably already corrected that and it will be harmless.

The problem would be for people that already created a new record for RUB, but I think that scenario is the minority (if any).

nmicoud:
hi @CarlosRuiz-globalqss ,
patch updated
I also move scripts to i8.2z as it may fail.
nb: my customer has also check other currencies and he send me the results, when i have time, i'll propose another patch

CarlosRuiz-globalqss (approved):

PR #904, opened 2021-10-04
IDEMPIERE-4721 : Financial Report : exclude all adjustment periods - …

nmicoud:
…fix NPE

reported by @CarlosRuiz-globalqss ; see https://idempiere.atlassian.net/browse/IDEMPIERE-4721?focusedCommentId=47539

### DIFF
--- a/org.adempiere.base/src/org/compiere/report/FinReport.java
+++ b/org.adempiere.base/src/org/compiere/report/FinReport.java
@@ -197,7 +197,7 @@ public class FinReport extends SvrProcess
 		sb.append(" - ").append(m_report);
 
 		// Exclude adjustment period(s) ?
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @nmicoud - I haven't tested but I think the solution is wrong.

When period is empty the system calculates the default period as the current period (based on today's date).
So, if you run the report on december and you have adjustment period on dec-31 - the patch will ignore that, unless you fill the period specifically.

I think (but haven't tested) a better solution is to call the setPeriods() just before the validation, this is, moving setPeriods() from line 218 to line 198 (something like that).

It is expected that setPeriods set the default value for p_C_Period_ID

Can you please check and test that scenario?

Regards,

Carlos Ruiz

PR #909, opened 2021-10-05
IDEMPIERE-1386 - email from attachments dialog

muriloht:


CarlosRuiz-globalqss (changes requested):
Hi @muriloht - can you please add the following patch to this pull request and test again:
[pr909_PR.txt](https://github.com/idempiere/idempiere/files/7295665/pr909_PR.txt)

The issue worked fine for me in desktop.

But in mobile the toolbar is now taking the whole screen and I'm not able to select a different attachment, the suggested patch changes the Hbox to Toolbar and makes it usable in mobile.

Regards,

Carlos Ruiz

PR #927, opened 2021-10-15
IDEMPIERE-4988 Fix NPE on change Role

ocurieles:
https://idempiere.atlassian.net/browse/IDEMPIERE-4988


hengsin:
Hi @CarlosRuiz-globalqss , this looks trivial but the PR is against 8.2 branch instead of master. What should be the flow here ?

PR #931, opened 2021-10-16
IDEMPIERE-4770 - fix advanced find and/or where clause

muriloht:
when trying to use the advanced find, the first condition is always returning an error :

image


hengsin:
haven't check the code but noted one interesting finding here - this is reproduce-able at test.idempiere.org at sales order window but not at product window.

tony-tspc:
I created [IDEMPIERE-4999](https://idempiere.atlassian.net/browse/IDEMPIERE-4999) a few days ago regarding this issue. The problem occurs on any window that has an extended where clause. It happens on the Sales Order window because it has an extended where of IsSOTrx = 'Y', but there is no extended where on the Product window. Also if you do an advanced query on a sub-tab, then the problem also occurs, because the extended where contains the parent record. 
Currently the Advanced Find query is added to the extended where clause without adding an "and", so the query fails. If there is no extended where clause it succeeds.

PR #935, opened 2021-10-19
IDEMPIERE-5005 - Chat window broken

saulojg:
- Fix event processing on chat window
- Removed confirm panel as it servers no other purpose than closing the window.


hengsin:
Hi @saulojg , why 2 pull request for IDEMPIERE-5005 ?

PR #170, opened 2020-07-17
IDEMPIERE-4364 Advanced Search - Allow select column from window tabs

igorpojzl:
https://idempiere.atlassian.net/browse/IDEMPIERE-4364

PR #959, opened 2021-10-31
IDEMPIERE-4472 Window Advanced Search - Allow select column from wind…

hengsin:
…ow tabs - 2pack compatibility

Based on pull request #170 from igorpojzl.
Difference from #170:
- use AD_Tab_UU instead of AD_Tab_ID
- use Combobox for table list

https://idempiere.atlassian.net/browse/IDEMPIERE-4364

### DIFF
--- a/migration/i8.2z/oracle/202111011110_IDEMPIERE-4364.sql
+++ b/migration/i8.2z/oracle/202111011110_IDEMPIERE-4364.sql
@@ -1,0 +1,15 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-4364 Advanced Search - Allow select column from window tabs
+-- Nov 1, 2021, 11:09:11 AM CET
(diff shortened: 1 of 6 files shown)


CarlosRuiz-globalqss:
Hi @hengsin 

I tested this successfully.

In the ticket it was mentioned to increase the size of AD_UserQuery.Code to 4000 char, can you please consider adding the following patch that adds that migration script:
[PR959_ms.txt](https://github.com/idempiere/idempiere/files/7452371/PR959_ms.txt)

I tested changing to 32767, but oracle doesn't support that lenght _by default_, it requires some initialization or migration stuff to work:
https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/MAX_STRING_SIZE.html#GUID-D424D23B-0933-425F-BC69-9C0E6724693C

We can use 32767 and declare that oracle databases MUST have that configuration, or keep the limit in 4000 char in core.
Not sure what is better.

CarlosRuiz-globalqss (approved):

PR #960, opened 2021-11-01
IDEMPIERE-5022 fix BOM Viewer data pane sometimes showing incorrect i…

tony-tspc:
Under certain conditions the BOM Viewer right-hand data pane is showing incorrect information.
As explained in the bug report, the issue only occurs when you enter data in a particular sequence. 
I have run a number of tests and the fix resolves the issue.


hengsin (approved):
looks fine.

PR #963, opened 2021-11-03
IDEMPIERE-5025 Easier usage of process parameters

saulojg:
Created `@Parameter` annotation in order to tag process class attributes as parameters and have them to have their values assigned automatically without extra code.


hengsin (approved):
looks good

PR #964, opened 2021-11-04
IDEMPIERE-5026 Create from Invoice showing already invoiced RTV docs

tony-tspc:
Create from Invoice is showing already invoiced Return to Vendor docs with double the original quantity.
The issue is related to the changing of the sign of the matched quantity in IDEMPIERE-4774.


tony-tspc:
Will close this for now. Not a complete solution and there are other issues to consider relating to IDEMPIERE-4774.

PR #965, opened 2021-11-04
IDEMPIERE-5027 Packing out lost translation of sub menu

Fikar:
Copy code snippet handle common translation from MenuElementHandler.createApplication.


CarlosRuiz-globalqss (approved):
tested

PR #968, opened 2021-11-05
IDEMPIERE-5029 User can make Shipment/Receipt Line Product different …

hengsin:
…from Order Line Product

https://idempiere.atlassian.net/browse/IDEMPIERE-5029


CarlosRuiz-globalqss:
Please check comments and suggested patch in ticket

CarlosRuiz-globalqss:
Hi @hengsin - I noticed another call to MStorageReservation.add in the method org.compiere.apps.form.Match.createMatchRecord - and is using the product and ASI from the receipt line instead of the order line, another source for problems, here also with ASI.

CarlosRuiz-globalqss (approved):

PR #969, opened 2021-11-06
IDEMPIERE-5031 Implement auto produce upon shipment for mfg light

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5031


CarlosRuiz-globalqss (approved):
tested successsfully with PatioSet in GW

unit test passed

PR #970, opened 2021-11-08
IDEMPIERE-5026 Create from Invoice showing already invoiced RMA's

tony-tspc:
This now fixes the issue with already invoiced RMA's still being shown in Create From Invoice in the "Receipt" pull down list and fixes the Create from Invoice form not showing any partially invoiced RMA’s in the RMA pulldown list.


hengsin (approved):
looks good

PR #971, opened 2021-11-08
IDEMPIERE-5034 NPE and other problems when a file is removed from store attachment filesystem

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5034


hengsin (approved):
looks good

PR #975, opened 2021-11-11
IDEMPIERE-5037 PackOut/PackIn must treat EE01/2/4/5 entries as official dictionary

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5037


hengsin (approved):
looks good

PR #976, opened 2021-11-12
IDEMPIERE-2168 Product with ASI Mandatory Type = When Shipping is sti…

hengsin:
…ll mandatory on sales orders

https://idempiere.atlassian.net/browse/IDEMPIERE-2168


CarlosRuiz-globalqss (approved):
tested

PR #980, opened 2021-11-14
IDEMPIERE-4956 fix unmatched quotes in displaylogic

tony-tspc:
These migration scripts were manually created to fix the unmatched quotes issue in 3 columns in the database. The update will only occur if the columns contain the original content as shown on the left:

AD_Field_ID = 55427 (DisplayLogic)
@IsDropShip@=Y'    ==>  @IsDropShip@='Y'

AD_Field_ID = 57734 (DisplayLogic)
@IsDropShip@=Y'    ==>    @IsDropShip@='Y'

AD_Field_ID = 59348 (DisplayLogic)
@A_SourceType@='MAN''    ==>    @A_SourceType@='MAN'

If the DisplayLogic contents have been changed, then no update will occur.


CarlosRuiz-globalqss:
Hi @tony-tspc - thanks for the fixes.

I did some further review of logic fields and found some additional errors, maybe these can be also tackled in this ticket?

1 - ad_column.readonlylogic wrong in C_POSPayment.Processed
it says just N - probably that will throw also error with the new SimpleBooleanLexer (didn't test)
I think solution is just to clear that field, the field is read-only in the window

2 - ad_field.displaylogic wrong in:
Expense Report > Report Line > Quantity Reimbursed | Price Reimbursed
it says Processed=Y -> I think it must be @Processed@=Y

3 - this is for an inactive field:
ad_field.displaylogic wrong in inactive ad_field_id=50135
AD_Package_Exp_Common.SQLStatement
@Type@='D'|Type@='SQL'
this is inactive, so it's harmless, but maybe is better to fix it just in case.

Regards,

Carlos Ruiz

tony-tspc:
Hi @CarlosRuiz-globalqss 
I have added the additional fixes as suggested. Tested OK on Postgresql.

C_POSPayment.Processed  (ReadOnlyLogic)
```
AD_Column_ID = 200452 
N ==> ''
```
ExpenseReport.Reportline.Quantity Reimbursed (DisplayLogic)
```
AD_Field_ID = 10456
Processed=Y    ==>    @Processed@=Y
```
ExpenseReport.Reportline.Price Reimbursed (DisplayLogic)
```
AD_Field_ID = 10457
Processed=Y    ==>    @Processed@=Y
```
AD_Package_Exp_Common.SQLStatement (DisplayLogic)
```
AD_Field_ID = 50135 
@Type@='D'|Type@='SQL'    ==>    @Type@='D'|@Type@='SQL' 
```
Best Regards,
Tony Snook

hengsin:
In addition to this, we have some display logic that need parenthesis fix, for e.g:

@IsSummary@='N' & @ProductType@=I | @ProductType@=S should be changed to @IsSummary@='N' & (@ProductType@=I | @ProductType@=S) for 53890,53592,59804,11782 and 3743

Perhaps there are other, needs to check further.

CarlosRuiz-globalqss:
Checked the logics with `&` and `|` combined and found these additional two:

* ad_field.displaylogic where ad_field_id in (1129,62343)
C_OrderLine.M_Warehouse_ID
`@DirectShip@='N' & @OrderType@='OB' | @OrderType@='SO' | @Processed@='Y'`
can be changed to
`@OrderType@='OB' | @OrderType@='SO' | @Processed@='Y'`
the field/context DirectShip doesn't exist, the result is harmless as is just displaying a read-only field, after the change the field will be displayed also for quotations[OB] (which was not before, not a big deal)

* ad_field.displaylogic where ad_field_id in (6551,200705)
C_Payment.C_Project_ID and C_PaymentTransaction.C_Project_ID
`@$Element_PJ@=Y & @C_Invoice_ID@=0 | @C_Invoice_ID@=''`
can be changed to
`@$Element_PJ@=Y & ( @C_Invoice_ID@=0 | @C_Invoice_ID@='' )`

also on the same line of the first:
* ad_field.displaylogic where ad_field_id in (3388)
`@DirectShip@='N'`
can be changed to 
`1=0`
because the DirectShip doesn't exist is always evaluated to false, or simply change the field to not displayed
this is the warehouse in purchase order line - harmless too

tony-tspc:
Other than those already mentioned, I found about 11 more. Will need to investigate further as some contexts like @OrderType@ are not used. Also, not sure about @#IsLiberoEnabled@

image

CarlosRuiz-globalqss:
Hi @tony-tspc - I checked those - as the logic is left to right, when all the OR conditions are at the beginning of the logic, then is the same as if they are within parenthesis.  So, those are OK, enclosing within parenthesis will make them more readable, but there is no change in the result.

About the _ParentIsSOTrx_ logic - that's correct, I tested those to come with a logic that worked in 8.2 and that's fine.

About OrderType - that doesn't exist as a column, but is a context variable set in GridTab.loadDependentInfo and also in CalloutOrder.docType - so it works fine.

About #IsLiberoEnabled - probably that's from a plugin, we don't have that column or variable in core.

Regards,

Carlos Ruiz

tony-tspc:
I have now added the following changes:

`AD_Field_ID IN (200705,6551)`
From: `@$Element_PJ@=Y & @C_Invoice_ID@=0 | @C_Invoice_ID@=''` 
To:      `@$Element_PJ@=Y & (@C_Invoice_ID@=0 | @C_Invoice_ID@='')`

`AD_Field_ID IN (1129,62343)`
From: `@DirectShip@='N' & @OrderType@='OB' | @OrderType@='SO' | @Processed@='Y'`
To:      `@OrderType@='OB' | @OrderType@='SO' | @Processed@='Y'`

`AD_Field_ID IN (3743,11782,59804,53592,53890)`
From: `@IsSummary@='N' & @ProductType@=I | @ProductType@=S`
To:      `@IsSummary@='N' & (@ProductType@=I | @ProductType@=S)`

`AD_Field_ID = 3388`
From: `@DirectShip@='N'`
To:     `1=0`

PR #982, opened 2021-11-15
IDEMPIERE-5012 org.idempiere.ui.zk.annotation.Form annotation

saulojg:
Expose Form annotation to other bundles


hengsin (approved):
trivial fix

PR #983, opened 2021-11-15
IDEMPIERE-4472 Window Advanced Search - Allow select column from wind…

hengsin:
…ow tabs - 2pack compatibility

Fix NPE


CarlosRuiz-globalqss (approved):
tested successfully

PR #984, opened 2021-11-16
IDEMPIERE-4808 Payment Allocation should be IsSOTrx neutral + Product Window

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4808


hengsin (approved):
trivial fix, good to go

PR #989, opened 2021-11-18
IDEMPIERE-5048 Minor fix for Scheduler and SvrProcess: set m_trx to null

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5048


CarlosRuiz-globalqss (approved):
looks OK - minor tests

PR #991, opened 2021-11-19
IDEMPIERE-5049 Zk Session and Desktop object not destroy immediately …

hengsin:
…after logout

https://idempiere.atlassian.net/browse/IDEMPIERE-5049


CarlosRuiz-globalqss (approved):
tested

PR #993, opened 2021-11-19
IDEMPIERE-5050 Service startup freeze

saulojg:
### Delayed service registration

Most annotation-based factories were switched to delayed OSGi component activation, in order to minimize the amount of threads simultaneously allocated to annotation scanning during service startup. This change doesn't seem to have any side effects. These factories are activated whenever needed, then the class scanning is performed synchronously upon the factory/component activation.

Classes such as `DefaultAnnotationBasedEventManager` were also switched to delayed initialization, since the `service` attribute of their `@Component` annotation shouldn't have effect as they don't implement any interface _directly_. More info on this at https://docs.osgi.org/javadoc/r4v43/cmpn/org/osgi/service/component/annotations/Component.html#service()

### Deferred class scanning

`DefaultAnnotationBasedEventManager` was modified to perform class scanning using background threads. This shouldn't pose any problems unless event processing is required very early upon service startup. A cached thread pool executor (see `java.util.concurrent.Executors.newCachedThreadPool()`) was used for this since the standard executor (i.e.: `org.compiere.Adempiere.getThreadPoolExecutor()` ) didn't solve the problem.

### DIFF
--- a/org.adempiere.base.callout/src/org/adempiere/base/callout/factory/AnnotationBasedColumnCalloutFactoryImpl.java
+++ b/org.adempiere.base.callout/src/org/adempiere/base/callout/factory/AnnotationBasedColumnCalloutFactoryImpl.java
@@ -33,7 +33,7 @@ import org.osgi.service.component.annotations.Component;
  * @author hengsin
  *
  */
-@Component(immediate = true, service = IColumnCalloutFactory.class)
+@Component(immediate = false, service = IColumnCalloutFactory.class)
 public class AnnotationBasedColumnCalloutFactoryImpl extends AnnotationBasedColumnCalloutFactory {
 
 	public AnnotationBasedColumnCalloutFactoryImpl() {
--- a/org.adempiere.base.process/src/org/adempiere/base/process/factory/BaseProcessFactoryImpl.java
+++ b/org.adempiere.base.process/src/org/adempiere/base/process/factory/BaseProcessFactoryImpl.java
@@ -33,7 +33,7 @@ import org.osgi.service.component.annotations.Component;
  * @author hengsin
  *
  */
-@Component(immediate = true, service = IProcessFactory.class, property = {"service.ranking:Integer=0"})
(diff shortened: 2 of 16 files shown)


hengsin:
It is kind of non deterministic with non-immediate service in core. I think it is probably safer to switch all scanning to asyn instead of making many core services as non-immediate.

saulojg:
> It is kind of non deterministic with non-immediate service in core. I think it is probably safer to switch all scanning to asyn instead of making many core services as non-immediate.

Deferred component activation shouldn't be non-deterministic at all, nothing changes except that the `activate()` method gets called whenever a component is needed. Only possible drawback I see is that, if we take the case of `IModelFactory`, the first call to any of its factory methods will take longer than the rest. 

On the other hand, if we perform class scanning asynchronously then the behavior _could_ be non-determinstic. If we take again `IModelFactory` as an example, calls to `IModelFactory.getClass(String)` could return null while the scanning is still in progress. In a case like that, the engine will not associate the table name with said factory. As these table name / factory associations are cached, the engine will understand that there's just no model class for that table.

hengsin:
hi @saulojg , that doesn't have to be the case with async scanning either as you can wait for it when it is invoke for the first time.

saulojg:
> hi @saulojg , that doesn't have to be the case with async scanning either as you can wait for it when it is invoke for the first time (i.e wait for Future<?> to complete in getClass(String) method).

Yes, or we could perform the scanning whenever any of the services are requested for the first time. What I'm trying to convey is that it shouldn't be necessary at all, unless OSGi is badly broken. IMHO most services shouldn't be activated immediately unless really needed (e.g.: a web services component).

https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e36929
https://thecarlhall.wordpress.com/2011/02/04/when-to-immediately-activate-an-osgi-component/

saulojg:
I would suggest to integrate and test these changes, as long as there are no other objections. Meanwhile we can work on an immediate activation + deferred scan alternative, seeking code reutilization across all annotation-based factories.

hengsin:
replace by PR 996

PR #996, opened 2021-11-21
IDEMPIERE-5050 Service startup freeze

saulojg:
Switched all annotation-based factories to asynchronous class scan.


hengsin:
hi @saulojg , both PR has pass the same freeze test ? Any other things to note here ?

saulojg:
> hi @saulojg , both PR has pass the same freeze test ? Any other things to note here ?

Yes. Since it was decided to keep immediate activation of these components, this PR should be used instead of the previous one.

There were several other issues we found while testing this branch, please see the commit log.

hengsin (approved):
tested

PR #998, opened 2021-11-22
IDEMPIERE-2620 Customer Drop Ships not implemented

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-2620

Drop BP and Location must be mandatory when Drop Ship is enabled


hengsin (approved):
looks good

PR #1006, opened 2021-11-26
IDEMPIERE-5054 Order Line: Shouldn't allow change of ASI when Qty Res…

hengsin:
…erved > 0

https://idempiere.atlassian.net/browse/IDEMPIERE-5054


CarlosRuiz-globalqss (approved):
tested

PR #1007, opened 2021-11-27
IDEMPIERE-4871 Print Format Item Script Data Type Support (fix backward compatibility issue)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4871

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -1052,7 +1052,7 @@ public class DataEngine
 								{
 									/** START DEVCOFFEE: script column **/
 									int displayType = pdc.getDisplayType();
-									if(rs.getMetaData().getColumnName(displayIndex).contains("SCRIPTCOLUMN")) {
+									if (display.startsWith("@SCRIPT")) {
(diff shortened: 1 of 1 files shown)


tony-tspc:
Hi Carlos as reported in Google Groups, everything is now working as expected. Just have an issue with what the following code highlighted in grey is trying to achieve. From what I can see the columnType will always be VARCHAR for the Script column itself, so the displayType wil be set to TEXT. Then further down the displyType is explicitly set to NUMBER.

script4

tony-tspc:
In the original IDEMPIERE-4871 commit, I think the displayType should be have been set based on the value returned from the beanshell interpreter. This would enable displaying results other than just a number. I am working on a solution to implement this if you think it would be worthwile.

tony-tspc:
Currently "NF8.2 Print Format Item Script" is not working in master, this pull request will at least allow that feature to work again as it does in 8.2 release. I will add my suggested changes to a feature request, if that is preferred.

hengsin:
Hi @tony-tspc , I will merge this and please add your suggested changes as a separate feature request ticket.

hengsin (commented):
tested by tony

tony-tspc:
Thanks @hengsin

PR #1008, opened 2021-11-27
IDEMPIERE-5058 Pack In from a tenant print format just packed out throws cross tenant error because of Print Paper

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5058


hengsin (approved):
trivial fix, good to go

PR #1009, opened 2021-11-28
IDEMPIERE-5060 iDempiere dictionary view doesn't support UNION ALL

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5060


hengsin (approved):
looks good

PR #1014, opened 2021-11-30
IDEMPIERE-2172 Attribute set with mandatory type = When Shipping not …

hengsin:
…working

https://idempiere.atlassian.net/browse/IDEMPIERE-2172


CarlosRuiz-globalqss:
Hi @hengsin - I tested this pull request.

The numbers on storage look correct, but the transactions are not reflecting what happened in storage.

Theoretically we must be able to reconstruct the storage from the M_Transaction table, in this case the inventory move that you are doing at MInOut.moveOnHandToShipmentASI is not reflected in transactions.

So, I think would be good to add the corresponding transactions there:
-1 for the non-ASI product
+1 for the new-ASI product
-1 for the new-ASI product (this is the only one being added at this moment)

CarlosRuiz-globalqss:
I'm also wondering if this must be restricted for the specific case of ASI mandatory when shipping (adding the validation before calling moveOnHandToShipmentASI)

I checked what happens if the shipment is reversed, and the storage is not the same as in the beginning, the new-ASI product is left with 1 on-hand (at the beginning there were zero).  So, if this is intentional is OK, but if is caused by a user mistake, then the inventory will be left different when reversing - that's why I think maybe better to restrict it to the specific case of "mandatory when shipping"

CarlosRuiz-globalqss:
Hmm - to make things worse with the "reverse the shipment" case.

Inventory move can be used to serialize an item - this is - move a non-ASI item to a new-ASI.

But, the opposite is not supported, I cannot de-serialize an item, cannot set new-ASI back to non-ASI, so the inventory cannot be restored to the original state easily in case of mistake.

hengsin:
alright, will add the m_transaction part. the can't de-serialize part is existing limitation that's not added by this ticket. this will just address the original issue and will not touch on that.

CarlosRuiz-globalqss (approved):
test OK

PR #1018, opened 2021-11-30
IDEMPIERE-1302 Validate flag IsPurchased and IsSold on Price List, Order, InOut, and Invoice

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-1302

Implemented as adding IsSold/IsPurchased in Product Info window with the same management as IsCustomer/IsVendor on BP Info window


hengsin (approved):
looks good

PR #1019, opened 2021-12-01
IDEMPIERE-5065 Material Receipt Line not available for matching after…

hengsin:
… reversal of Match PO

https://idempiere.atlassian.net/browse/IDEMPIERE-5065


tony-tspc:
I have retested the updated pull request and everything works as expected. Also ran all unit tests which also passed. 
Had no idea that consolidate method existed, was even in the Adempiere version which I used for many years.

CarlosRuiz-globalqss (approved):
tested successfully by @tony-tspc

PR #1027, opened 2021-12-05
IDEMPIERE-5073 M_InOutLineMA.DateMaterialPolicy should comes from MSt…

hengsin:
…orageOnHand or Shipment instead of through user entry

https://idempiere.atlassian.net/browse/IDEMPIERE-5073


org.adempiere.base/src/org/compiere/model/MInOutLineMA.java
```diff
@@ -277,6 +277,15 @@ protected boolean beforeSave (boolean newRecord)
 				dateMPolicy = inout.getMovementDate();
 			}
 			
+			setDateMaterialPolicy(dateMPolicy);
+		} else if (newRecord && getM_AttributeSetInstance_ID() > 0 && getDateMaterialPolicy() == null) {
+			I_M_InOutLine line = getM_InOutLine();
```
CarlosRuiz-globalqss, line 282: Hi @hengsin - I see this is basically copied from the lines above, two considerations:

1 - wouldn't be easier and better for maintenance to add an OR to the condition in line 264
2 - the getM_InOutLine() and line.getM_InOut() are really bad for performance - calling MTable.get(...).getPO(...) - I think for performance would be better to avoid that in core (or improve the generated methods to be faster)

CarlosRuiz-globalqss (approved):
unit test pass

PR #1031, opened 2021-12-06
IDEMPIERE-5077 Process (gear) button unusable when the workflow is suspended

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5077


hengsin (approved):
looks good

PR #1034, opened 2021-12-06
IDEMPIERE-4186 allow workflow to be suspended until after commit of actual  transaction

CarlosRuiz-globalqss:
fix Split Join = AND

https://idempiere.atlassian.net/browse/IDEMPIERE-4186

### DIFF
--- a/org.adempiere.base/src/org/compiere/wf/MWFProcess.java
+++ b/org.adempiere.base/src/org/compiere/wf/MWFProcess.java
@@ -49,7 +49,7 @@ public class MWFProcess extends X_AD_WF_Process
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -4447369291008183913L;
+	private static final long serialVersionUID = 5981488658756275526L;
 
 	/**
 	 * 	Standard Constructor
@@ -256,9 +256,9 @@ public class MWFProcess extends X_AD_WF_Process
 	
 	/**************************************************************************
 	 * 	Check Status of Activities.
-	 * 	- update Process if required
 	 * 	- start new activity
 	 * 	@param trxName transaction
+	 * 	@param lastPO PO
 	 */
 	public void checkActivities(String trxName, PO lastPO)
 	{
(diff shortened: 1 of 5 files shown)


hengsin:
will be good if we can have some unit test added for workflow stuff.

CarlosRuiz-globalqss:
Ready @hengsin - I added a unit test for _Split AND_ that is failing in current master and passing with the modifications of this pull request.

There is a comment in WFMaterialReceiptTest:122 - for this unit test I set the second wait to zero, in the test case attached via packout to the ticket I'm setting the wait to -1 for "wait until commit" - but that cannot be tested in the unit test because we don't do commit.  Do you know how could we manage such test for after commit?

hengsin:
Hi @CarlosRuiz-globalqss , I think there's 2 possible approach to this, one is to use a fake transaction object that doesn't do real commit but will fire all the commit related events. Alternative is to perform real commit and perform clean up using delete after that.

CarlosRuiz-globalqss:
Anyways, that's maybe for future, I thought the issue was related to _Wait Commit_ combined with _Split AND_ - but the unit test made me realize the problem was from _Split AND_ in general.

So, I think if you consider it OK, the ticket is OK to move ahead as is.  WaitCommit unit test can wait.

hengsin (changes requested):
unit test fail with the following error:org.opentest4j.AssertionFailedError: 
expected: <7> but was: <3>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
	at org.idempiere.test.workflow.WFMaterialReceiptTest.testSplitAndWorkflow(WFMaterialReceiptTest.java:167)

CarlosRuiz-globalqss:
Hi @hengsin, thanks for the peer review.

I committed a fix for the unit test.  Most of the static workflow code is working without Trx (which is good for performance), so I needed to commit and cleanup the changes to the workflow.

hengsin (approved):
unit test pass

PR #1038, opened 2021-12-07
IDEMPIERE-5078 - M_inout movementtype add to before save

PeterTakacs300:
When we call m_inout then - field - movementtype must be set by frontend - otherwise REST call fail.
We suggest to add to beforesave from callout - then m_inout.movementtype will be handled by backend.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -1037,7 +1037,34 @@ public class MInOut extends X_M_InOut implements DocAction
             MDocType docType = MDocType.get(getCtx(), rma.getC_DocType_ID());
             setC_DocType_ID(docType.getC_DocTypeShipment_ID());
         }
+        
+        // Set MovementType
+        if(newRecord || (get_ValueDifference("C_DocType") != null)) {
+			MDocType docType = new MDocType(getCtx(), getC_DocType_ID(), get_TrxName());
+			String DocBaseType = docType.getDocBaseType();
+			Boolean IsSOTrx = docType.isSOTrx();
+			
+			if (DocBaseType.equals("MMS"))					//	Material Shipments
+				
+				{
+						if (IsSOTrx)
+							setMovementType("C-");				// Customer Shipments
+						else
+							setMovementType("V-");				// Vendor Return
 
+				}
+				
+			else if (DocBaseType.equals("MMR"))				//	Material Receipts
+			    
+				{
+						if (IsSOTrx)
+							setMovementType("C+");				 // Customer Return
+						else
+							setMovementType("V+");				 // Vendor Receipts
+				}				
+				
+		}
+        
 		return true;
 	}	//	beforeSave
 
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -977,9 +977,19 @@ public class MInOut extends X_M_InOut implements DocAction
 		}
 	}	//	setM_Warehouse_ID
 
+	/**
+	 * Gets Movement Type based on Document Type's DocBaseType and isSOTrx
+	 * @param ctx 
+	 * @param C_DocType_ID Document Type ID
+	 * @param issotrx is sales transaction
+	 * @param trxName transaction name
+	 * @return Movement Type
+	 */
 	public static String getMovementType(Properties ctx, int C_DocType_ID, boolean issotrx, String trxName) {
 		String movementType = null;
-		MDocType docType = new MDocType(ctx, C_DocType_ID, trxName);
+		MDocType docType = MDocType.get(C_DocType_ID);
+		
+		if (docType == null) return null;
 		
         if (docType.getDocBaseType().equals(MDocType.DOCBASETYPE_MaterialDelivery)) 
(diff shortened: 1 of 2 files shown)


d-ruiz:
Hi, 

My2cents here. I prefer to wait for the opinion of Heng Sin or @CarlosRuiz-globalqss to know if this is the right approach or if this should be handled in a custom EventHandler.

Regarding the code, this is my peer review:
Please avoid hardcoding strings, such as MMS, C-, V-, V+, C+. 
That's what the existing static final variables are for. For instance instead of C- you can code MOVEMENTTYPE_CustomerShipment. It is clearer and you don't need the comment _// Customer Shipments_ anymore as it is explicit in the variable name.

In fact, you can see many points in the same class (f.i line 204 or 375) that do the same as your code but use static variables. 
Perhaps it would be better to move that logic to a method and call it from all those points.

Another question, I haven't tested but wouldn't that throw an NPE because before the code you add there's this validation:
`getMovementType().contentEquals`. If movement type = null, the system would never reach the change. Maybe there's a default value I'm unaware of, just asking.

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -1037,7 +1037,34 @@ protected boolean beforeSave (boolean newRecord)
             MDocType docType = MDocType.get(getCtx(), rma.getC_DocType_ID());
             setC_DocType_ID(docType.getC_DocTypeShipment_ID());
         }
+        
+        // Set MovementType
+        if(newRecord || (get_ValueDifference("C_DocType") != null)) {
```
hengsin, line 1042: should use is_ValueChanged("C_DocType") instead of get_ValueDifference(..)

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -1037,7 +1037,34 @@ protected boolean beforeSave (boolean newRecord)
...
+			    
+				{
+						if (IsSOTrx)
+							setMovementType("C+");				 // Customer Return
+						else
+							setMovementType("V+");				 // Vendor Receipts
+				}				
+				
+		}
+
```
hengsin, line 1067: I think this adapted from CalloutInOut.doctype. To ease future maintenance, it would be much better to refactor CalloutInOut.doctype and share the logic here (obviously, the logic of both must be the same).

PeterTakacs300:
Thank you for the comment and review, I applied the changes You suggested.

I made a static method getMovementType(Properties ctx, int C_DocType_ID, boolean issotrx, String trxName), which contains the logic, and I call it in CalloutInOut.doctype

I also made a local method setMovementType(), which calls the static one, and this local method is called in MInOut class in beforeSave, and also in the places, You pointed out, where same logic is applied.

Peter Takács, Cloudempiere

PS: I am a new developer in Cloudempiere, so excuse me for some rookie mistakes, I am still learning, and I make the commits temporarily with my colleague, Igor Pojzl's account (he also helps me a lot :) )

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -991,14 +977,40 @@ public void setM_Warehouse_ID (int M_Warehouse_ID)
 		}
 	}	//	setM_Warehouse_ID
 
+	public static String getMovementType(Properties ctx, int C_DocType_ID, boolean issotrx, String trxName) {
```
hengsin, line 980: please add javadoc

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -991,14 +977,40 @@ public void setM_Warehouse_ID (int M_Warehouse_ID)
...
+		
+        if (docType.getDocBaseType().equals(MDocType.DOCBASETYPE_MaterialDelivery)) 
+            movementType = docType.isSOTrx() ? MOVEMENTTYPE_CustomerShipment : MOVEMENTTYPE_VendorReturns; 
+        else if (docType.getDocBaseType().equals(MDocType.DOCBASETYPE_MaterialReceipt)) 
+            movementType = docType.isSOTrx() ? MOVEMENTTYPE_CustomerReturns : MOVEMENTTYPE_VendorReceipts;  
+        
+		return movementType;
+	}
 
+	public void setMovementType() {
```
hengsin, line 992: missing javadoc

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -991,14 +977,40 @@ public void setM_Warehouse_ID (int M_Warehouse_ID)
 		}
 	}	//	setM_Warehouse_ID
 
+	public static String getMovementType(Properties ctx, int C_DocType_ID, boolean issotrx, String trxName) {
+		String movementType = null;
+		MDocType docType = new MDocType(ctx, C_DocType_ID, trxName);
```
hengsin, line 982: use MDocType docType = MDocType.get(C_DocType_ID) instead. And to check C_DocType_ID is valid, we can add the checking if (docType == null) return null; (or just throw illegalargumentexception).

hengsin (changes requested):
looks good thus far, just need some minor refinements.

hengsin:
Hi @d-ruiz , actually, most of the time, it is a good practice to have the same call out logic in a model's before save method. This help to ensure the model is usable when it is not use from ZK front end where call out is not invoke.

PeterTakacs300:
Added the javadoc for the two new methods + the minor changes you mentioned.

hengsin (approved):
looks good

PR #1041, opened 2021-12-08
IDEMPIERE-5087 - Wrong behaviour on the more toolbar popup after pres…

d-ruiz:
…sing the Archived Documents button

https://idempiere.atlassian.net/browse/IDEMPIERE-5087


hengsin (approved):
looks good

PR #1043, opened 2021-12-09
IDEMPIERE-5091 - Before Save Set Default Locator From Product Or Warehouse when creating m_inout programatically

PeterTakacs300:
issue: when m_inout created programmatically locator is not set properly (atm works as callout)
clodempiere provide PR soon.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MInOutLine.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOutLine.java
@@ -544,7 +544,22 @@ public class MInOutLine extends X_M_InOutLine
 		{
 			if (getM_Locator_ID() <= 0 && getC_Charge_ID() <= 0)
 			{
-				throw new FillMandatoryException(COLUMNNAME_M_Locator_ID);
+				// Try to load Default Locator
+				 I_M_Locator product_Locator = getM_Product().getM_Locator();
+
+				MWarehouse warehouse = (MWarehouse) getParent().getM_Warehouse();
+				if(warehouse != null) {
+					MLocator defaultLocator = warehouse.getDefaultLocator();
+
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/MInOutLine.java
```diff
@@ -544,7 +544,22 @@ protected boolean beforeSave (boolean newRecord)
 		{
 			if (getM_Locator_ID() <= 0 && getC_Charge_ID() <= 0)
 			{
-				throw new FillMandatoryException(COLUMNNAME_M_Locator_ID);
+				// Try to load Default Locator
+				 I_M_Locator product_Locator = getM_Product().getM_Locator();
+
+				MWarehouse warehouse = (MWarehouse) getParent().getM_Warehouse();
+				if(warehouse != null) {
+					MLocator defaultLocator = warehouse.getDefaultLocator();
```
hengsin, line 552: It is slightly more efficient to avoid the loading of default locator if product has a valid locator.

CarlosRuiz-globalqss: BTW, those methods from X classes returning the object, like MInOut.getM_Warehouse() are highly inefficient.

It's always better to use MWarehouse.get() (with cache when the object is not going to be modified), or new MWarehouse(ctx,id,trx) constructor.

I think we must avoid to use those in core, and indeed I would recommend against using that in plugins.  It's a shame that we cannot remove them, but maybe we could:
- mark them deprecated
- or replace in the X_ class the MTable.get().getPO() with something more efficient

hengsin (changes requested):
please address the issues raise

norbertbede:
@PeterTakacs300 pls finish this task next week !

hengsin (approved):
looks good

PR #1045, opened 2021-12-10
IDEMPIERE-5078 - fix typo C_DocType should be C_DocType_ID

tony-tspc:
Found a typo in beforeSave method, C_DocType should have been C_DocType_ID


hengsin (approved):
good catch.

PR #1049, opened 2021-12-13

CarlosRuiz-globalqss:

Fix wrong matches pattern -> eclipse 2021-12 complains with error
Invalid escape sequence (valid ones are  \b  \t  \n  \f  \r  \"  \'  \\ )


hengsin (approved):
trivial fix, good to go.

PR #1053, opened 2021-12-16
IDEMPIERE-5102 Deactivate AD_Desktop and AD_Workbench

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5102


CarlosRuiz-globalqss (approved):
Hi @hengsin - I'm OK with this cleanup.

It breaks severely the APanel class in swing-client (not an easy patch).

PR #1055, opened 2021-12-16
IDEMPIERE-4448 Autocomplete/Info Window doesn't work correctly with Search fields with Reference

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4448

* Change 188 reftables to use the ID for display - this displays the reference based on the identifiers of the table, making it consistent with Search/Table without reference
* Use the MSysConfig.IDENTIFIER_SEPARATOR for the cases where reference is displaying value and another column


hengsin (approved):
tested

PR #1058, opened 2021-12-17
IDEMPIERE-5105 Remove unused and/or unnecessary code in Print Format …

tony-tspc:
…Item Script feature

I have tested this with a number of examples and all existing functionality is working as before.

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -24,7 +24,6 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Timestamp;
-import java.sql.Types;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


hengsin:
I think the original intention here is not to hard code the display type as Number only. Not sure why it get mix up in the process but  I think if possible, we should fix it in that direction instead (using either the columnType or the Java type of value return from bsh.eval).

tony-tspc:
Yes, my solution looks at the return java type from bsh.eval() and in part looks like this:
image

tony-tspc:
Even ignoring the resetting of display type to number, the other issue with the current code is the statement:
int columnType = rs.getMetaData().getColumnType(displayIndex);
This will always return the value for VARCHAR (at least in Postgres), as this is the sql type of the Script column itself, if I am understanding the code correctly.

hengsin:
hi @tony-tspc , yes, that's probably not working. Perhaps, it will works better if we use the Java class of value instead (for e.g, if (value instanceof Number) displayType = Number) ?

tony-tspc:
Hi @hengsin , This ticket is just about tidying up the code before I submit my new code. I was going to create a separate ticket for that. The new code, which I haven't yet submitted already does as you have suggested.
```
if (value instanceof Number)
        displayType = DisplayType.Number;
else if (value instanceof Boolean)
	displayType = DisplayType.YesNo;
else if (value instanceof Date)
	displayType = DisplayType.Date;
else
	displayType = DisplayType.Text;
```

hengsin (approved):
looks good

PR #1062, opened 2021-12-18
IDEMPIERE-5101-InfoGeneralPanel-do-not-show-column-headers-when-opened

gsjeffen:
All infowindows show the column headers, but this is not the case of lookup search, that uses InfoGeneralPanel. This should be enabled in order to get a consisting look and feel of iDempiere.

Examples of these lookups are RMA on Customer Returns, Cash Plan Line on Sales Order. Business Partner, Invoice and Payment on the Allocation report to name a few.


gsjeffen:
Wrong

PR #1063, opened 2021-12-19
IDEMPIERE-5111 Data type improvements to Print Format Item Script

tony-tspc:
- have tested with several example scripts and all appears to be working well


hengsin (approved):
looks good

PR #1064, opened 2021-12-19
IDEMPIERE-5113 MInfoWindow: column access doesn't work if From clause…

hengsin:
… is created without synonym/alias

https://idempiere.atlassian.net/browse/IDEMPIERE-5113


CarlosRuiz-globalqss (approved):
tested

PR #1067, opened 2021-12-19
IDEMPIERE-5110 Manufacturing performance enhancements

saulojg:


org.adempiere.base/src/org/compiere/model/MProductionLineMA.java
```diff
@@ -99,17 +100,24 @@ public static MProductionLineMA get( MProductionLine parent, int asi, Timestamp
...
-			MProductionPlan plan = new MProductionPlan(getCtx(), parentLine.getM_ProductionPlan_ID(), get_TrxName());
-			prodParent = new MProduction(getCtx(), plan.getM_Production_ID(), get_TrxName());
-		}
-		if (newRecord && prodParent.isProcessed()) {
-			log.saveError("ParentComplete", Msg.translate(getCtx(), "M_Production_ID"));
-			return false;
+		if(newRecord) {
+			String sql = "SELECT 1 FROM M_Production "
+					+ "WHERE Processed = 'Y' AND M_Production_ID IN"
+					+ " (SELECT M_Production_ID FROM M_ProductionLine WHERE M_ProductionLine_ID = ?)";
```
hengsin, line 106: I think it is more natural and efficient that this should be changed to M_Production Inner Join M_ProductionLine instead.

migration/i8.2z/oracle/202112181000_IDEMPIERE-5110.sql
```diff
@@ -0,0 +1,14 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Dic 18, 2021, 10:00:00 AM ART
+CREATE INDEX pp_product_bom_product_id_idx ON PP_Product_BOM (M_Product_ID)
+;
+
+-- Dic 18, 2021, 10:00:00 AM ART
+CREATE INDEX pp_product_bomline_pp_product_bom_id_idx ON PP_Product_BOMLine (PP_Product_BOM_ID)
```
CarlosRuiz-globalqss, line 9: Please create the indexes in dictionary and then synchronize the index from there.

Alternatively, I think you can create the index in database as this script - but then please run the "Create Table Index" process to create the index on dictionary with official IDs

NOTE aside: this create index will not work in oracle 11 because of the name length, but we already decided that we are not supporting that anymore, so, not a problem, just a note.

org.adempiere.base/src/org/compiere/model/MProductionLineMA.java
```diff
@@ -99,17 +100,24 @@ public static MProductionLineMA get( MProductionLine parent, int asi, Timestamp
...
-		if (parentLine.getM_Production_ID() > 0) {
-			prodParent = new MProduction(getCtx(), parentLine.getM_Production_ID(), get_TrxName());
-		} else {
-			MProductionPlan plan = new MProductionPlan(getCtx(), parentLine.getM_ProductionPlan_ID(), get_TrxName());
-			prodParent = new MProduction(getCtx(), plan.getM_Production_ID(), get_TrxName());
-		}
-		if (newRecord && prodParent.isProcessed()) {
-			log.saveError("ParentComplete", Msg.translate(getCtx(), "M_Production_ID"));
-			return false;
+		if(newRecord) {
```
CarlosRuiz-globalqss, line 103: This is changing the logic of beforeSave - not validating for update records, just insert, I think that's wrong

migration/i8.2z/postgresql/202112181000_IDEMPIERE-5110.sql
```diff
@@ -0,0 +1,10 @@
+-- Dic 18, 2021, 10:00:00 AM ART
+CREATE INDEX pp_product_bom_product_id_idx ON pp_product_bom (m_product_id ASC NULLS LAST)
```
CarlosRuiz-globalqss, line 2: It is better to avoid the usage of "NULLS LAST" - we better stick to postgresql defaults in this case, that's another reason why is better to generate the index from the dictionary for consistency.

CarlosRuiz-globalqss (changes requested):
I think is better to revert the changes on the beforeSave part.

As I understand the indexes are the cause of the speed improvement, but the MProductionLineMA.beforeSave changes are dangerous and breaking previous functionality.

I don't see a big gain on the speed of the changes on beforeSave as MProductionLine and MProduction are accessed there with primary key, the new queries are practically doing the same using SQL instead of the model.

Probably the speed gain is because is not validating updates anymore (just insert), but that's the dangerous part that must be avoided, it opens the possibility that a plugin update MProductionLineMA in completed production orders.

Furthermore, there are two windows to do production:
* **Production (Single Product)**: with a direct relation between ProductionLine and Production
* **Production**: with an intermediate relationship using ProductionPlan, it is disabled by default, but we preserve compatibility, and some implementations re-enable and use this window.

The changes are not taking into account the second case AFAICS, or at least is making the code harder to follow.

org.adempiere.base/src/org/compiere/model/MProductionLineMA.java
saulojg: Here I'm preserving the original logic that performed this validation upon record creation (not update). Do you want to change this? Just to confirm.

CarlosRuiz-globalqss: Ah OK - my fault - didn't notice the newRecord below, you're totally right that moving the "f (newRecord" is an improvement, but I think is better to preserve the actual logic instead of replacing it with SQL

This way:
```java
	@Override
	protected boolean beforeSave(boolean newRecord) 
	{
		if (newRecord) {
			MProductionLine parentLine = new MProductionLine(getCtx(), getM_ProductionLine_ID(), get_TrxName());
			MProduction prodParent;
			if (parentLine.getM_Production_ID() > 0) {
				prodParent = new MProduction(getCtx(), parentLine.getM_Production_ID(), get_TrxName());
			} else {
				MProductionPlan plan = new MProductionPlan(getCtx(), parentLine.getM_ProductionPlan_ID(), get_TrxName());
				prodParent = new MProduction(getCtx(), plan.getM_Production_ID(), get_TrxName());
			}
			if (prodParent.isProcessed()) {
				log.saveError("ParentComplete", Msg.translate(getCtx(), "M_Production_ID"));
				return false;
			}
		}
		return true;
	}
```

saulojg:
> I think is better to revert the changes on the beforeSave part.
> 
> As I understand the indexes are the cause of the speed improvement, but the MProductionLineMA.beforeSave changes are dangerous and breaking previous functionality.

As stated on the commit messages, missing indexes are not the only issue.

Could you please be more specific about why you consider these changes to be dangerous? The replacement code provides the exact same functionality without unnecessary database work/hits.

While designing this replacement code we've assumed a very low probability for these model classes instances to be needed in the future.

> 
> I don't see a big gain on the speed of the changes on beforeSave as MProductionLine and MProduction are accessed there with primary key, the new queries are practically doing the same using SQL instead of the model.

After running a benchmark you won't help to see that gain, after removing the 1st worst performing query on the whole process.

> 
> Probably the speed gain is because is not validating updates anymore (just insert), but that's the dangerous part that must be avoided, it opens the possibility that a plugin update MProductionLineMA in completed production orders.

This point was already clarified, but we aren't just guessing where the performance gain is, we've measured it.

> 
> Furthermore, there are two windows to do production:
> 
>     * **Production (Single Product)**: with a direct relation between ProductionLine and Production
> 
>     * **Production**: with an intermediate relationship using ProductionPlan, it is disabled by default, but we preserve compatibility, and some implementations re-enable and use this window.
> 
> 
> The changes are not taking into account the second case AFAICS, or at least is making the code harder to follow.

In fact we've covered both cases, but aside of that, these changes avoid unnecessary database work (reads) when saving `MProductLineMA` records in just about any context.

The only arguably possible downside with changes made to `MProduction` could be a future need to work with virtual columns on any of the  `MProductLines` instances within `MProduction.completeIt()`. We've assumed a very low probability for that to happen as well.

About code readability there's no comment other that we haven't change the code flow at all. What's probably not looking too good is the need to convert lists to arrays, but that can't be helped except by creating an utility method to take care of that (we haven't found none so far).

org.adempiere.base/src/org/compiere/model/MProductionLineMA.java
saulojg: Modified code to perform this check upon record creation AND update. Keeping the rest unmodified due to reasons detailed earlier.

migration/i8.2z/oracle/202112181000_IDEMPIERE-5110.sql
saulojg: Done

org.adempiere.base/src/org/compiere/model/MProductionLineMA.java
saulojg: Done

migration/i8.2z/oracle/202112181000_IDEMPIERE-5110.sql
CarlosRuiz-globalqss: I must give you a developer ID to use here:
https://wiki.idempiere.org/en/Centralized_ID_Management

I'm in a meeting - will do it after

CarlosRuiz-globalqss:
OK - I did a visual code review - please give me a bit more time to do real tests

saulojg:
Cancelling pull request as we're going to try a different approach.

PR #1068, opened 2021-12-19
IDEMPIERE-4262 on some condition, record insert by direct sql missing UUID

CarlosRuiz-globalqss:
fix role access uuids

https://idempiere.atlassian.net/browse/IDEMPIERE-4262


hengsin (approved):
looks good.

PR #1069, opened 2021-12-20
IDEMPIERE-5114 MProjectTypePhase: getTasks() should return only activ…

hengsin:
…e task records

https://idempiere.atlassian.net/browse/IDEMPIERE-5114


CarlosRuiz-globalqss (approved):
Looks good

PR #1070, opened 2021-12-20
IDEMPIERE-3476 2Pack: attribute set install is not reference by UUID

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-3476


hengsin (approved):
looks good

PR #1083, opened 2021-12-23
IDEMPIERE-5122: Enforce Product UOM not work when user update UOM con…

dpansheriya:
…version record


CarlosRuiz-globalqss (approved):
tested

PR #1084, opened 2021-12-23
IDEMPIERE-5123: On location dialog and location Editor, Region transl…

dpansheriya:
…ation is not respected.


CarlosRuiz-globalqss (approved):
tested

PR #1088, opened 2021-12-23
IDEMPIERE-5109 Release 9 Horizon

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5109

- Fill UUID in missing AD_TreeNodeMM_UU


hengsin:
Hi @CarlosRuiz-globalqss , shouldn't we rename i8.2z to i9 instead of i8.2 to i9 ?

CarlosRuiz-globalqss:
Hi @hengsin - do you mean the commit https://github.com/idempiere/idempiere/pull/1088/commits/617c6c3 ?

Is not renaming the folder, just moving the build.xml files.

Those build.xml files are useless I think, they were some legacy way to run psql via ant, now I'm just leaving them in the new folder, they are doing the same purpose as a README or a .gitignore, just to create the folder with some file.

The folders i8.2 and i8.2z will stay there, and the new version will have a i9 folder where we put the scripts for the release 9 maintenance.

My plan is also on master to create a folder named iD10 - to put the release 10 scripts there.
The reason to use iD10 instead of i10 is because the normal ordering of folders will make the i10 to be shown and processed before i9, so, from version 10 to 99 we'll have iD as prefix.

CarlosRuiz-globalqss:
BTW - the release 9 is deployed to the server https://test-feature.idempiere.org:3243/

hengsin (approved):
looks good

PR #1094, opened 2021-12-25
IDEMPIERE-5101 InfoGeneralPanel don’t show column headers when opened

gsjeffen:
All infowindows show the column headers, but this is not the case of lookup search, that uses InfoGeneralPanel. This should be enabled in order to get a consisting look and feel of iDempiere.

This was not captured in the ticket IDEMPIERE-4863 “Enable save column width in WListBox” where the use of WListbox in the infoWindows where implemented.

Examples of these lookups are RMA on Customer Returns, Cash Plan Line on Sales Order.


hengsin (approved):
looks good.

PR #1095, opened 2021-12-26
IDEMPIERE-5126 Virtual column lazy loading

saulojg:
Changes made to `PO` and `Query` classes in order to load virtual columns on a per-needed fashion.

Additionally we've changed the model generator so that it ignores virtual columns having the `@SQL` prefix.


org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1394,12 +1400,18 @@ protected boolean load (String uuID,String trxName)
 		{
 			if (i != 0)
 				sql.append(",");
-			String columnSQL = p_info.getColumnSQL(i);
-			if (!p_info.isVirtualColumn(i))
+			String columnSQL; 
+			if (p_info.isVirtualColumn(i))
+			{
+				// initialize with NULL - values will be lazy-loaded by getters
+				columnSQL = "NULL AS " + p_info.getColumnName(i);
```
hengsin, line 1407: I think this needs to be changed to DB.getDatabase().quoteColumnName(p_info.getColumnName(i)) since p_info.getColumnName(i) could be a db reserved word.

saulojg: Yes, or it could simply contain spaces (not sure how the model generator would react to that, though). I deliberately tried to leave the original code untouched, which is:

```
			String columnSQL = p_info.getColumnSQL(i);
			if (!p_info.isVirtualColumn(i))
			{
				columnSQL = DB.getDatabase().quoteColumnName(columnSQL);
			}
			sql.append(columnSQL);	//	Normal and Virtual Column
```

I doesn't seem to make sense to treat virtual columns differently, especially when their names come from metadata.

Maybe a better approach could be to make this quoting at `POInfo.getColumnSQL(int)`

We'll make that change during the day. Thanks for reviewing.

saulojg: This could be probably out of the scope of this issue, but quoting the name assigned to virtual columns should be also performed at `POInfo.buildSelect(boolean, boolean)`, which is currently used by `Query` in most cases. This could be an indicator that this is not a current issue with iDempiere core. In any case, if developers use a column name that is a reserved keyword they should notice and fix that problem on the spot.

org.idempiere.test/src/org/idempiere/test/base/POTest.java
```diff
@@ -449,4 +451,13 @@ public void testOptimisticLocking()
 		msg2 = new MMessage(Env.getCtx(), msg1.getAD_Message_ID(), getTrxName());
 		assertEquals(msg1.getMsgText(), msg2.getMsgText());
 	}
+
+	@Test
+	public void testVirtualColumnLoad() {
+		MTest testPo = new MTest(Env.getCtx(), getClass().getName(), 1);
+		testPo.save();
+		BigDecimal expected = new BigDecimal(123.45d).setScale(2, RoundingMode.HALF_DOWN);
+		assertTrue(expected.compareTo(testPo.getTestVirtualQty()) == 0);
```
hengsin, line 460: assertEqual is a better choice here. With assertEqual, when it fail, the error message will shown you what's the expected value and what's the actual value.

saulojg: Except that there's no support for `BigDecimal`. In fact, we've changed our original code based on `BigDecimal.doubleValue()` just to keep consistency with other tests we've found (not a thorough analysis though). Maybe this should be addressed for all similar test cases in a separate ticket, perhaps involving Hamcrest. Let me know if you think otherwise.

hengsin: It does works for BigDecimal, it fall back to the standard Object.equals method (i.e assertEqual(Object, Object) which works for BigDecimal too.

saulojg: No, that's not how you'd want to compare two BDs, ever. Please see https://github.com/junit-team/junit4/issues/95

hengsin: It doesn't sound as severe as you make it looks like. As long as the two BD is having the same precision, it work just fine.

org.idempiere.test/src/org/idempiere/test/base/QueryTest.java
```diff
@@ -333,4 +336,23 @@ public void testOnlySelection() throws Exception
...
+	public void testVirtualColumnLoad() {
+		// create bogus record
+		MTest testPo = new MTest(Env.getCtx(), getClass().getName(), 1);
+		testPo.save();
+
+		BigDecimal expected = new BigDecimal(123.45d).setScale(2, RoundingMode.HALF_DOWN);
+
+		// test with virtual column lazy loading
+		Query query = new Query(Env.getCtx(), MTest.Table_Name, MTest.COLUMNNAME_Test_ID + "=?", getTrxName());
+		testPo = query.setParameters(testPo.get_ID()).first();
```
tony-tspc, line 350: Running this test on its own works, but when running all unit tests I am getting a Class Exception here:
java.lang.ClassCastException: class org.idempiere.test.model.MappedModelFactoryTest$MyTest2 cannot be cast to class org.compiere.model.MTest (org.idempiere.test.model.MappedModelFactoryTest$MyTest2 is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @24949509; org.compiere.model.MTest is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @67b28368)
	at org.idempiere.test.base.QueryTest.testVirtualColumnLoad(QueryTest.java:350)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)

saulojg: Good catch, it's fixed already. Thank you for testing.

hengsin:
For performance sake, I think it is useful to add a new constructor that take an array of virtual columns that's to be loaded immediately ( i.e PO(Properties ctx, int id, String trxName, String[] virtualColumns) ). Perhaps, the same could be added to Query, i.e instead of just 1 boolean for all virtual columns, allow developer to specify the list of virtual columns that's to be loaded immediately.

saulojg:
> For performance sake, I think it is useful to add a new constructor that take an array of virtual columns that's to be loaded immediately ( i.e PO(Properties ctx, int id, String trxName, String[] virtualColumns) ). Perhaps, the same could be added to Query, i.e instead of just 1 boolean for all virtual columns, allow developer to specify the list of virtual columns that's to be loaded immediately.

Ok

saulojg:
> For performance sake, I think it is useful to add a new constructor that take an array of virtual columns that's to be loaded immediately ( i.e PO(Properties ctx, int id, String trxName, String[] virtualColumns) ). Perhaps, the same could be added to Query, i.e instead of just 1 boolean for all virtual columns, allow developer to specify the list of virtual columns that's to be loaded immediately.

Submitted changes to both `PO` and `Query`. 

Model classes needing this feature will need to have their _X_ classes generated and their constructors updated. For now I've done this only for the `Test` table.

saulojg:
> The model class generator now uses `PO.PO(Properties, int, String, String...)` _instead_ of `PO.PO(Properties, int, String)`. This doesn't break any existing class but it might break classes instantiated via reflection. There's the option to keep both constructors in X classes.

Just changed this behavior so that the model class generator includes both constructors. Also made `AbstractModelFactory` support model classes that use the constructor `PO.PO(Properties, int, String, String...)` as a replacement of `PO.PO(Properties, int, String)`.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1394,12 +1418,34 @@ protected boolean load (String uuID,String trxName)
...
+							break;
+						}
+					}
+				}
+
+				if(lazyLoad)
+					// initialize with NULL - values will be lazy-loaded by getters
+					columnSQL = "NULL AS " + DB.getDatabase().quoteColumnName(p_info.getColumnName(i));
+				else
+					columnSQL = p_info.getColumnSQL(i);
```
hengsin, line 1441: needs to update loadedVirtualColumns here

org.idempiere.test/src/org/idempiere/test/base/POTest.java
```diff
@@ -449,4 +451,19 @@ public void testOptimisticLocking()
...
+		MTest testPo = new MTest(Env.getCtx(), getClass().getName(), 1);
+		testPo.save();
+
+		// asynchronous (default) virtual column loading
+		BigDecimal expected = new BigDecimal(123.45d).setScale(2, RoundingMode.HALF_DOWN);
+		assertTrue(expected.compareTo(testPo.getTestVirtualQty()) == 0);
+
+		// synchronous virtual column loading
+		testPo = new MTest(Env.getCtx(), testPo.get_ID(), getTrxName(), MTest.COLUMNNAME_TestVirtualQty);
+		assertTrue(expected.compareTo(testPo.getTestVirtualQty()) == 0);
```
hengsin, line 466: would be nice if we can add test to assert for both lazy and eager loading of virtual columns (probably need to add method to support such assertion).

org.adempiere.base/src/org/compiere/model/PO.java
saulojg: Not really, the set is updated automatically whenever there's data for these columns. This is performed at `PO.loadColumn(ResultSet, int)`.

org.idempiere.test/src/org/idempiere/test/base/POTest.java
saulojg: Right now the test just makes sure that things are working. How the data is being loaded is confirmed by debugger / code coverage.
There could be a hack-ish way to better confirm this by using `PO.get_ValueOld(int)`. What do you think?

org.adempiere.base/src/org/compiere/model/PO.java
hengsin: hmm ... can you clarify how would this work if the virtual column return a null value ?

org.idempiere.test/src/org/idempiere/test/base/POTest.java
hengsin: I guess the straightforward approach is to add method that allow us to query the state of the virtual columns. for e.g, isVirtualColumnLoaded or getVirtualColumnsNotLoaded (return array of virtual columns that's not loaded yet).

org.adempiere.base/src/org/compiere/model/PO.java
saulojg: We've found a cleaner (and more efficient) approach to do this, without `Null.NULL`.  Virtual columns whose queries return `NULL` are treated no different than then rest.

org.idempiere.test/src/org/idempiere/test/base/POTest.java
saulojg: While I agree on the goal, I believe there's currently no way for testing this properly, without using reflection or some other hack. We could expose and check the set of virtual columns loaded but that wouldn't guarantee that the values were properly loaded. We'd have to peek into `PO.m_oldValues`, which is private.
At time of this writing all tests are passing and code coverage is OK for all relevant paths.
Now if there are no other observations that can prevent the merging of this branch, we're switching our focus to other (unreported) core issues.

hengsin: m_oldValues is private but you can use get_ValueOld method to access it. The current test doesn't cover the difference between PO(ctx, id, trxName) and PO(ctx, id, trxName, VirtualColumns ...) so it would be nice if we can get that cover as well.

hengsin:
[1095_patch.txt](https://github.com/idempiere/idempiere/files/7812329/1095_patch.txt)

Please merge the attach patch to resolve Eclipse warnings.

saulojg:
> [1095_patch.txt](https://github.com/idempiere/idempiere/files/7812329/1095_patch.txt)
> 
> Please merge the attach patch to resolve Eclipse warnings.

Done. Thank you.

hengsin (approved):
tested

PR #1102, opened 2022-01-03
IDEMPIERE-3561 QuickEntry panel doesn't use ZoomCondition

nmicoud:
I've added migration scripts in ID10 folder.
Not sure if they should have been in i9


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2225,7 +2225,24 @@ protected int getADWindowID() {
...
+				if (isSOTrx.equals("Y") && userDef.getAD_Window_ID() > 0)
+					return userDef.getAD_Window_ID();
+				else if (isSOTrx.equals("N") && userDef.getPO_Window_ID() > 0)
+					return userDef.getPO_Window_ID();	
+			}
+
+			if (isSOTrx.equals("Y") && infoWindow.getAD_Window_ID() > 0)
+				return infoWindow.getAD_Window_ID();
+			else if (isSOTrx.equals("N") && infoWindow.getPO_Window_ID() > 0)
+				return infoWindow.getPO_Window_ID();
```
hengsin, line 2243: for isSOTrx.equals("N"), we should return infoWindow.getAD_Window_ID() if infoWindow.getPO_Window_ID()== 0 and infoWindow.getAD_Window_ID() > 0

hengsin (changes requested):

hengsin:
Hi @CarlosRuiz-globalqss, do you wants to add this to 9 ?

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss, do you wants to add this to 9 ?

Yes, I think is a bug fix, sounds better for 9

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
nmicoud: done

hengsin (approved):
tested

PR #1103, opened 2022-01-03
IDEMPIERE-5133 Organization Window & Tab newly created records does not show on idempiere but its created in the database

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5133


hengsin (approved):
looks good

PR #1110, opened 2022-01-04
IDEMPIERE-5137 Login page reload continuously using http

hengsin:
add https check to index.zul


CarlosRuiz-globalqss:
Hi @hengsin - I tested this in demo.globalqss.com and it makes the proxy forward fails.

The nginx configuration there looks like this:
https://wiki.idempiere.org/en/Proxy_iDempiere_Through_Nginx

The proxy forwarding is done to the port http.  I think is a common practice.

I think I read somewhere that is recommended when the nginx and the proxied app are in the same server, performance is better as it doesn't need to run the encryption part in the jetty side.

hengsin:
Does adding "proxy_set_header X-Forwarded-Proto: https" fix it ?

hengsin:
looks like with jetty10, you have to use the Forwarded header instead - https://www.eclipse.org/jetty/documentation/jetty-10/operations-guide/index.html#og-protocols-proxy-forwarded

hengsin:
hmm .., looks like we are missing this https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-server/src/main/config/etc/jetty-http-forwarded.xml in our configuration.

CarlosRuiz-globalqss:
How must I configure the nginx to do the test?

BTW - I'm curious about what caused this, is it a zk issue?  Or it was introduced in some ticket?

hengsin:
some additional info here - https://www.eclipse.org/jetty/documentation/jetty-10/operations-guide/index.html#og-protocols-proxy

hengsin:
It is the combination of secure only session cookie and I've turn off the sending of jsessionid in server push request (jsessionid param is a security risk).

hengsin:
Hi @CarlosRuiz-globalqss ,

Tested working with nginx with the following changes:

- add "proxy_set_header  X-Forwarded-Proto $scheme;" to both /webui and /webui/zkau/comet (the equivalent for haproxy should be "http-request set-header X-Forwarded-Proto")
- add etc/jetty-http-forwarded.xml (we should add this anyway as noted in the jetty link I posted above)

CarlosRuiz-globalqss (approved):
Tested successfully in:
https://test-feature.idempiere.org:3243/webui/
http://test-feature.idempiere.org:3280/webui/

Working fine with the changes described:
https://wiki.idempiere.org/w-en/index.php?title=Proxy_iDempiere_Through_Nginx&type=revision&diff=19881&oldid=19338

CarlosRuiz-globalqss:
No preference from my side, whatever is better for performance and security.

Is both are the same, then could be better to keep running in the actually
configured servers.


hengsin ***@***.***> schrieb am Mi., 5. Jan. 2022, 2:13 PM:

> Hi @CarlosRuiz-globalqss <https://github.com/CarlosRuiz-globalqss> ,
>
> Ok to proceed as it is or do you think we should use a javascript only
> approach instead that doesn't depends on the correct configuration of
> reverse proxy (I've a patch for that too) ?
>

hengsin:
alright, I will proceed as it is. If there are issue later from other, we can switch to a javascript only approach.

PR #1113, opened 2022-01-04
IDEMPIERE-5136 : Set Messages at tenant level

nmicoud:


nmicoud:
Hi,

The only issue is when logged in GardenWorld and adding a record in the new window with language = es_CO, after saving the record, the translation from System is shown.
If you reopen the window, the record disappear

org.adempiere.base/src/org/compiere/model/MSysConfig.java
```diff
@@ -839,4 +842,20 @@ public String toString()
 			+"]";
 	}
 
+	/** Returns list of clientID with a System Configurator equals to value */
```
CarlosRuiz-globalqss, line 845: The name of this method MSysConfig.getAll is misleading

But, I think is not necessary, you can drop this method, see comment below in Msg.

org.adempiere.base/src/org/compiere/util/Msg.java
```diff
@@ -143,29 +144,30 @@ private Msg()
...
-				StringBuilder MsgText = new StringBuilder();
-				MsgText.append(rs.getString(2));
-				String MsgTip = rs.getString(3);
-				//
-				if (MsgTip != null)			//	messageTip on next line, if exists
-					MsgText.append(" ").append(SEPARATOR).append(MsgTip);
-				msg.put(AD_Message, MsgText.toString());
+			addMessagesInCache(pstmt, rs, msg);
+
+			for (int clientID : MSysConfig.getAll(MSysConfig.MESSAGES_AT_TENANT_LEVEL, "Y")) {
```
CarlosRuiz-globalqss, line 162: This is not required, no need to loop between clients, you simply need to add to the cache this same query with:
`t.AD_Client_ID != 0`

CarlosRuiz-globalqss (changes requested):

org.adempiere.base/src/org/compiere/util/Msg.java
nmicoud: done

CarlosRuiz-globalqss:
I'm testing this in https://test-feature.idempiere.org:3243/webui/index.zul

Two recommendations:
- Because _Tenant_ is still an "internal" notation, I think is better to use _Client level messages_ as the name of the window (there is a ticket pending to change all client->tenant, but it doesn't have progress)
- Better to set the _Tenant level messages_ window to show non-grid by default (this is the default in all the system)
- The where clause in the tab is not working (if you open the window and refresh, then System records are shown).  To solve this you must define the column AD_Message_Trl.AD_Client_ID to be a Parent link column - this also solves the issue you mentioned above
- it can be helpful to have a callout that brings the system translation when changing the fields message or language

nmicoud:
Migration scripts updated.

For the callout, I use CostAdjustmentLineASI as template. Not sure if it was the best practice to follow

org.adempiere.base.callout/src/org/adempiere/base/callout/MessageTrl.java
```diff
@@ -0,0 +1,65 @@
...
+				messageID = (Integer) value;
+				lang = mTab.get_ValueAsString("AD_Language");
+			}
+			else if (mField.getColumnName().equals("AD_Language")) {
+				messageID = mTab.getValue("AD_Message_ID") != null ? (Integer) mTab.getValue("AD_Message_ID") : 0;
+				lang = (String) value;
+			}
+
+			mTab.setValue("MsgText", Msg.getMsg(lang, MMessage.get(ctx, messageID).getValue(), true));
+			mTab.setValue("MsgTip", Msg.getMsg(lang, MMessage.get(ctx, messageID).getValue(), false));
```
hengsin, line 61: As per @CarlosRuiz-globalqss  suggestion, this callout is to copy message text from system to tenant, so it should have ad_client_id check here (i.e if Env.getAD_Client_ID(Env.getCtx()) > 0). Also, should get ad_message record from system tenant (i.e with ad_client_id=0) using Query instead of MMessage.get.

org.adempiere.base/src/org/compiere/util/Msg.java
```diff
@@ -143,30 +144,29 @@ private Msg()
...
-				String MsgTip = rs.getString(3);
-				//
-				if (MsgTip != null)			//	messageTip on next line, if exists
-					MsgText.append(" ").append(SEPARATOR).append(MsgTip);
-				msg.put(AD_Message, MsgText.toString());
-			}
+			addMessagesInCache(pstmt, rs, msg);
+
+			// load translated messages at tenant level (using AD_Client_ID|Value as key)
+			pstmt = DB.prepareStatement("SELECT t.AD_Client_ID || '|' || m.Value, t.MsgText, t.MsgTip"
```
hengsin, line 163: you need to close the previous pstmt before create a new one

org.adempiere.base/src/org/compiere/util/Msg.java
```diff
@@ -143,30 +144,29 @@ private Msg()
...
+					+ " FROM AD_Message_Trl t, AD_Message m"
+					+ " WHERE m.AD_Message_ID=t.AD_Message_ID"
+					+ " AND t.AD_Client_ID != 0"
+					+ " AND t.AD_Language=?", null);
```
hengsin, line 167: missing isactive check

org.adempiere.base/src/org/compiere/util/Msg.java
```diff
@@ -143,30 +144,29 @@ private Msg()
...
+			
 			if (AD_Language == null || AD_Language.length() == 0 || Env.isBaseLanguage(AD_Language, "AD_Language"))
 				pstmt = DB.prepareStatement("SELECT Value, MsgText, MsgTip FROM AD_Message",  null);
 			else
 			{
 				pstmt = DB.prepareStatement("SELECT m.Value, t.MsgText, t.MsgTip "
 					+ "FROM AD_Message_Trl t, AD_Message m "
 					+ "WHERE m.AD_Message_ID=t.AD_Message_ID"
+					+ " AND t.AD_Client_ID = 0" // load only translated messages at System level (using Value as key)
 					+ " AND t.AD_Language=?", null);
```
hengsin, line 156: missing isactive check

org.adempiere.base/src/org/compiere/util/Msg.java
hengsin: Anyway, it seems unnecessary to break the loading in 2 sql statement. You just need to retrieve AD_Client_ID and check that in the addMessagesInCache method (ie if (AD_Client_ID > 0) AD_Message = AD_Client_ID+"|"+AD_Message).

org.adempiere.base.callout/src/org/adempiere/base/callout/MessageTrl.java
nmicoud: I understand the adding of  if Env.getAD_Client_ID(Env.getCtx()) > 0.
But I do not see the interest of adding ad_client_id = 0 when retrieving ad_message ; as records in ad_message are always with ad_client_id = 0. Am I missing something ?

hengsin: yes, miss that, the ad_client_id=0 is a must here.

org.adempiere.base/src/org/compiere/util/Msg.java
nmicoud: In fact, yes we must => if you use Base language as language when you log in, initMsg will only load records from AD_Message (using the first query). The second one (which read data from both tables is used when user select a different language.

And I didn't find a solution to avoid those 2 separate queries

org.adempiere.base.callout/src/org/adempiere/base/callout/MessageTrl.java
nmicoud: check for current tenant added

org.adempiere.base/src/org/compiere/util/Msg.java
hengsin: ok but please the isactive and close of pstmt issue.

nmicoud: Ok, should the rs be closed too ?
If so, DB.close(rs, pstmt); ?

hengsin: Just close pstmt is sufficient. rs is auto close when you close pstmt.

nmicoud: ok, done

CarlosRuiz-globalqss:
@nmicoud - I would suggest the following patch:
[pr1113_PR.txt](https://github.com/idempiere/idempiere/files/7813085/pr1113_PR.txt)

Passing a null resultset to be executed within the function is a bit obscure.

org.adempiere.base/src/org/compiere/util/Msg.java
hengsin: Should have t.IsActive='Y' too

org.adempiere.base/src/org/compiere/util/Msg.java
hengsin: Should have t.IsActive='Y' too

hengsin:
I know we have been doing DB.close(rs, pstmt) but it is kind of weird that we continue to do that even though it is not really the recommended approach (see https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/Statement.html#close()).

org.adempiere.base/src/org/compiere/util/Msg.java
nmicoud: hum, weird, i added the 3 missing checks (one per query) ; can you recheck ?

hengsin: you added IsActive check for AD_Message but not AD_Message_Trl

nmicoud: ok done

CarlosRuiz-globalqss:
maybe is because of caution, in case JDBC drivers don't implement that correctly, this probably explain why compiere had this in code:
https://coddingbuddy.com/article/59247377/must-jdbc-resultsets-and-statements-be-closed-separately-although-the-connection-is-closed-afterwards
> I encountered this with Oracle 9i driver.The JDBC contract says that the statement should close the resultset, but i have observed that this is not the implementation of the driver.If your resultset is a cursor in oracle, closing the statement does nothing!

Also there:
> Close ResultSet object as soon as you finish working with ResultSet object even though Statement object closes the ResultSet object implicitly when it closes, closing ResultSet explicitly gives chance to garbage collector to recollect memory as early as possible because ResultSet object may occupy lot of memory depending on query

So, I think maybe is not a bad practice after all.

CarlosRuiz-globalqss (approved):
tested successfully

PR #1115, opened 2022-01-06
IDEMPIERE-5063 Improve Unit Tests - fix ProductionTest RollUp test

tony-tspc:
- set mulch product category to Standard Costing category, so that the standard cost of the component is compared to the standard cost of the parent, mulchx.
Currently the test is comparing the Average PO cost of mulch with the rolled up Standard Cost of mulchx. In the seed database the Average PO and Standard Cost of mulch are the same, so the test passes.


tony-tspc:
I have also discovered other issues, will update soon

hengsin (approved):
looks good

PR #1118, opened 2022-01-07
IDEMPIERE-5142 Auto Deactivate price list version after user deactive…

hengsin:
… price list

https://idempiere.atlassian.net/browse/IDEMPIERE-5142

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MPriceList.java
+++ b/org.adempiere.base/src/org/compiere/model/MPriceList.java
@@ -18,6 +18,7 @@ package org.compiere.model;
 
 import java.sql.ResultSet;
 import java.sql.Timestamp;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I've always recommended against inactivating children records when inactivating a parent.

The reason is this potential scenario:

* Somebody has one price list with 10 versions, 2 of them are inactive by some historical reasons
* User mistakenly click the active button and save it -> all versions are inactivated
* The user notices the mistake and reactivate the price list
  * With this current pull request the 10 versions will remain inactive and the user must figure out which were the 2 that were really inactive - probably from change log if is something made long time ago, hoping that they preserve that
  * If we do the change to activate automatically versions when price list is reactivated, then we end with 10 versions active, which was not the original scenario

PR #1124, opened 2022-01-10
IDEMPIERE-5147  Improve Query.first() methods - performance

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5147

Question: Do i need to move 
`String sql = buildSQL(null, true);`
in TRY block? 
buildSQL can throw IllegalStateException. Which can cause that pageSize is not set to original value.

### DIFF
--- a/org.idempiere.test/src/org/idempiere/test/base/QueryTest.java
+++ b/org.idempiere.test/src/org/idempiere/test/base/QueryTest.java
@@ -233,7 +233,6 @@ public class QueryTest extends AbstractTestCase {
 
 	@Test
 	public void testPaging() {
-		System.setProperty("org.idempiere.db.debug", "true");
 		DB.executeUpdateEx("DELETE FROM Test WHERE Name LIKE 'QueryTest%'", getTrxName());
 		for (int i=101; i<=130; i++) {
 			PO testPo = new MTest(Env.getCtx(), "QueryTest", i);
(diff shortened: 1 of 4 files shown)


CarlosRuiz-globalqss:
> Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5147
> 
> Question: Do i need to move `String sql = buildSQL(null, true);` in TRY block? buildSQL can throw IllegalStateException. Which can cause that pageSize is not set to original value.

Yes, please

igorpojzl:
Commited.

CarlosRuiz-globalqss (changes requested):
Hi @igorpojzl

I tested this against oracle and the impact on the database is too high, because the poor implementation of the paging for oracle.

Can you please integrate the following patch:
[pr1124_PR.txt](https://github.com/idempiere/idempiere/files/7847290/pr1124_PR.txt)

I added changes to the oracle paging - and the same works perfect for postgresql.

Also added a unit test case for this paging, and it works fine.

org.idempiere.test/src/org/idempiere/test/base/QueryTest.java
```diff
@@ -230,7 +230,42 @@ public void testFirstIdOnly() throws Exception
 			.firstIdOnly();
 		});
 	}
-	
+
+	@Test
+	public void testPaging() {
+		System.setProperty("org.idempiere.db.debug", "true");
```
CarlosRuiz-globalqss, line 236: Sorry @igorpojzl  - can you please delete this line - I left that by mistake

CarlosRuiz-globalqss (approved):

PR #1130, opened 2022-01-12
IDEMPIERE-5152 : Add title on broadcast messages

nmicoud:
I also set AD_BroadcastMessage_Trl.BroadcastMessage IsAlwaysUpdateable to 'N' to make it consistent with header


nmicoud:
I also remove "width: 370px; " from the style of divAlign so the message can be displayed using all the width (in my tests, seems ok)

org.adempiere.base/src/org/compiere/model/X_AD_BroadcastMessage.java
```diff
@@ -47,6 +47,20 @@ public X_AD_BroadcastMessage (Properties ctx, int AD_BroadcastMessage_ID, String
...
+      /** if (AD_BroadcastMessage_ID == 0)
+        {
+			setAD_BroadcastMessage_ID (0);
+			setBroadcastFrequency (null);
+			setBroadcastMessage (null);
+			setBroadcastType (null);
+			setTarget (null);
+        } */
+    }
+
```
hengsin, line 63: as it is, if your changes is for release_9 branch, you have to re-generate model using release_9 branch. The model class generated from master branch is not compatible with release_9 source.

nmicoud: done

CarlosRuiz-globalqss:
> I also remove "width: 370px; " from the style of divAlign so the message can be displayed using all the width (in my tests, seems ok)

did you test it in mobile?

nmicoud:
just now ; using Firefox developer tools, seems ok to me

CarlosRuiz-globalqss:
> just now ; using Firefox developer tools, seems ok to me

I just did a test here:
https://test-feature.idempiere.org:3243/webui/index.zul

You can try logging in with GardenAdmin user.
The window is not really usable in mobile - it doesn't show the full message, it cannot be resized, and it cannot be acknowledged  (I suspect this was not introduced by your changes)

nmicoud:
ok, so let's remove this change from the ticket

I think there are other improvements for the style (at the moment, vertical-align: middle; text-align: center; is hardcoded ; that should be better to be able to manage that per message)

CarlosRuiz-globalqss:
> ok, so let's remove this change from the ticket

tested in test.idempiere.org - is not caused by you, and indeed it seems your change made it slightly better, but still failing

nmicoud:
so let's keep it :)
Anyway, wouldn't be better to be able to manage the whole style from AD_BroadcastMessage ?

org.adempiere.base/src/org/compiere/model/X_AD_BroadcastMessage.java
CarlosRuiz-globalqss: > as it is, if your changes is for release_9 branch, you have to re-generate model using release_9 branch. The model class generated from master branch is not compatible with release_9 source.

This is going to do the maintenance of release-9 and master too difficult - I think better we are needing to integrate the changes on IDEMPIERE-5126

nmicoud:
done
Ability to move the panel is a good thing !
Thanks

CarlosRuiz-globalqss (approved):

PR #1136, opened 2022-01-16
IDEMPIERE-5126 Virtual column lazy loading - Propagate new constructor

saulojg:
Added new constructor PO(Properties, int, String, String...) to most
model classes.


CarlosRuiz-globalqss (approved):
basic test in UI OK
all unit tests passed

PR #1138, opened 2022-01-18
IDEMPIERE-4962 Performance: change PostgreSQL native dialect mode default to true

CarlosRuiz-globalqss:
reverts partially f4894c4cb

https://idempiere.atlassian.net/browse/IDEMPIERE-4962
https://idempiere.atlassian.net/browse/IDEMPIERE-4465

Regression for problem solved at IDEMPIERE-4465 Native Postgres - Error on Convert Layer
Revert the changes on Convert_PostgreSQL and Convert_PostgreSQLTest
Fix the Convert_PostgreSQLTest to allow testing native and non-native cases

### DIFF
--- a/org.compiere.db.postgresql.provider/src/org/compiere/dbPort/Convert_PostgreSQL.java
+++ b/org.compiere.db.postgresql.provider/src/org/compiere/dbPort/Convert_PostgreSQL.java
@@ -81,15 +81,64 @@ public class Convert_PostgreSQL extends Convert_SQL92 {
 	 */
 	protected ArrayList<String> convertStatement(String sqlStatement) {
 		ArrayList<String> result = new ArrayList<String>();
-		
-		String statement = null;
+		/** Vector to save previous values of quoted strings **/
+		Vector<String> retVars = new Vector<String>();
+
+		String nonce = sharedNonce;
+
+		// check for collision with nonce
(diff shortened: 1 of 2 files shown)


hengsin:
I don't like the changes to the unit test. To make the convert postgresql part testable, we just need to extract the convert postgresql part to public convertPostgresqlStatement and call it from unit test (similar to convertOracleStatement). DB.getDatabase() depends on the idempiere.properties setting and make unit test harder.

CarlosRuiz-globalqss:
> I don't like the changes to the unit test. To make the convert postgresql part testable, we just need to extract the convert postgresql part to public convertPostgresqlStatement and call it from unit test (similar to convertOracleStatement). DB.getDatabase() depends on the idempiere.properties setting and make unit test harder.

I just reverted the commit that caused the problem (big issue I think) - and as the Tests are broken, then I needed to adapt the test cases.

Personally, I think the change on the Unit Test is better this way, because it mimics correctly what the application is doing, with just the convertOracleStatement approach this issue is not discovered by the unit test.


But anyways, reverting the change is important, as at this moment we have a stable version with a big data-corrupting bug.

If possible, I would prefer to push this as is and then you can try to adapt it later.

hengsin (approved):
nvm

PR #1139, opened 2022-01-19
IDEMPIERE-2853: Fixing bug of usage of isSortTab flag on display logic

dpansheriya:


CarlosRuiz-globalqss (approved):
Tested and added also fix for DisplayLogic in Tab Editor button

PR #1140, opened 2022-01-19
IDEMPIERE-3044 : Add IsHighVolume on Tab

nmicoud:
Hi,
Model classes have Release 10 (whereas I think it should be 9).
I think is because the branch is created against master branch.
If it needs to be changed, can you explain the steps or point me to a place in wiki (I didn't find it)
Thanks
Nicolas


CarlosRuiz-globalqss:
> Hi, Model classes have Release 10 (whereas I think it should be 9). I think is because the branch is created against master branch. If it needs to be changed, can you explain the steps or point me to a place in wiki (I didn't find it) Thanks Nicolas

It doesn't matter now, model classes from R9 and R10 are now the same.

Regards,

Carlos Ruiz

CarlosRuiz-globalqss (approved):
Tested

nmicoud:
I was not speaking about the 'structure', but just of those changes:
- *  @version Release 9
+ *  @version Release 10

I_AD_BroadcastMessage.java which was updated 3 days ago has @version Release 9
whereas I_AD_Tab (just updated) has @version Release 10

Shouldn't all those classes have the same release version number?

CarlosRuiz-globalqss:
> Shouldn't all those classes have the same release version number?

Is just a comment, not a big deal.

PR #1141, opened 2022-01-20
IDEMPIERE-5155 Inventory Document: Auto picking of storage record should ignore record with Serial

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5155


CarlosRuiz-globalqss (approved):
unit tests pass

PR #1143, opened 2022-01-25
IDEMPIERE-4931 Running jobs never stop

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4931


hengsin (approved):
looks good

PR #1144, opened 2022-01-26
IDEMPIERE-5063 Improve Unit Tests - MatchInv Tests

tony-tspc:
- Add MatchInv test using standard costing and check Invoice Price Variance 
- General improvements using assertEquals where appropriate and
  comparing actual values rather than just >0
- minor fixes to MProduct test


hengsin (approved):
looks good

PR #1147, opened 2022-01-27
IDEMPIERE-5171 GridTable duplicate WhereClause Fix

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5171
Mattermost Discussion: https://mattermost.idempiere.org/idempiere/pl/encj8awdzbgt5jtf8y4hfxz1ay

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -3012,7 +3012,6 @@ public class GridTable extends AbstractTableModel
 		close(false);
 		if (retainedWhere != null)
 		{
-			// String whereClause = m_whereClause;
 			if (m_whereClause != null && m_whereClause.trim().length() > 0)
 			{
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -3015,7 +3015,9 @@ public void dataRefreshAll(boolean fireStatusEvent, int rowToRetained)
 			// String whereClause = m_whereClause;
```
CarlosRuiz-globalqss, line 3015: Hi @igorpojzl - in my suggested patch I removed the two comment lines 3015 and 3023.

I think is better to remove those lines as they are misleading and inserting a problem when the comment is removed.

Regards,

Carlos Ruiz

igorpojzl: Commited Changes

CarlosRuiz-globalqss (approved):

PR #1149, opened 2022-01-27
IDEMPIERE-1104 : SysConfig for posting using bank statement line date

nmicoud:


org.adempiere.base/src/org/compiere/model/MBankStatementLine.java
```diff
@@ -269,5 +278,19 @@ protected boolean updateHeader()
...
+
+	/**
+	 * Test if the line's date is on the same period as header's, only if BANK_STATEMENT_POST_WITH_DATE_FROM_LINE SysConfig is set to Y 
+	 */
+	public boolean checkDateIfUsedForPosting() {
+		if (MBankStatement.isPostWithDateFromLine(getAD_Client_ID())) {
+			MPeriod headerPeriod = MPeriod.get(getCtx(), getParent().getDateAcct(), getParent().getAD_Org_ID(), get_TrxName());
+			MPeriod linePeriod = MPeriod.get(getCtx(), getDateAcct(), getParent().getAD_Org_ID(), get_TrxName());
+
+			return headerPeriod != null && linePeriod != null && headerPeriod == linePeriod;
```
hengsin, line 291: should use ID for equal comparison instead.

nmicoud: done

org.adempiere.base/src/org/compiere/model/MBankStatementLine.java
```diff
@@ -269,5 +278,19 @@ protected boolean updateHeader()
 		}
 		return true;
 	}	//	updateHeader
+
+
+	/**
+	 * Test if the line's date is on the same period as header's, only if BANK_STATEMENT_POST_WITH_DATE_FROM_LINE SysConfig is set to Y 
+	 */
+	public boolean checkDateIfUsedForPosting() {
```
CarlosRuiz-globalqss, line 286: @nmicoud - the patch looks fine (I still haven't tested), but I find difficult to understand the name and javadoc of this method.

If I'm reading right it is returning true when the header/line periods match, and when is not validated.
And it returns false when the periods don't match.

If that's the case would be better to explain that in the javadoc, and maybe the name of the method would be better isValidPeriodInLine (or something like that)

I mean, if the method returns a boolean is better to name it using the "is" at the beginning.

"check" is a verb that doesn't tell what's happening inside, this name is probably used when the method inside throws an Exception.

nmicoud: hi @CarlosRuiz-globalqss 
you're right, what about isDateConsistentIfUsedForPosting ?
and "If the posting is based on the date of the line (ie SysConfig BANK_STATEMENT_POST_WITH_DATE_FROM_LINE = Y), make sure line and header dates are on the same period" for javadoc ?
I think is important to highlight the fact that it only apply when the SysConfig is activated.
wdyt?

CarlosRuiz-globalqss: Yes- sounds good

PR #1150, opened 2022-01-28
IDEMPIERE-5173 fix Accounting fact quantity incorrect

tony-tspc:
- also add unit tests
https://idempiere.atlassian.net/browse/IDEMPIERE-5173

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
+++ b/org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
@@ -44,6 +44,7 @@ import org.compiere.model.MInvoice;
 import org.compiere.model.MInvoiceLine;
 import org.compiere.model.MMatchInv;
 import org.compiere.model.MOrderLandedCostAllocation;
+import org.compiere.model.MUOM;
 import org.compiere.model.ProductCost;
 import org.compiere.model.Query;
 import org.compiere.model.X_M_Cost;
@@ -295,7 +296,12 @@ public class Doc_MatchInv extends Doc
 			}
 			else
 			{
-				cr.setQty(getQty().multiply(multiplier).negate());
+				if (m_invoiceLine.getC_UOM_ID() != 0)
+				{
+					int precision = MUOM.getPrecision(getCtx(), m_invoiceLine.getC_UOM_ID());
+					cr.setQty(getQty().multiply(multiplier).negate().setScale(precision, RoundingMode.HALF_UP));
+				} else
+					cr.setQty(getQty().multiply(multiplier).negate().stripTrailingZeros());
 			}
 		}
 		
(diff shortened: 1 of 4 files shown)


org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
```diff
@@ -295,7 +296,12 @@ public BigDecimal getBalance()
 			}
 			else
 			{
-				cr.setQty(getQty().multiply(multiplier).negate());
+				if (m_invoiceLine.getC_UOM_ID() != 0)
+				{
+					int precision = MUOM.getPrecision(getCtx(), m_invoiceLine.getC_UOM_ID());
+					cr.setQty(getQty().multiply(multiplier).negate().setScale(precision, RoundingMode.HALF_UP));
+				} else
```
hengsin, line 303: hmm .. isn't it not possible to have invoice line with C_UOM_ID == 0 ?

org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
```diff
@@ -339,20 +345,17 @@ public BigDecimal getBalance()
...
-		if (m_matchInv.getReversal_ID() == 0) 
+		cr.setC_Activity_ID(m_invoiceLine.getC_Activity_ID());
+		cr.setC_Campaign_ID(m_invoiceLine.getC_Campaign_ID());
+		cr.setC_Project_ID(m_invoiceLine.getC_Project_ID());
+		cr.setC_ProjectPhase_ID(m_invoiceLine.getC_ProjectPhase_ID());
+		cr.setC_ProjectTask_ID(m_invoiceLine.getC_ProjectTask_ID());
+		cr.setC_UOM_ID(m_invoiceLine.getC_UOM_ID());
+		cr.setUser1_ID(m_invoiceLine.getUser1_ID());
+		cr.setUser2_ID(m_invoiceLine.getUser2_ID());
+		if (m_matchInv.getReversal_ID() > 0)
```
hengsin, line 356: can you check whether m_matchInv.isReversal() works here ? If yes, that's the prefer approach.

org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
tony-tspc: Yes, as a match invoice is only created for products, there must be a uom set.
I will remove the 4 checks in this file.

org.adempiere.base/src/org/compiere/acct/Doc_MatchInv.java
tony-tspc: Yes, using m_matchInv.isReversal() works here. 
There are many occurences of "if (m_matchInv.getReversal_ID() == 0) " and "if (m_matchInv.getReversal_ID() > 0)" in this file. Did you mean for all of them to be changed?

hengsin: Yes, I think we should change those as well.

tony-tspc: OK, will do the changes and test over the weekend.

tony-tspc: I have made the requested changes and all unit tests are passing.

tony-tspc: @hengsin,  After further testing it appears that the new method MMatchInv.isReversal() is not working as expected. With a Credit Memo the match qty = -1 and the related m_inout movement qty = +1, so when you reverse the credit memo the original match invoice returns true for isReversal(), not the newly created match invoice. Below is a possible solution, which is similar to how the existing code in Doc_MatchInv determines the true reversal in several places.

```
	public boolean isReversal() {
		if (getReversal_ID() > 0) {
			MMatchInv reversal = new MMatchInv (getCtx(), getReversal_ID(), get_TrxName());
			if (reversal.getM_MatchInv_ID() < getM_MatchInv_ID())
				return true;
		}
		return false;
	}
```
Using this method all existing tests still pass, plus a new test reversing a credit memo.

hengsin: hi @tony-tspc , yes, that looks more reliable. please add that in as well, thanks.

tony-tspc: hi @hengsin , all done and added a new test for isReversal() when reversing a credit memo.
Thanks, Tony

hengsin (approved):
unit test pass

PR #1151, opened 2022-01-28
IDEMPIERE-5049 Zk Session and Desktop object not destroy immediately …

hengsin:
…after logout

Fix "zkoss.removeDesktop:70: Removing non-existent desktop: [Desktop z_7zm:/index.zul]" warning at server log (from curl/wget request with X-PING http header).


CarlosRuiz-globalqss (approved):
looks OK

PR #1152, opened 2022-01-30
IDEMPIERE-5175: on GL Journal Batch window, DateAct on line defaulted…

dpansheriya:
… from GL Journal Batch instead of Journal


CarlosRuiz-globalqss (approved):
tested

PR #1153, opened 2022-01-30
IDEMPIERE-5155 fix intermittant Inventory Test failure by sorting MSt…

tony-tspc:
…orageOnHand return value


hengsin (approved):
looks good

PR #1154, opened 2022-01-30
IDEMPIERE-5168 column AD_AuthorizationAccount.AccessToken is short on some case

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5168

- Allow reference Text with length 0 to be managed as CLOB/TEXT
- Increase AD_ChangeLog.OldValue/NewValue to 4000 to allow saving more information about changing long texts
- Change AD_AuthorizationAccount.AccessToken to CLOB/TEXT
- Change some important actual columns defined as Text Long, to Text with length 0 (this allows changelog, better 2Pack, encryption)

### DIFF
--- a/migration/i9/oracle/202201302336_IDEMPIERE-5168.sql
+++ b/migration/i9/oracle/202201302336_IDEMPIERE-5168.sql
@@ -1,0 +1,49 @@
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-5168 column AD_AuthorizationAccount.AccessToken is short on some case
+-- Jan 30, 2022, 11:18:41 PM CET
+UPDATE AD_Column SET FieldLength=4000,Updated=TO_DATE('2022-01-30 23:18:41','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=8805
+;
+
+-- Jan 30, 2022, 11:18:43 PM CET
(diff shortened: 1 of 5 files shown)


hieplq:
it look ok but "Access Token" field on gird mode become not secure anymore

image

CarlosRuiz-globalqss:
> it look ok but "Access Token" field on gird mode become not secure anymore
> 

Hi @hieplq - sorry, don't understand what you are reporting.

The Encryption defined in Column just happen on database, the UI still shows the values to the user.

The Encryption defined in Field just happen on UI, masking the value with ********

hieplq:
> > it look ok but "Access Token" field on gird mode become not secure anymore
> 
> Hi @hieplq - sorry, don't understand what you are reporting.
> 
> The Encryption defined in Column just happen on database, the UI still shows the values to the user.
> 
> The Encryption defined in Field just happen on UI, masking the value with ********

"Access Token" is same as "Refresh Token" both is encryption field also encryption column

i found logic make that issue

by bellow logic then text long and text secure only on gird model

image

CarlosRuiz-globalqss:
> "Access Token" is same as "Refresh Token" both is encryption field also encryption column
> i found logic make that issue
> by bellow logic then text long and text secure only on gird model

Thanks @hieplq - I committed a fix for that

hieplq (approved):

PR #1157, opened 2022-02-04
IDEMPIERE-5179 : Search fields : use reference table to determine whi…

nmicoud:
…ch image to show

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
@@ -192,7 +192,11 @@ public class WSearchEditor extends WEditor implements ContextMenuListener, Value
 	{
 
 		columnName = this.getColumnName();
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
```diff
@@ -192,15 +192,19 @@ private void init()
...
+		if (m_tableName.equals("C_BPartner"))
 		{
 			popupMenu = new WEditorPopupMenu(true, true, isShowPreference(), true, true, false, lookup);
 			if (ThemeManager.isUseFontIconForImage())
 				imageUrl = "z-icon-BPartner";
 			else
 				imageUrl = ThemeManager.getThemeResource("images/BPartner16.png");
 		}
-		else if (columnName.equals("M_Product_ID"))
+		else if (m_tableName.equals("M_Product"))
```
hengsin, line 207: shouldn't we check both m_tableName and m_keyColumnName here ?

hengsin (approved):
looks good

PR #1161, opened 2022-02-05
added static PO.shallowCopy() function 

Soundbytes:
It facilitates a quasi downcast of any PO implementation into a child class.


hengsin:
this is not safe as it ignore the fact that "T base" can have additional instance variables.

PR #1165, opened 2022-02-07
IDEMPIERE-5168 column AD_AuthorizationAccount.AccessToken is short on some case

CarlosRuiz-globalqss:
Fix ORA-22858: invalid alteration of datatype


CarlosRuiz-globalqss:
fix jenkins daily build broken

PR #1168, opened 2022-02-07
IDEMPIERE-4297 Improve test data / fixes for Garden World data

tony-tspc:
- Intercompany GL Journal Line has wrong Org
- Replace old ADempiere images & URL's with iDempiere


hengsin (approved):
looks good

PR #1170, opened 2022-02-07

CarlosRuiz-globalqss:

Database Error when query/ReQuery in info window "Info System Element" as System User


hengsin (approved):
looks good

PR #1171, opened 2022-02-07
IDEMPIERE-4863 Enable save column width in WListBox

CarlosRuiz-globalqss:
Do not use centralized ID for AD_USERPREFERENCE_ID and AD_WLISTBOX_CUSTOMIZATION_ID


hengsin (approved):
looks good

PR #1172, opened 2022-02-08

CarlosRuiz-globalqss:

 - change approach to replace System user instead of removing it

### DIFF
--- a/migration/iD10/oracle/202202072157_IDEMPIERE-5174.sql
+++ b/migration/iD10/oracle/202202072157_IDEMPIERE-5174.sql
@@ -2,33 +2,93 @@ SET SQLBLANKLINES ON
 SET DEFINE OFF
 
 -- IDEMPIERE-5174 Deactivate System user
+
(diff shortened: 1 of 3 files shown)


hengsin (approved):
looks good

nmicoud:
seems ok (oracle) ; but there is I think a missing piece here:  what about customized tables/columns?
is it up to each implementor to update them?

CarlosRuiz-globalqss:
Yes - I thought the same about AD_User custom columns - but ended thinking that yes - is an implementor task as well as new tables.  But, if you can do a script that do the work for oracle and postgresql is fine, I'll test it.

nmicoud:
After writing the comment, I tried to think how to make a generic script that will work in every instances but is too complex (ie: sometimes I can use CreatedBy to determine to who belongs the record, sometimes no ; so should I consider that column?)

In the end, I think Implementor task is good :)
A note in the wiki on migration is enough and then is up to every implementor to think about what changes need to be done.
wdyt?

hengsin:
I think a note in the wiki should be sufficient. In practice, the usage of System user should be very very low (I've never use it).

CarlosRuiz-globalqss:
> I think a note in the wiki should be sufficient. In practice, the usage of System user should be very very low (I've never use it).

https://wiki.idempiere.org/en/Migration_Notes#System_user_with_ID_zero_deprecated

PR #1174, opened 2022-02-09

CarlosRuiz-globalqss:

Verify hardcoded comparisons for role and system users


hengsin (approved):
looks good

PR #1175, opened 2022-02-09
IDEMPIERE-5184 Cost Adjustment sets incorrect cost for other schema …

tony-tspc:
…when currency blank on header
- also add check to unit test


tony-tspc:
Thought it was worth also enforcing the currency_id in m_inventory before save, for cost adjustment documents

hengsin (approved):
looks good

PR #1177, opened 2022-02-11
IDEMPIERE-5188 : Reset content of color field

nmicoud:


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
```diff
@@ -312,6 +316,26 @@ else if (WEditorPopupMenu.REQUERY_EVENT.equals(evt.getContextEvent()))
...
+				adwindowContent.getADTab().getSelectedTabpanel().switchRowPresentation();
+
+			GenericPO gpo = new GenericPO(gridTab.getTableName(), Env.getCtx(), gridTab.getRecord_ID());
+			gpo.set_ValueNoCheck(gridField.getColumnName(), null);
+			gpo.saveEx();
+
+			gridTab.dataRefresh();
+
+			if (needSwitch)
+				adwindowContent.getADTab().getSelectedTabpanel().switchRowPresentation();
```
hengsin, line 337: this is not right, should use the editor to reset value to null or empty.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
nmicoud: I agree, but using setValue (null) and saving record doesn't empty the field, but set it to the "default" value ("#000000").
And I can't find where this is set.
The field is "created" on DefaultEditorFactory.getEditor at line 100 ; but I can't find any clue there.
Any idea ?

nmicoud: done

CarlosRuiz-globalqss:
Reading the link that @hengsin shared, I think is maybe possible another solution:

- set the field to be type text if empty, and color if filled (see DefaultEditorFactory:99-101)
- when empty set the type color when the field is clicked
- the "Reset" option added by @nmicoud could set the field back to type text and clear it

haven't tested, but it sounds like a potential solution without creating a new editor

nmicoud:
So maybe the best way is to have a dedicated editor, that should be be easier to work on UX.

I have a window where user can choose several colors for an external service.
The service has default colors and they can be overwritten within iDempiere.
So user can define a color, but is not possible to return to the default value.

CarlosRuiz-globalqss:
Understood, and agree - a dedicated editor sounds the best option probably.

nmicoud:
Hi,

I've added a WColorEditor and ColorBox classes (duplicated from *Filename*) which are used when the field is Color.
It should be considered as a WIP commit (I know there are TODO, System.out.println, ...)

I think is better to display the color instead of its html code (from a UX point of view).
That's why I would like to be able to open the color picker from the button instead of the 'content' of the textbox.
But I don't know how to do that, I cant't find the code which is opening it (and I fear it is inside the textbox component itself).

For the 'reset' part, I'm trying to do something like :
 - change field type from 'color' to 'text'
 - setValue(null)
The content will be empty - and I think a placeholder can be displayed (something like "click on the button to choose a color").
But is not working. Can you see a way to do this ?
Otherwise, I see no other option but to do what Hengsin suggest (change type from color to text ; but I still need to know how to open the color picker)

BTW, while making test and search, I make a change so color can be seen in grid mode (see changes in GridTabRowRenderer).
But ATM, is not possible to interract with the editor.

Thanks

Nicolas

nmicoud:
Thanks, is working !
It can be opened from the field itself (click on the color), the button or the popup.
We probably don't need all of them, but it will be easier to remove unnecessary ones.
nb: room for improvement for that part -> the color picker opens at upper left, would be better if centered, but not a big deal.

I also commit a change to reset the content, based on @CarlosRuiz-globalqss's suggestion.
Seems to work. If you're ok with that approach, we can say that 'technically', is working.
But from UX perspective, is not good as you can't see that the value in DB is null. I try to make changes in the getDisplay() method but not sure is the correct place to do this.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ColorBox.java
```diff
@@ -0,0 +1,32 @@
+/******************************************************************************
+ * Copyright (C) 2008 Low Heng Sin                                            *
```
hengsin, line 2: not sure where you copy this from but it should be remove

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ColorBox.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program; if not, write to the Free Software Foundation, Inc.,    *
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.                     *
+ *****************************************************************************/
+package org.adempiere.webui.component;
+
+/**
+ * 
+ * @author Low Heng Sin
```
hengsin, line 17: well, I'm not the author of this class :)

hengsin:
Hi @nmicoud,
This is not good from the UI perspective and it is fragile since when the db value is null, the text input element actually will have the value of "#000000" (you can check that through Chrome Dev Tool's inspect element).

Other than using a popup dialog, perhaps you can try the following approach too:
- keep the actual Text box as "text" type. Make it readonly and display using the selected color. You can show a place holder text when the field is null (this shouldn't override the gridfield.getPlaceHolder() feature though so take note of that).
- add another input text that's of type "color". use absolute positioning to place it at the beginning of the component and make it invisible to user (try width and height of 0 or display:none or visibility: hidden)

Regards,
Low

hengsin:
[1177_pr.patch.txt](https://github.com/idempiere/idempiere/files/8122623/1177_pr.patch.txt)

Patch for the hidden color text field approach.

Note: have not been able to figure out why the color type text field doesn't works in grid mode, seems like a zk issue.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ColorBox.java
nmicoud: Duplicated from FilenameBox

nmicoud:
Hi @hengsin,

Thanks a lot for you patch, it works like a charm !
I've commited it and I will start to review header/javadoc, add messages and search for icons.

Is there any interest to keep Colorbox.java?
I also think that changes done in GridTabRowRenderer should be removed too. 

wdyt?

Regards,

Nicolas

hengsin:
Hi @nmicoud ,
- yes, the GridTabRowRenderer changes should be dropped.
- I think you can keep ColorBox and move some of the code from WColorEditor to ColorBox.

nmicoud:
Hi @hengsin,
Not sure what code should be moved from WColorEditor to ColorBox, so I let it as it. If you think changes are required, can you guide me ?
I put migration scripts on i9 ; if you prefer on ID10, I'll change that.

Thanks a lot !
Regards,
Nicolas

CarlosRuiz-globalqss:
> I put migration scripts on i9 ; if you prefer on ID10, I'll change that.

i9 is OK if you need it there.

hengsin:
Hi @nmicoud , in this case, I think it is better to just drop the ColorBox class, making it simpler to understand.

nmicoud:
hi @hengsin,
ColorBox.java has been removed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WColorEditor.java
```diff
@@ -0,0 +1,268 @@
...
+		else if (Events.ON_CHANGE.equals(event.getName()) || Events.ON_OK.equals(event.getName()))
+		{
+			String newValue = getComponent().getTextbox().getValue();
+			processNewValue(newValue);
+		}
+		else
+		{
+			return;
+		}
+	}
```
hengsin, line 229: I think both event is no longer need here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditorPopupMenu.java
```diff
@@ -62,6 +62,8 @@
     public static final String SHOWLOCATION_EVENT = "SHOW_LOCATION";
     public static final String CHANGE_LOG_EVENT = "CHANGE_LOG";
     public static final String EDITOR_EVENT = "EDITOR";
+    public static final String RESET_EVENT = "RESET";
+    public static final String COLOR_PICKER_EVENT = "COLOR_PICKER";
```
hengsin, line 66: COLOR_PICKER_EVENT constant should be moved to the WColorEditor class

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WColorEditor.java
```diff
@@ -0,0 +1,268 @@
...
+import org.zkoss.zul.Menuitem;
+
+/**
+ *
+ * @author Nicolas Micoud (TGI)
+ *
+ */
+public class WColorEditor extends WEditor implements ContextMenuListener
+{
+	private static final String[] LISTENER_EVENTS = {Events.ON_CLICK, Events.ON_CHANGE, Events.ON_OK};
```
hengsin, line 57: I think on ON_CLICK is needed here

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WColorEditor.java
nmicoud: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditorPopupMenu.java
nmicoud: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WColorEditor.java
nmicoud: done

hengsin (approved):
tested

PR #1188, opened 2022-02-16
IDEMPIERE-5191 SEARCH FAILURE BY DATE AND HOUR

CarlosRuiz-globalqss:
IDEMPIERE-4962 IDEMPIERE-4724

https://idempiere.atlassian.net/browse/IDEMPIERE-5191


hengsin:
It is rather unfortunate that a lot of our issue is due to the use of the old and non standard Oracle Date datatype and the use of PostgreSQL Timestamp (instead of Timestamp with time zone) datatype to have better compatibility with Oracle Date.

hengsin (approved):
looks good

PR #1191, opened 2022-02-17
IDEMPIERE-5195 SEARCH FAILURE FOR SAVED QUERIES + IDEMPIERE-4658 #resolve

CarlosRuiz-globalqss:
- Fix issue reported in IDEMPIERE-5195 saving wrongly Yes/No values as true/false
- Fix also wrong recover of timestamps just as dates (ignoring the time part)
- Fix also IDEMPIERE-4568 reopened, adding wrongly operators

https://idempiere.atlassian.net/browse/IDEMPIERE-5195


hengsin (approved):
looks good

PR #1193, opened 2022-02-18
IDEMPIERE-3831 - Implement AD_FieldStyle for Info Windows

PeterTakacs300:
Issue: CSS style replaces the entire style, instead of appending.

Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-3831

Use case: Product Info: OnHandQty @QtyOnHand@>5 then color: green.

In this case 'replacelD' style was removed, ensure number right alignment,  just the color was added. So cells with the colors were aligned left, w/o right (legacy)

solution: append style instead replace.


PeterTakacs300:
Created from wrong branch..

PR #1194, opened 2022-02-18
IDEMPIERE-3831 - css style replaces the entire style, instead of append

PeterTakacs300:
Issue: CSS style replaces the entire style, instead of appending.

Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-3831

Use case: Product Info: OnHandQty @QtyOnHand@>5 then color: green.

In this case 'replacelD' style was removed, ensure number right alignment, just the color was added. So cells with the colors were aligned left, w/o right (legacy)

solution: append style instead replace.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/WInfoWindowListItemRenderer.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/WInfoWindowListItemRenderer.java
@@ -37,7 +37,6 @@ import org.compiere.model.MStyle;
 import org.compiere.util.Env;
 import org.compiere.util.Evaluatee;
 import org.compiere.util.KeyNamePair;
-import org.compiere.util.Util;
 import org.zkoss.zul.Listcell;
 
 public class WInfoWindowListItemRenderer extends WListItemRenderer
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/WInfoWindowListItemRenderer.java
```diff
@@ -156,7 +156,7 @@ public String get_ValueAsString(String variableName) {
 					String zclass = styleStr.substring(MStyle.ZCLASS_PREFIX.length());
 					listcell.setZclass(zclass);
 				} else {
-					listcell.setStyle(styleStr);
+					listcell.setStyle(listcell.getStyle() + ";" + styleStr);
```
hengsin, line 159: need to check whether listcell.getStyle() is null or empty.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/WInfoWindowListItemRenderer.java
```diff
@@ -156,7 +157,15 @@ public String get_ValueAsString(String variableName) {
...
+										
+					if(!Util.isEmpty(prevStyle)) {
+						listcell.setStyle(prevStyle + ";" + styleStr);
+					}
+					else {
+						listcell.setStyle(styleStr);
+					}
 				}
 			}
 		}
```
hengsin, line 171: Hi, just recall that we has an existing helper method for this - ZkCssHelper.appendStyle(component, style). Should use that instead.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/WInfoWindowListItemRenderer.java
PeterTakacs300: Hello, thank you, I will check it out.

hengsin (approved):
looks good

PR #1195, opened 2022-02-18
IDEMPIERE-5197 Add target platform file for org.idempiere.p2/target/r…

hengsin:
…epository

https://idempiere.atlassian.net/browse/IDEMPIERE-5197


CarlosRuiz-globalqss (approved):
looks good

PR #1196, opened 2022-02-21
IDEMPIERE-5198 Print Format Cache do not reload Print Format Items

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5198

Force Reload Items when loading Print Format From cache, applying Role Access.


CarlosRuiz-globalqss (changes requested):
Hi @igorpojzl - I think reloading the items is wrong, haven't tested but I can imagine a concurrency problem when two different roles are accessing the print format and reloading the items at the same time.

Instead of this, I suggest the pull request https://github.com/idempiere/idempiere/pull/1197 - but I haven't tested it, can you please test and check if that solves your problem?

Regards,

Carlos Ruiz

hengsin:
I would add that if the item is not cache, the print format cache is of not much use here since the item cache is the significant part of it.

CarlosRuiz-globalqss:
Closing this in favour of https://github.com/idempiere/idempiere/pull/1197

PR #1197, opened 2022-02-21
IDEMPIERE-5198 Print Format Cache do not reload Print Format Items

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5198

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/MPrintFormat.java
+++ b/org.adempiere.base/src/org/compiere/print/MPrintFormat.java
@@ -1153,7 +1153,43 @@ public class MPrintFormat extends X_AD_PrintFormat implements ImmutablePOSupport
 	}
 
 	/** Cached Formats						*/
-	static private ImmutablePOCache<String,MPrintFormat> s_formats = new ImmutablePOCache<String,MPrintFormat>(Table_Name, 30);
+	static private ImmutablePOCache<String,MPrintFormat> s_formats = new ImmutablePOCache<String,MPrintFormat>(Table_Name, 30) {
+		private static final long serialVersionUID = 2428566381289874703L;
+
+		@Override
+		public int reset(int recordId) {
+			if (recordId <= 0)
+				return reset();
+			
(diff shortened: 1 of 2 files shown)


hengsin:
It would be better if this can override the reset(int id) method and clear the corresponding AD_PrintFormat_ID cache entries when the record is change.

igorpojzl (approved):
Tested with 2 roles. No MissMatch Report Rendered.

CarlosRuiz-globalqss:
> It would be better if this can override the reset(int id) method and clear the corresponding AD_PrintFormat_ID cache entries when the record is change.

Hi @hengsin - I didn't understand the suggestion.  The change proposed here (and successfully tested by @igorpojzl) is simply to change the key for the print format, that will use a bit more memory I guess, but is safe as the cached data depends on the role, in general the cache keys must reflect the variables that can potentially make the cache record different, that has been in past a common error for example caching translated things without language in the key.

hengsin:
Hi @CarlosRuiz-globalqss , if reset(int id) is not override, the update of any print format record will trigger the reset/clean up of the whole print format cache. It is not a must but it is less efficient without that.

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss , if reset(int id) is not override, the update of any print format record will trigger the reset/clean up of the whole print format cache. It is not a must but it is less efficient without that.

Understood, can you please provide a patch, I don't know how to do that, and didn't find any example of "reset(int id)" in code

Maybe the only one applicable is in MWorkflow line 191, but the implementation there seems to reset the whole cache when calling to reset one id.

hengsin:
[1197_pr.patch.txt](https://github.com/idempiere/idempiere/files/8123451/1197_pr.patch.txt)

Hi @CarlosRuiz-globalqss, please check the attached patch is ok for you.

CarlosRuiz-globalqss:
Ready @hengsin - tested and integrated your patch.

hengsin (approved):
tested by carlos and igor

PR #1201, opened 2022-02-23

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5202


CarlosRuiz-globalqss:
Hi @hengsin - testing this, two things noticed:

1 - in osgi console I execute this:
```setprop org.idempiere.db.debug=true```
and then editing a text field is generating a lot of queries going to the database on every keystroke
from what I see in console every keystroke is generating one or two updates to AD_RecentItem and 20 queries to AD_UserQuery, AD_Attachment_ID, AD_PostIt_ID, AD_RecentItem_ID, CM_Chat_ID, AD_RecentItem and the table used for the test (in my case M_FreightCategory)

2 - on every keystroke the save button is disabled (as if the record was saved), but in reality the record has not been saved in the database until the text field lose the focus.
Two consequences noticeable:
- I change the Description field, but I don't lose the focus
- if I close the iDempiere Tab - then a message appears "There are unsaved changes - are you sure to close?" - but in reality the record was already saved because the text field lost the focus, so the message is unnecessary in this case
- if I close the browser Tab - then a message appears confirming to abandon the page - if I confirm, the record is not saved in database (although the save button was disabled)
I think the save button must remain enabled until the record is really saved.

3 - Another consequence of this approach for the database, is that the UPDATEs are created on every field change, with the Save button, one update can change several fields at once, with the auto-save approach every field is saved when losing focus.  What means more traffic to the database in most cases.

Based on this last comment, maybe it makes sense to implement this as an option useful for some windows, but not in general for all windows, something like a flag configurable in AD_Window, and maybe the SysConfig determines the default behavior.    WDYT?

hengsin:
I will looks at other issues. For 3, it is obvious this is about trading some performance for better user experience. From the UI aspect, it would be confusing that some window does auto save and some doesn't.

hengsin:
Hi @CarlosRuiz-globalqss , this is ready for review again.

PR #1203, opened 2022-02-24
IDEMPIERE-5205 PO is Immutable: org.compiere.model.MCtxHelpMsg

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5205


hengsin (approved):
trivial fix, good to go

PR #1204, opened 2022-02-24
IDEMPIERE-4706 New field "Dynamic Validation (Lookup)" in AD_Column>AD_Val_Rule_Lookup_ID not working

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4706

### DIFF
--- a/migration/i9/oracle/202202241712_IDEMPIERE-4706.sql
+++ b/migration/i9/oracle/202202241712_IDEMPIERE-4706.sql
@@ -1,0 +1,18 @@
+-- IDEMPIERE-4706 New field "Dynamic Validation (Lookup)" in AD_Column>AD_Val_Rule_Lookup_ID not working
+SELECT register_migration_script('202202241712_IDEMPIERE-4706.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Feb 24, 2022, 5:12:26 PM CET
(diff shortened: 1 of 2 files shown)


hengsin:
Hi @CarlosRuiz-globalqss , this confuse me a bit. It seems the issue describe in the ticket need more than just AD change or did I misunderstood it ?

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss , this confuse me a bit. It seems the issue describe in the ticket need more than just AD change or did I misunderstood it ?

Hi @hengsin, sorry I didn't explain better.

1 - In the [forum thread](https://groups.google.com/g/idempiere/c/6sJN7223DrA/m/XoZOeJlDBQAJ), Michael described the following:
> dynamic validation is separated into document field validation and lookup field validation. 
> We were assuming that recording the same validation in both field should yield the same result as before, but it seems it doesn't:
> E.g. C_Order>m_warehouse_id>DynVal=M_Warehouse Org
> This dynamic validation ensures that only warehouses belonging to the @AD_Org_ID@ can be selected. It works as expected > in the document field validation but no longer under lookup validation

I did the tests and I see it's working, but precisely the purpose of the lookup validation is the opposite, for example, in orders, the warehouse needs to be filtered by organization when being captured, but in the Find dialog the warehouse must show all, not related to the org.

So, I created a reference validation with name "All Records" - and validation code is `1=1`, assigned this as lookup validation for the warehouse and it works as expected, filtered in the window, not-filtered in the lookup.

I just set the "All Records" to warehouse here, but I think it's useful for any field that is intended to show all records, not filtered by context, just I don't have the time to check which fields require that - must be a lot.

2 - Additionally on the ticket I wrote
> Also I noticed that in "Validation Rules" window, the "Used in Column (Lookup)" is not showing this relationship, and also is not discovered in the "Zoom Across" toolbar button.

The last update setting the Parent_Column_ID for the AD_Tab record did the trick to solve this one.

Regards,

Carlos Ruiz

hengsin (approved):
tested

PR #1206, opened 2022-02-28
IDEMPIERE-5206 - HTML Report CSS style - add AD_FieldStyle_ID on AD_PrintFormatItem table

PeterTakacs300:
Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5206

Try to add AD_FieldStyle_ID on AD_PrintFormatItem table.

Improvement: you will be able to change css style of data in single cells of HTML report, based on the Display Logic defined in AD_FieldStyle.

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/ReportEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/ReportEngine.java
@@ -931,25 +931,27 @@ queued-job-count = 0  (class javax.print.attribute.standard.QueuedJobCount)
 							Object obj = instanceAttributeColumn != null ? instanceAttributeColumn.getPrintDataElement(row)
 									: m_printData.getNodeByPrintFormatItemId(item.getAD_PrintFormatItem_ID());
 							//set style
-                            MStyle style = MStyle.get(Env.getCtx(), item.getAD_FieldStyle_ID());
-                            X_AD_StyleLine[] lines = style.getStyleLines();
-                            StringBuilder styleBuilder = new StringBuilder();
-                            for (X_AD_StyleLine line : lines)
-                            {
-                                
-                                String inlineStyle = line.getInlineStyle().trim();
-                                String displayLogic = line.getDisplayLogic();
-                                if (!Util.isEmpty(displayLogic))
-                                {
-                                    if (!Evaluator.evaluateLogic(new PrintDataEvaluatee(null, m_printData), displayLogic))
-                                        continue;
-                                }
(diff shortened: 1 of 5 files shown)


org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -928,6 +930,27 @@ else if (colObj instanceof InstanceAttributeColumn)
 							tr.addElement(td);
 							Object obj = instanceAttributeColumn != null ? instanceAttributeColumn.getPrintDataElement(row)
 									: m_printData.getNodeByPrintFormatItemId(item.getAD_PrintFormatItem_ID());
+							//set style
+                            MStyle style = MStyle.get(Env.getCtx(), item.getAD_FieldStyle_ID());
```
hengsin, line 934: need to check item.getAD_FieldStyle_ID() > 0 here since it is optional

org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -928,6 +930,27 @@ else if (colObj instanceof InstanceAttributeColumn)
...
+                                if (!Util.isEmpty(displayLogic))
+                                {
+                                    if (!Evaluator.evaluateLogic(new PrintDataEvaluatee(null, m_printData), displayLogic))
+                                        continue;
+                                }
+                                if (styleBuilder.length() > 0 && !(styleBuilder.charAt(styleBuilder.length()-1)==';'))
+                                    styleBuilder.append("; ");
+                                styleBuilder.append(inlineStyle);
+                            }
+                            if(styleBuilder != null)
```
hengsin, line 951: styleBuilder can't be null here, should check styleBuilder.length() > 0 instead

org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -928,6 +930,29 @@ else if (colObj instanceof InstanceAttributeColumn)
 							tr.addElement(td);
 							Object obj = instanceAttributeColumn != null ? instanceAttributeColumn.getPrintDataElement(row)
 									: m_printData.getNodeByPrintFormatItemId(item.getAD_PrintFormatItem_ID());
+							//set style
```
hengsin, line 933: It looks strange to insert this block of code here. Should be before or after tr.addElement(td) above.

hengsin (changes requested):

hengsin (approved):
looks good

PR #1210, opened 2022-03-02
IDEMPIERE-5168 column AD_AuthorizationAccount.AccessToken is short on…

nmicoud:
… some case - fix oracle script when table has data


CarlosRuiz-globalqss (approved):
tested manually in oracle

PR #1215, opened 2022-03-03

hengsin:
- Fix auto save sometimes not happening for string field update

https://idempiere.atlassian.net/browse/IDEMPIERE-5202


CarlosRuiz-globalqss (approved):
tested successfully

PR #1217, opened 2022-03-03
IDEMPIERE-5216 - HTML Report - replace Record_ID with Identifier column

PeterTakacs300:
Jira issue: https://idempiere.atlassian.net/browse/IDEMPIERE-5216

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/ReportEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/ReportEngine.java
@@ -89,6 +89,7 @@ import org.compiere.model.MQuery;
 import org.compiere.model.MRfQResponse;
 import org.compiere.model.MRole;
 import org.compiere.model.MStyle;
+import org.compiere.model.MSysConfig;
 import org.compiere.model.MTable;
 import org.compiere.model.PrintInfo;
 import org.compiere.model.X_AD_StyleLine;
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -992,7 +992,23 @@ else if (obj instanceof PrintDataElement)
...
-												String foreignColumnName = mTable.getTableName() + "_ID";
+												String tableName = mTable.getTableName();
+												
+												ArrayList<MColumn> list = new ArrayList<MColumn>();
+												for (String idColumnName : mTable.getIdentifierColumns()) {
+													MColumn column = mTable.getColumn(idColumnName);
+													list.add (column);
+												}
+												if(list.size() > 0) {
+													StringBuilder displayColumn = new StringBuilder(list.get(0).getColumnName());
```
hengsin, line 1003: We should use value from MSysConfig.IDENTIFIER_SEPARATOR to combine the list of value from identifiers (use nvl to handle null). I've not tested but the ColumnLookup.getLookupDisplay approach might work here as well.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -992,7 +993,32 @@ else if (obj instanceof PrintDataElement)
...
+												}
+												if(list.size() > 0) {
+													StringBuilder displayColumn = new StringBuilder();
+													String separator = MSysConfig.getValue(MSysConfig.IDENTIFIER_SEPARATOR, "_", Env.getAD_Client_ID(Env.getCtx()));
+													
+													for(int i = 0; i < list.size(); i++) {
+														if(i > 0)
+															displayColumn.append("||'").append(separator).append("'||");
+														
+														displayColumn.append(list.get(i).getColumnName());
```
hengsin, line 1011: this wouldn't work if one of the identifier column is null, you need to use nvl to cater for that.

hengsin (approved):
looks good

PR #1219, opened 2022-03-03
IDEMPIERE-5218 Improvements for Record Info dialog

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5218


org.adempiere.ui.zk/WEB-INF/src/web/js/org/idempiere/commons/clipboard.js
```diff
@@ -0,0 +1,12 @@
+if (typeof window.idempiere === 'undefined')
+	window.idempiere = {};
+
+window.idempiere.copyToClipboard = function(prm){
+	var tempInput = document.createElement("input");
```
hengsin, line 5: the navigator.clipboard api is pretty well supported now and we should use that instead.
see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard and https://www.w3schools.com/howto/howto_js_copy_clipboard.asp

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordInfo.java
```diff
@@ -341,8 +345,37 @@ else if (dse.getSource() instanceof GridTable)
...
+						boolean addAnd = false;
+						for (String key : po.get_KeyColumns()) {
+							if (addAnd)
+								query.append(" AND ");
+							query.append(key)
+								.append("=")
+								.append(po.get_Value(key));
+							addAnd = true;
+						}
+						query.append("\");");
```
hengsin, line 374: shouldn't we just use PO.get_WhereClause instead ?

hengsin (changes requested):

hengsin (approved):
looks good

PR #1220, opened 2022-03-03
Idempiere 5214

nmicoud:


CarlosRuiz-globalqss:
@nmicoud - tested successfully.

I would think this patch is worthy to add:
[IDEMPIERE-5214_PR.txt](https://github.com/idempiere/idempiere/files/8178581/IDEMPIERE-5214_PR.txt)

To do the same in new tenants created.

nmicoud:
indeed ; done

CarlosRuiz-globalqss (approved):
tested successfully

PR #1222, opened 2022-03-03
IDEMPIERE-5219 Sales Order Window: initial focus not on header tab's …

hengsin:
…field

https://idempiere.atlassian.net/browse/IDEMPIERE-5219


CarlosRuiz-globalqss (approved):
tested successfully

PR #1223, opened 2022-03-04

hengsin:
- Table/List Editor: auto drop upon focus works better for auto save.
Without auto drop upon focus, auto save is trigger for every arrow
up/down key stroke.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WTableDirEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WTableDirEditor.java
@@ -207,6 +207,16 @@ ContextMenuListener, IZoomableEditor
         	getComponent().addScrollSelectedIntoViewListener();
         }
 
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I cannot reproduce the issue as you describe it in a vanilla iDempiere.

How can I check the effect?

hengsin:
Hi @CarlosRuiz-globalqss ,

With current code, after turning on auto save, navigate to a field that's either Table, Table Direct or List, save will be trigger everytime you change the selected value using up/down arrow key. I think that's too eager.

CarlosRuiz-globalqss:
Thanks @hengsin - testing it I notice there is a change also when navigating to this field the list is automatically opened.

Is that necessary?   It doesn't look good.

hengsin:
Yes, that's needed to delay the triggering of save. If you don't think it is good, you can close the PR and we just leave it as it is.

hengsin:
need further investigation

PR #1226, opened 2022-03-04

hengsin:
- Auto open the dropdown list upon pressing of up/down arrow key
- Do not auto move focus to next field


CarlosRuiz-globalqss (approved):
tested successfully

PR #1229, opened 2022-03-07
IDEMPIERE-5220 Duplicated indexes in postgresql (and mismatch against oracle) / FHCA-2740

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5220

Please note:
* I separated the tickets because the first one is intended to integrate in a 7.1 installation
* Oracle script differs from postgresql because in oracle those indexes are OK - oracle doesn't allow duplicated indexes


tony-tspc:
Tested in postgresql and confirmed removal of duplicate indexes in database and dictionary.

hengsin (approved):
tested by tony

PR #1232, opened 2022-03-08
IDEMPIERE-5063 Improve Unit Tests

tony-tspc:
https://idempiere.atlassian.net/browse/IDEMPIERE-5063
- Close DB connections after each unit test class
- POTest and QueryTest: use assert equals, use bigdecimal string constructor in preference to double

When running the Unit Tests, the number of DB connections continuously grows until all the tests are completed. Currently this is  over 80 connections. As the number of unit tests increase, this could become an issue. Closing the database after each test class greatly reduces the number of concurrent connections and doesn't add more than a couple of seconds to the total time taken to run all the tests.

Also added some minor improvevements to POTest and QueryTest.


hengsin (approved):
looks good

PR #1233, opened 2022-03-08
IDEMPIERE-4586 Update translation of Client to Tenant

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4586

### DIFF
--- a/migration/iD10/oracle/202203081457_IDEMPIERE-4586.sql
+++ b/migration/iD10/oracle/202203081457_IDEMPIERE-4586.sql
@@ -1,0 +1,283 @@
+UPDATE AD_Element SET Description=REPLACE(Description, 'client', 'tenant') WHERE AD_Element_ID<1000000 AND REGEXP_LIKE(Description,'(^|\W)client(\W|$)') AND Description NOT LIKE '%client-%';
+
+UPDATE AD_Element SET Description=REPLACE(Description, 'clients', 'tenants') WHERE AD_Element_ID<1000000 AND REGEXP_LIKE(Description,'(^|\W)clients(\W|$)');
+
+UPDATE AD_Element SET Description=REPLACE(Description, 'Client', 'Tenant') WHERE AD_Element_ID<1000000 AND REGEXP_LIKE(Description,'(^|\W)Client(\W|$)');
+
+UPDATE AD_Element SET Description=REPLACE(Description, 'Clients', 'Tenants') WHERE AD_Element_ID<1000000 AND REGEXP_LIKE(Description,'(^|\W)Clients(\W|$)');
+
+UPDATE AD_Element SET Help=REPLACE(Help, 'client', 'tenant') WHERE AD_Element_ID<1000000 AND REGEXP_LIKE(Help,'(^|\W)client(\W|$)') AND Help NOT LIKE '%client-%';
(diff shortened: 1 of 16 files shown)


nmicoud:
Hi @CarlosRuiz-globalqss,
I've just look at the change, not tested locally.

Some remarks:
- I see Placeholder is not updated. Is it intentional?
- Shouldn't we regenerate all I/X classes based on tables with AccessLevel using Client/Tenant to use the new 'label'?
- Shouldn't we regenerate all I to "fix" javadoc for public int getAD_Client_ID(); (Get Client => Get Tenant) - not a big deal as the 2nd line is "Client/**Tenant** for this installation.").

I'll make tests tomorrow

Regards,

Nicolas

CarlosRuiz-globalqss:
>     * I see Placeholder is not updated. Is it intentional?

I just included script for tables/columns that are affected in the seed database, not for empty columns, or columns not containing the "client" word.

For extending this to non-core tables or custom values, I think is very easy to just copy the script and include your own tables/columns in a custom migration script.

>     * Shouldn't we regenerate all I/X classes based on tables with AccessLevel using Client/Tenant to use the new 'label'?
>     * Shouldn't we regenerate all I to "fix" javadoc for public int getAD_Client_ID(); (Get Client => Get Tenant) - not a big deal as the 2nd line is "Client/**Tenant** for this installation.").

I think is not necessary as they are just comments.

But thanks to your comment I noticed the impact of changing AD_Ref_List.Name is too big, I reverted this one from the scripts.

hengsin:
Perhaps we should update the Updated column too so that it is easier to trace back which records have been updated.

What about those that maintain en as translation instead ?

nmicoud:
Hi,

I've tested the script on both Oracle and PG, seems ok.
I didn't find in code "Client" which should be shown to users - and as you said, is ok to let "Client" if is comment.

Just found some descriptions that needs to be updated :
- AD_Client_ID element description is "Tenant/Tenant for this installation."
- Tenant window description is "Maintain Tenants/Tenants"
- Tenant tab description is "Tenant/Tenant Definition"
- Table AD_Client description is "Client/Tenant for this installation." ; so maybe an update is also needed here (or wait for the next run of SyncTerminology process?)

Regards,

Nicolas

CarlosRuiz-globalqss:
@hengsin 
> Perhaps we should update the Updated column too so that it is easier to trace back which records have been updated.

Done - added a timestamp for the Updated column in all affected records.

> What about those that maintain en as translation instead ?

The script just cover what is in seed, for customizations is easy to use this script as a template and add more tables.
Also, the actual recommended setup doesn't require English as translation.


@nmicoud 
> Just found some descriptions that needs to be updated :

Thanks, added those, and one more found to the scripts.

nmicoud:
Hi @CarlosRuiz-globalqss,
One last... CommitWarning field on AD_Tab (eg: on Tenant window, there is still "Initial Client Setup".

CarlosRuiz-globalqss:
> One last... CommitWarning field on AD_Tab (eg: on Tenant window, there is still "Initial Client Setup".

Thanks @nmicoud - added

nmicoud (approved):
looks good

PR #1234, opened 2022-03-08
IDEMPIERE-1474 The class org.compiere.model.MPaymentValidate is not u…

laydasalasc:
https://idempiere.atlassian.net/browse/IDEMPIERE-1474


hengsin (approved):
looks good

PR #1235, opened 2022-03-09
IDEMPIERE-5094 Implement validation for costing level change ( Accoun…

hengsin:
…ting Schema and Product Category Accounting )

https://idempiere.atlassian.net/browse/IDEMPIERE-5094

### DIFF
--- a/migration/i9/oracle/202203090311_IDEMPIERE-5094.sql
+++ b/migration/i9/oracle/202203090311_IDEMPIERE-5094.sql
@@ -1,0 +1,10 @@
+SELECT register_migration_script('202203090311_IDEMPIERE-5094.sql') FROM dual
+;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- IDEMPIERE-5094 Implement validation for costing level change ( Accounting Schema and Product Category Accounting )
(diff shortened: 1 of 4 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I don't understand but something doesn't look right here:

The message says:
Costing level can't be changed if there are product cost details created **with current costing level**

The query on Accounting Schema validation is:
SELECT DISTINCT p.Value
FROM M_Product p JOIN M_CostDetail d ON p.M_Product_ID = d.M_Product_ID
JOIN M_Product_Category_Acct pc ON p.M_Product_Category_ID = pc.M_Product_Category_ID AND d.C_AcctSchema_ID = pc.C_AcctSchema_ID
WHERE p.IsActive = 'Y' AND pc.IsActive = 'Y' **AND pc.CostingLevel IS NULL** AND d.C_AcctSchema_ID =?

The query on Product Category is:
SELECT DISTINCT p.Value
FROM M_Product p
JOIN M_CostDetail d ON p.M_Product_ID = d.M_Product_ID
WHERE p.IsActive = 'Y' AND p.M_Product_Category_ID =? AND d.C_AcctSchema_ID =?
(**without costing level check**)


Something looks wrong, or the validations are wrong, or the message is misleading.

Regards,

Carlos Ruiz

hengsin:
hi @CarlosRuiz-globalqss ,

For the first query, the costing level check is needed since MAcctSchema.CostingLevel doesn't apply if a product's product category has costing level set.

For the second query, that's not needed, the costing level change have been check before the getProductWithCosts call and we don't really care whether the product category's costing is null since we have check before the call that the product category's effective costing level have been changed.

CarlosRuiz-globalqss (approved):
understood, thanks

PR #1236, opened 2022-03-09
IDEMPIERE-5224 - Locator Field type Caching Improvement

PeterTakacs300:
Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5224

Implemented changes made by @igorpojzl

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MLocatorLookup.java
+++ b/org.adempiere.base/src/org/compiere/model/MLocatorLookup.java
@@ -92,7 +92,7 @@ public final class MLocatorLookup extends Lookup implements Serializable
 	private int					m_only_Product_ID = 0;
 
 	/** Storage of data  MLookups		*/
-	private volatile LinkedHashMap<Integer,MLocator> m_lookup = new LinkedHashMap<Integer,MLocator>();
+	private volatile LinkedHashMap<Integer,KeyNamePair> m_lookup = new LinkedHashMap<Integer,KeyNamePair>();
 
 	/**
 	 *  Dispose
@@ -178,9 +178,9 @@ public final class MLocatorLookup extends Lookup implements Serializable
 			return null;
 
 		//	try cache
-		MLocator loc = (MLocator) m_lookup.get(key);
+		NamePair loc = m_lookup.get(key);
 		if (loc != null)
-			return new KeyNamePair (loc.getM_Locator_ID(), loc.toString());
+			return loc;
 
 		//	Not found and waiting for loader
 		if (m_loader.isAlive())
@@ -188,10 +188,10 @@ public final class MLocatorLookup extends Lookup implements Serializable
 			log.fine("Waiting for Loader");
 			loadComplete();
 			//	is most current
-			loc = (MLocator) m_lookup.get(key);
+			loc = m_lookup.get(key);
 		}
 		if (loc != null)
-			return new KeyNamePair (loc.getM_Locator_ID(), loc.toString());
+			return loc;
 
 		//	Try to get it directly
 		return getDirect(key, true, null);
@@ -244,9 +244,9 @@ public final class MLocatorLookup extends Lookup implements Serializable
 			return null;
 		//
 		int key = loc.getM_Locator_ID();
+		KeyNamePair retValue = new KeyNamePair(key, loc.toString());
 		if (saveInCache)
-			m_lookup.put(Integer.valueOf(key), loc);
-		NamePair retValue = new KeyNamePair(key, loc.toString());
+			m_lookup.put(Integer.valueOf(key), retValue);
 		return retValue;
 	}	//	getDirect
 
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/model/MLocatorLookup.java
```diff
@@ -483,12 +486,13 @@ public void run()
...
-		ArrayList<Object> list = new ArrayList<Object>(collection.size());
-		Iterator<MLocator> it = collection.iterator();
+		Collection<KeyNamePair> collection = getData();
+		ArrayList<Object> list = new ArrayList<Object>(collection);
+		Iterator<KeyNamePair> it = collection.iterator();
 		while (it.hasNext())
 		{
-			MLocator loc = it.next();
+			KeyNamePair locPair = it.next();
+			MLocator loc = MLocator.get(locPair.getKey());
```
PeterTakacs300, line 495: To improve caching, this needs to be removed.

org.adempiere.base/src/org/compiere/model/MLocatorLookup.java
```diff
@@ -483,12 +486,13 @@ public void run()
...
 			if (isValid(loc))				//	only valid warehouses
 				list.add(loc);
```
PeterTakacs300, line 497: To improve caching, this needs to be removed.

hengsin:
hi @PeterTakacs300 , to facilitate review, please give a brief description on the solution to the issue mention in the ticket, thanks.

norbertbede:
This was originally developer by @igorpojzl. Please provide required informations

igorpojzl:
Goul in This PR is to remove all MLocator.get calls in MLocatorLookup.

MLocatorLookup is Storing MLocators(cache) to its collection which can cause to overflow MLocator Cache (between multiple clients). Which cause high volume of calls to Database. 
We prepared to change LocatorLookup to store KeyNamePairs.

There is still question and issue with Java method 
**public ArrayList<Object> getData (boolean mandatory, boolean onlyValidated, boolean onlyActive, boolean temporary, boolean shortlist) // IDEMPIERE 90**

which is loading MLocators from cache to validate each locator, and negating change in this PR.

hengsin:
Hi @igorpojzl , if the goals is to remove MLocator.get() db call, perhaps we should change the Loader to select * from MLocator and we can use the ResultSet constructor of MLocator instead ? That would works with the isvalid call too.

igorpojzl:
hi @hengsin we also removed logic for only stored product to allow set any locator for Product.

hengsin:
hi @igorpojzl , it seems the isValid method is make dummy now. If that's the case, we should mark that as deprecated and remove all call of that method from cire.

hengsin:
hi @CarlosRuiz-globalqss , this looks fine to me but I'm not sure whether the isValid change will break anything, wdyt ?

hengsin:
hi @CarlosRuiz-globalqss , this looks fine to me but I'm not sure whether the isValid change will break anything, wdyt ?

CarlosRuiz-globalqss:
Hi team,

Is not easy to track the impact of the changes.

From what I see, it seems the "only_Product" feature of locator editor is destroyed.

I followed the test case described by @hengsin at [IDEMPIERE-4382](https://idempiere.atlassian.net/browse/IDEMPIERE-4382) and it happens exactly what is described there, so this is creating a regression about that fix, and I think is because the locator doesn't respect now the "only product" stuff.


I have one doubt, according to the mattermost conversation, this is caused by REST loading the lookup, this is on the window rest?  Asking because I have not used those, just the model part.


Also, can you please describe better what is the issue?  Is MLookup loading the full list of 6.000 locators in memory?  If that's the case, how the changes make that better, how is not loading the same number of MLocator records now?


Sorry, I'm a bit lost jumping so late here.

Regards,

Carlos Ruiz

norbertbede:
Facts: 17 000 locators, 10 000 cache size, approx 6000/most active 2 clients. this produced approx 1mill calls.
we analyze cache in idempiere monitor and first 1 product call loaded 6000 locators to cache, next used it. 2nd tenant made eg. material receipt in webui loaded next 6000 so first 2000 was destroyed in cache.

we are sure this happened in REST model calls, and webui also. 
(i made 2 upgrades ferst rest, significant reduction then 2nd day webui, then goes down for )

After @igorpojzl  changes we save daily approx 900 000 db calls.

So the code is good for me :) Cant comment it.

1 more issue: Locator fieldtype lists only locators with storage levels, we changed this also. because the locator in product s for set default, this is considered in code in many places.

hengsin:
hi @norbertbede , @igorpojzl  & @CarlosRuiz-globalqss ,
I've implemented an alternative approach at pull/1242 with the following difference:
- use MLocator ResultSet constructor and is backward compatible
- implemented lazy loading of MLocatorLookup. With lazy loading, the locator editor wouldn't pull all locator records from db until the user open the locator dialog.

I've not done extensive testing with the lazy loading part so that might have issue(s) that I've not discover yet.

CarlosRuiz-globalqss (changes requested):
Comments on ticket:
https://idempiere.atlassian.net/browse/IDEMPIERE-5224?focusedCommentId=48122

This change is reopening an old fixed bug.

CarlosRuiz-globalqss:
Replaced with https://github.com/idempiere/idempiere/pull/1242

PR #1237, opened 2022-03-09
IDEMPIERE-5225 - HTML Report minification

PeterTakacs300:
Jira ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5225

Using com.googlecode.htmlcompressor.compressor.HtmlCompressor

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/ReportEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/ReportEngine.java
@@ -697,7 +697,7 @@ queued-job-count = 0  (class javax.print.attribute.standard.QueuedJobCount)
 			table.setNeedClosingTag(false);
 			PrintWriter w = new PrintWriter(writer);
 			XhtmlDocument doc = null;
-			boolean minify = MSysConfig.getBooleanValue(MSysConfig.HTML_REPORT_MINIFY, false, Env.getAD_Client_ID(getCtx()));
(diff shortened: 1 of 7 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @PeterTakacs300 / @norbertbede 

According to the latest comments in ticket, can you please add a client configurable SysConfig key to enable turning off the minification.

The SysConfig can be named "HTML_REPORT_MINIFY" and default to Y

org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -695,9 +697,10 @@ public boolean createHTML (Writer writer, boolean onlyTable, Language language,
 			table.setNeedClosingTag(false);
 			PrintWriter w = new PrintWriter(writer);
 			XhtmlDocument doc = null;
-					
+			boolean minify = MSysConfig.getBooleanValue(MSysConfig.HTML_REPORT_MINIFY, false, Env.getAD_Client_ID(getCtx()));
```
CarlosRuiz-globalqss, line 700: The default here must be true.  I know in some way is useless as the SysConfig record already exists, but is better when doing maintenance to know here which is the default.

CarlosRuiz-globalqss (approved):

PR #1241, opened 2022-03-10
IDEMPIERE-2494 Attribute Value allowing duplicated values #resolve #f…

laydasalasc:
https://idempiere.atlassian.net/browse/IDEMPIERE-2494


CarlosRuiz-globalqss (approved):
tested

PR #1242, opened 2022-03-12
IDEMPIERE-5224 Locator Field type Caching Improvement

hengsin:
- Replace M_Locator.get with MLocator ResultSet constructor
- Implement lazy loading for MLocatorLookup


CarlosRuiz-globalqss (commented):
Comments on ticket:
https://idempiere.atlassian.net/browse/IDEMPIERE-5224?focusedCommentId=48122

Suggesting that we don't need to add a new constructor, just make the MLocatorLookup lazy always (same as all the other lookups).

CarlosRuiz-globalqss (approved):

PR #1244, opened 2022-03-13
IDEMPIERE-3334: Adding support for Readonly Logic, Set Pressed Logic …

dpansheriya:
…on custom toolbar button

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
@@ -898,6 +898,8 @@ public class ADWindowToolbar extends FToolbar implements EventListener<Event>
 		ADWindow adwindow = ADWindow.findADWindow(this);
 
 		dynamicDisplay();
+		pressedLogic();
+		readOnlyLogic();
 		// If no workflow set for the table => disable btnWorkflow
 		if (!btnActiveWorkflows.isDisabled()) {
 			GridTab gridTab = adwindow.getADWindowContent().getActiveGridTab();
@@ -977,6 +979,22 @@ public class ADWindowToolbar extends FToolbar implements EventListener<Event>
 					}
 				}
 			}
+		}
+	}
+	
+	public void pressedLogic()
+	{
+		for (ToolbarCustomButton toolbarCustomBtn : toolbarCustomButtons)
+		{
+			toolbarCustomBtn.pressedLogic();
+		}
+	}
+
+	public void readOnlyLogic()
+	{
+		for (ToolbarCustomButton toolbarCustomBtn : toolbarCustomButtons)
+		{
+			toolbarCustomBtn.readOnlyLogic();
 		}
 	}
 
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
@@ -898,8 +898,6 @@ public class ADWindowToolbar extends FToolbar implements EventListener<Event>
 		ADWindow adwindow = ADWindow.findADWindow(this);
 
 		dynamicDisplay();
-		pressedLogic();
-		readOnlyLogic();
 		// If no workflow set for the table => disable btnWorkflow
 		if (!btnActiveWorkflows.isDisabled()) {
 			GridTab gridTab = adwindow.getADWindowContent().getActiveGridTab();
@@ -980,9 +978,11 @@ public class ADWindowToolbar extends FToolbar implements EventListener<Event>
 				}
 			}
 		}
+		pressedLogic();
+		readOnlyLogic();
 	}
 	
-	public void pressedLogic()
+	private void pressedLogic()
 	{
 		for (ToolbarCustomButton toolbarCustomBtn : toolbarCustomButtons)
 		{
@@ -990,7 +990,7 @@ public class ADWindowToolbar extends FToolbar implements EventListener<Event>
 		}
 	}
 
-	public void readOnlyLogic()
+	private void readOnlyLogic()
 	{
 		for (ToolbarCustomButton toolbarCustomBtn : toolbarCustomButtons)
 		{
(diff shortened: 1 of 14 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
```diff
@@ -898,6 +898,8 @@ public void updateToolbarAccess() {
 		ADWindow adwindow = ADWindow.findADWindow(this);
 
 		dynamicDisplay();
+		pressedLogic();
```
hengsin, line 901: I think it is better to call both logic method at the end of dynamicDisplay instead. With this, we can make both *logic method private and don't have to hunt for places where dynamicDisplay is call.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
dpansheriya: @hengsin  Please note that I made changes as suggested.

hengsin (changes requested):
please rename migration script to current timestamp instead of using timestamp from 2017.

CarlosRuiz-globalqss:
The oracle script has two problems:

- ALTER TABLE ... VARCHAR2(2000) ->  must be written as VARCHAR2(2000 CHAR)
- there are lines starting with the character @ - these throw errors when running with sqlplus - this was solved in [IDEMPIERE-4481](https://idempiere.atlassian.net/browse/IDEMPIERE-4481) - and the script here is using the old description of the element

hengsin:
hi @dpansheriya , any update on this ?

dpansheriya:
> * sqlplus

It got completely missed in holiday season here. Let me work on it.

dpansheriya:
> * VARCHAR2(2000 CHAR)

Please note that done all and pushed.

dpansheriya:
> please rename migration script to current timestamp instead of using timestamp from 2017.

This is done.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
```diff
@@ -104,7 +107,77 @@ public void dynamicDisplay(boolean forceValidation) {
...
+			isSQL = true;
+		}
+
+		logic = Env.parseContext(Env.getCtx(), windowNo, tabNo, logic, false);
+
+		if (isSQL) {
+			int result = DB.getSQLValue(null, logic);
+
+			if (result == 1)
+				isValid = true;
```
hengsin, line 173: This is inconsistent with the use of SQL logic in other part of the application. Preferably, should use Evaluator.parseSQLLogic instead.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
dpansheriya: Changed to use Evaluator.parseSQLLogic

hengsin (changes requested):
Not visible (not implemented ?) for Window > Tab > Toolbar Button.

dpansheriya:
> 

@hengsin  may be I am not understanding your questions. What I understand is you are asking that display logic not implemented. But display logic field already exists in master branch.  Can you please clarify if I misunderstood your point?

hengsin:
hi @dpansheriya , no, I means at "Window, Tab and Field", the Readonly Logic and Pressed Logic is not shown. Is that by design or just overlook ?

dpansheriya:
> hi @dpansheriya , no, I means at "Window, Tab and Field", at the "Toolbar Button" tab, the Readonly Logic and Pressed Logic field is not shown. Is that by design or just overlook ?

Ahh, We have added Pressed Logic and Read Only logic on Toolbar window only. Not on Window->Tab->Toolbar button.

hengsin:
hi @dpansheriya , that's using the same table (AD_ToolbarButton), shouldn't that be added there as well ?

dpansheriya:
> hi @dpansheriya , that's using the same table (AD_ToolbarButton), shouldn't that be added there as well ?

I think yes, So let us add and test.

dpansheriya:
> hi @dpansheriya , that's using the same table (AD_ToolbarButton), shouldn't that be added there as well ?

@hengsin  please note this support added.

hengsin:
[pr1244.patch.txt](https://github.com/idempiere/idempiere/files/10049193/pr1244.patch.txt)
- context will be evaluated in evaluateLogic. current approach cause evaluation exception if context is not parseable.

hengsin (changes requested):
hi @dpansheriya , please see attached patch above.

dpansheriya:
> hi @dpansheriya , please see attached patch above.

Applied

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
```diff
@@ -104,7 +106,70 @@ public void dynamicDisplay(boolean forceValidation) {
...
+	private boolean validateLogic(String logic, int tabNo) {
+		boolean isValid = false;
+
+		if (logic.startsWith("@SQL="))
+		{
+			isValid = Evaluator.parseSQLLogic(logic, Env.getCtx(), windowNo, tabNo, "");
+		}
+		else
+		{
+			logic = Env.parseContext(Env.getCtx(), windowNo, tabNo, logic, false);
```
hengsin, line 166: hi @dpansheriya , miss this in my previous patch, this line should be removed as well.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ToolbarCustomButton.java
dpansheriya: @hengsin  this is done.

hengsin (approved):
tested

PR #1247, opened 2022-03-15
IDEMPIERE-5230 DesktopImpl.warn: UUID recycle is enabled and it's better to disable it by specifying a library property 'org.zkoss.zk.ui.uuidRecycle.disabled' with true to prevent some unwanted widget uuid reusing at client side accidentally.

CarlosRuiz-globalqss:
[…](https://idempiere.atlassian.net/browse/IDEMPIERE-5230)


hengsin (approved):
looks good

PR #1250, opened 2022-03-15
IDEMPIERE-4878 : Thin scrollbars

nmicoud:
Co-Authored-By: pmargery <87129041+pmargery@users.noreply.github.com>

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/metainfo/zk/lang-addon.xml
+++ b/org.adempiere.ui.zk/WEB-INF/src/metainfo/zk/lang-addon.xml
@@ -56,5 +56,5 @@ Copyright (C) 2007 Ashley G Ramdass (ADempiere WebUI).
 	
 	<!-- this js module doesn't actually exists and it is here for default theme version -->
 	<!-- since loading of  js module is on demand, it doesn't cause any error as long as you don't try to load it -->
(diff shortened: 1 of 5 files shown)


nmicoud:
Hi,
Is a first commit for thin scrollbars.
There are still several ones which are still "big":
- horitontal on windows
- vertical on process
- vertical/horizontal on reports
- vertical on About (Credits) and horizontal on About (Info)

Any hint will be appreciated so we can fix them

Regards

Nicolas

hengsin:
Hi @nmicoud ,

Can you post some before and after images to the tickets?

I'm not seeing any differences at my testing here, perhaps this is OS and Browser dependent.

hengsin:
https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width - very poor compatibility across browser. While you can try to simulate that at Chrome/Safari with ::-webkit-scrollbar, it is not exactly the same thing. I would recommend not to add this to core until it is widely supported by most main stream browser (as it is, firefox has < 5% market share).

nmicoud:
Not sure to understand what you recommend here ; stop working on that ticket and return "later" (when a single syntax will work for all browsers)?

hengsin:
Hi @nmicoud , that's what I recommend. It is very verbose and difficult to maintain with current compatibility issue.

nmicoud:
ok, thanks for confirming this

PR #1252, opened 2022-03-17
IDEMPIERE-4117 String editor with copy button

hengsin:
- mouse over a string editor field and the copy text button is visible
for about 5s


CarlosRuiz-globalqss:
Comments in ticket
https://idempiere.atlassian.net/browse/IDEMPIERE-4117?focusedCommentId=48149

PR #1253, opened 2022-03-17
IDEMPIERE-5056 Order and Invoice: Tax lookup interface

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5056

### DIFF
--- a/org.adempiere.base/src/org/adempiere/base/Core.java
+++ b/org.adempiere.base/src/org/adempiere/base/Core.java
@@ -1068,7 +1068,7 @@ public class Core {
 	 * @return ITaxLookup service
 	 */
 	public static ITaxLookup getTaxLookup() {
-		String service = MSysConfig.getValue(MSysConfig.TAX_LOOKUP_SERVICE, Env.getAD_Client_ID(Env.getCtx()));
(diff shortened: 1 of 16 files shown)


org.adempiere.base/src/org/adempiere/base/Core.java
```diff
@@ -1061,4 +1062,18 @@ public static IReservationTracerFactory getReservationTracerFactory() {
 		
 		return DefaultReservationTracerFactory.getInstance();
 	}
+	
+	/**
+	 * Get tax lookup service
+	 * @return ITaxLookup service
+	 */
+	public static ITaxLookup getTaxLookup() {
+		String service = MSysConfig.getValue(MSysConfig.TAX_LOOKUP_SERVICE, Env.getAD_Client_ID(Env.getCtx()));
```
CarlosRuiz-globalqss, line 1071: My only suggestion would be to add explicitly here the default, like:
```
String service = MSysConfig.getValue(MSysConfig.TAX_LOOKUP_SERVICE, "org.adempiere.base.DefaultTaxLookup", Env.getAD_Client_ID(Env.getCtx()));
```

Just as a precaution, so this can keep working in case somebody delete or inactivate the SysConfig record.

CarlosRuiz-globalqss (approved):

PR #1254, opened 2022-03-17
IDEMPIERE-5089 Add MSysConfig.ALLOW_REVERSAL_OF_RECONCILED_PAYMENT

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5089

### DIFF
--- a/org.adempiere.base/src/org/compiere/wf/MWFActivity.java
+++ b/org.adempiere.base/src/org/compiere/wf/MWFActivity.java
@@ -990,7 +990,11 @@ public class MWFActivity extends X_AD_WF_Activity implements Runnable
 				if (m_process != null)
 				{
 					m_process.setProcessMsg(this.getTextMsg());
-					m_process.saveEx();
+					try {
+						m_process.saveEx();
+					} catch (Exception ex) {
+						log.log(Level.SEVERE, ex.getMessage(), ex);
+					}
 				}
 			} finally {
 				if (contextLost)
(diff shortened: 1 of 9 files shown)


org.idempiere.test/src/org/idempiere/test/model/MProductTest.java
```diff
@@ -231,8 +231,9 @@ public void testRemoveStorageRecords() {
 		product.setIsActive(false);
 		assertThrows(AdempiereException.class, () -> product.saveEx());
 		
-		//clear on hand so that we can deactivate product
+		//clear on hand and reservation so that we can deactivate product
```
CarlosRuiz-globalqss, line 234: maybe this change is from another ticket?

org.adempiere.base/src/org/compiere/wf/MWFActivity.java
```diff
@@ -990,7 +990,11 @@ else if (savepoint != null)
 				if (m_process != null)
 				{
 					m_process.setProcessMsg(this.getTextMsg());
-					m_process.saveEx();
+					try {
```
CarlosRuiz-globalqss, line 993: this looks dangerous - it is ignoring exceptions when saving MWFProcess - is that what you want?

org.idempiere.test/src/org/idempiere/test/model/MProductTest.java
hengsin: I just fix the unit test error when I'm testing this ticket. I can move it else where if that's better.

CarlosRuiz-globalqss: no problem - it was just curious

CarlosRuiz-globalqss (approved):
tested

PR #1255, opened 2022-03-17
IDEMPIERE-5235 : Usage of MOrg.getOfClient without PO

nmicoud:


org.adempiere.base/src/org/compiere/model/MOrg.java
```diff
@@ -48,18 +48,38 @@
...
+	}
+
+	/**
+	 * 	Get Active Organizations Of Client
+	 *	@param ctx context
+	 *	@param int clientID
+	 *	@return array of orgs
+	 */
+	public static MOrg[] getOfClient (Properties ctx, int clientID)
+	{
```
hengsin, line 70: this can be simplified by removing the ctx parameter. use env.getctx() here and change the 2 call above to return getOfClient(po.getAD_Client_ID()) and return getOfClient(Env.getAD_Client_ID(Env.getCtx())

hengsin (approved):
looks good

PR #1256, opened 2022-03-17
IDEMPIERE-5233 - Invoice created from the Commission Run window canno…

d-ruiz:
…t be easly found after created

https://idempiere.atlassian.net/browse/IDEMPIERE-5233


CarlosRuiz-globalqss (approved):
Tested successfully

PR #1257, opened 2022-03-18
IDEMPIERE-5221 Performance - indexes on heavily used queries / FHCA-3486

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5221


CarlosRuiz-globalqss:
NOTE: I added "drop index if exists" before creating the index, because in some instances is possible that the index is already created.  It is harmless if the index does not exist.

The oracle version of "drop index if exists" involves pl/sql, so I wrote that part too.

Regards,

Carlos Ruiz

hengsin (approved):
looks good

PR #1263, opened 2022-03-24
IDEMPIERE-5232 : Chart field at tenant level - filter datasource (Sys…

nmicoud:
…tem and current tenant only)


CarlosRuiz-globalqss (approved):
looks good

PR #1264, opened 2022-03-24
IDEMPIERE-4669 Adding RfQ responses manually is prohibited, because the field for the RfQ is read only

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4669


hengsin (approved):
looks good

PR #1266, opened 2022-03-25
IDEMPIERE-3219 Reset Accounting can't reset documents without fact acct

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-3219


hengsin (approved):
looks good

PR #1267, opened 2022-03-26
IDEMPIERE-4550 Fix columns length difference between AD and DB #resolve

laydasalasc:
https://idempiere.atlassian.net/browse/IDEMPIERE-4550


CarlosRuiz-globalqss (approved):
tested successfully in oracle and postgresql

PR #1273, opened 2022-03-31
IDEMPIERE-5248 - Report View reference: list doesn't work

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5248

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -535,8 +535,7 @@ public class DataEngine
 					}
 					// 	TableName.ColumnName,
 					sqlSELECT.append(lookupSQL).append(" AS ").append(ColumnName).append(",");
-					if (!IsGroupFunction)
-						groupByColumns.add(ColumnName); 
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -535,6 +535,8 @@ else if (DisplayType.isList(AD_Reference_ID)
 					}
 					// 	TableName.ColumnName,
 					sqlSELECT.append(lookupSQL).append(" AS ").append(ColumnName).append(",");
+					if (!IsGroupFunction)
+						groupByColumns.add(ColumnName);
```
hengsin, line 539: 2 difference from the "AD_Reference_ID == DisplayType.Table" block:

1. It is groupByColumns.add(lookupSQL) at Table block, not groupByColumns.add(ColumnName)
2. No IsGroupFunction check

Please check the inconsistency, thanks.

hengsin (changes requested):

hengsin (approved):
looks good

PR #1251, opened 2022-03-16
IDEMPIERE-4753 Make printing filename configurable

PeterTakacs300:
Jira ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4753

Implemented changes from https://github.com/idempiere/idempiere/pull/709

Multilanguage and Jasper report implemented

PR #1278, opened 2022-04-04
IDEMPIERE-4753 Make printing filename configurable - bug

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4753
Referenced PR: https://github.com/idempiere/idempiere/pull/1251

Bug: When we change print format, the file name stays the same. 

E.g. we have 2 print formats with different file name pattern, then only that one is applied, with which the report was opened. If we change print format after the report is opened, the filename stays the same.


hengsin (approved):
looks good

PR #1281, opened 2022-04-05
IDEMPIERE-5088 Implement Bank Transfer as Transactional Document like…

hengsin:
… Payment

https://idempiere.atlassian.net/browse/IDEMPIERE-5088


CarlosRuiz-globalqss (changes requested):
The migration script has several errors, please check the following patch:
[migr5088.txt](https://github.com/idempiere/idempiere/files/8457081/migr5088.txt)

About testings, please check comments in ticket

CarlosRuiz-globalqss (changes requested):
Thanks @hengsin - basic tests went good.

The only thing I see pending (according to our ticket conversation) would be to restore back the old "Bank/Cash Transfer" process (sorry, I suggested to inactivate it) - and rename it to "Bank/Cash Transfer Process"

Regards,

Carlos Ruiz

hengsin:
hi @CarlosRuiz-globalqss ,

Perhaps it is better to change the process to create the new bank transfer document instead ?

Regards,
Low

CarlosRuiz-globalqss:
> Perhaps it is better to change the process to create the new bank transfer document instead ?

Yes, sounds like a good alternative.

CarlosRuiz-globalqss (approved):

CarlosRuiz-globalqss:
@hengsin - can you please check the commit https://github.com/idempiere/idempiere/pull/1281/commits/388eb9b

I changed the bank transfer process as you suggested to use the new document (optionally with a parameter).

My plan initially was a complete replacement, but then I found two use cases that sometimes I require in some implementations:
- the bank transfer process can manage creating different document numbers for the payment and the receipt
- the bank transfer process can manage using a different currency than the bank account

So, with my changes is possible to use the "old way" or the new document, both work.

Can you please check and if everything is right, please push it.

Regards,

Carlos Ruiz

PR #1282, opened 2022-04-06
IDEMPIERE-5256  Add Cache to MSession

igorpojzl:
Jira Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5256

Added Caching to MSession. Added Parameter IsImmutable to select if we want to return immutable object (cached) or load from DB.

It Looks like Vector had some validation logic (Records Check) which cannot be copied to new Caching. Updating MSession record will delete cached data. 

Added Deprecated tag to previous methods.


hengsin (approved):
looks fine

PR #1283, opened 2022-04-07
IDEMPIERE-5256 Refactor

igorpojzl:
Previous PR: https://github.com/idempiere/idempiere/pull/1282
JIRA Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5256


hengsin:
@CarlosRuiz-globalqss , this is probably a cleaner approach than the previous PR, do you prefer this over what have been merged ?

@igorpojzl , we have getCopy approach implemented for other cache to get a writable copy of a record from cache and avoid loading from db at the same time, perhaps that can be an alternative option here too.

CarlosRuiz-globalqss:
Hi @hengsin - yes, the approach looks good.

Also, I tested and is solving the problem reported by @tony-tspc in the ticket.

hengsin (approved):
tested by Carlos

PR #1287, opened 2022-04-08
IDEMPIERE-5262 Implement readonly protection for DB.getSQLValueEx call

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5262


CarlosRuiz-globalqss (changes requested):
marking as required changes as per the conversation on ticket

hengsin:
> marking as required changes as per the conversation on ticket

Please see my reply in the ticket, the raise issue is not valid.

CarlosRuiz-globalqss (approved):
Looks good - all unit tests passed

PR #1289, opened 2022-04-11
IDEMPIERE-5265 : Wizard for Chosen Multiple Selection List editor

nmicoud:
Hi,

Can you have a look and tell me if is ok to integrate it.
If so, I'll add missing messages and javadoc.

That can be tested via on any Chosen Multiple Selection List field.
In the standard => "Authorization Credential" window > Scope list field

nb: in our implementation, we didn't have dynamic validation on those fields.
But you'll see it will be needed here.
My idea is to add a 'filter' parameter to MRefList.getList (see TODO in code).

wdyt?

Regards,

Nicolas


hengsin:
Hi @nmicoud ,

The changes looks fine but calling it "Wizard" is misleading.
There are many method that's not being marked as private, protected or public, that's not a good practice.
For reordering, I think the more natural improvement is to allow reordering of item by drag and drop.

Regards,
Low

nmicoud:
hi @hengsin,
In our French version, I used "Assistant", that can be translated to Wizard or Helper. Maybe Helper is a better name?
I'll add missing access modifiers.
We had > 20 items in our list, and in this case, I think is more comfortable to have a "real" panel for adding/removing/reordering.
But I think both approachs can coexist
wdyt?

hengsin:
hi @nmicoud ,

"Assistant" sounds better to me.

Regards,
Low

nmicoud:
Ho @hengsin ,
Changes commited

Regards
Nicolas

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
```diff
@@ -622,4 +680,399 @@ public void newRecord(int record_ID) {
...
+			setTitle(gridField.getHeader() + " " + Msg.getMsg(Env.getCtx(), "Assistant"));
+			init();
+			load();
+			setClosable(true);
+			setBorder("normal");
+			setShadow(true);
+			setMaximizable(true);
+			setSizable(true);
+			setHeight("600px");
+			setWidth("700px");
```
hengsin, line 708: this is not mobile friendly, use ZKUpdateUtil.setWindowHeightX and ZKUpdateUtil.setWindowWidthX instead

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
hengsin: to use isUseCSSForWindowSize(), you have to change add the configuration for your window to the default theme's css, it is not auto.

nmicoud: Hi,
Ok, should be ok now
Thanks

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
```diff
@@ -416,26 +456,30 @@ public void onEvent(Event event)
...
-		        	}
-		        }
-    		} finally {
-    			onselecting = false;
-    		}
+    		updateValue(getValueFromComponent());
     	}
     }
+    
+    void updateValue(Object newValue) {
```
hengsin, line 463: please mark method as private or protected

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
```diff
@@ -622,4 +680,410 @@ public void newRecord(int record_ID) {
...
+					Clients.response(focus);
+				}
+			}
+		}
+
+		private String getNewValue() {
+			return m_newValue;
+		}
+
+		private class ListElement extends NamePair {
```
hengsin, line 1075: why do we need this instead of just using the existing ValueNamePair class ?

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
nmicoud: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WChosenboxListEditor.java
nmicoud: we don't - is the result of useless copy-pastes

nmicoud:
I also fix a java.lang.StringIndexOutOfBoundsException when closing the panel and there's no items inside

hengsin:
[1289_pr.patch.txt](https://github.com/idempiere/idempiere/files/8569230/1289_pr.patch.txt)
Fix:
- focus after closing of assistant dialog
- selected and available list not of same height
- Authorization Credential window: assistant dialog throw sql error if authorization provider field is empty

nmicoud:
integrated, thanks

hengsin (approved):
tested

PR #1292, opened 2022-04-12
IDEMPIERE-5267 - Change AD_Column default cache size

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5267

Implemented changes from @igorpojzl CLDE#314


CarlosRuiz-globalqss:
Suggesting https://github.com/idempiere/idempiere/pull/1325 instead of this patch.
1325 allows to set the cache max size in a system property

PR #1293, opened 2022-04-14
IDEMPIERE-5254 - AD create postgres sequences and ad_sequence for views

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5254

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MTable.java
+++ b/org.adempiere.base/src/org/compiere/model/MTable.java
@@ -721,13 +721,15 @@ public class MTable extends X_AD_Table implements ImmutablePOSupport
 		if (!success)
 			return success;
 		//	Sync Table ID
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -722,9 +722,9 @@ protected boolean afterSave (boolean newRecord, boolean success)
 			return success;
 		//	Sync Table ID
 		MSequence seq = MSequence.get(getCtx(), getTableName(), get_TrxName());
```
hengsin, line 724: The whole MSequence seq block should be enclosed inside a !isView() block (why call MSequence.get if it is a view ?).

hengsin (changes requested):

CarlosRuiz-globalqss (approved):
tested successfully

PR #1294, opened 2022-04-15
IDEMPIERE-5259 : Labels

nmicoud:


nmicoud:
see comments in ticket

norbertbede:
we did some tests, looks properly as first look. technically PR work, we are able add/remove labels.
i made some notes here: https://mattermost.idempiere.org/idempiere/pl/pqqdus888br1jned54f8wn4o7o
however as-is good start. to get it really attractive from UX perspecitve need some additional works.
+ 1 commit to the core.

norbertbede:
@nmicoud we suggest to allow assign labels in system tenant well. would be good to change data access level to system+client.  we done it locally, then we got crosstenant issue.

nmicoud:
hi @norbertbede 
done !

PR #1295, opened 2022-04-19
IDEMPIERE-5271 : Can't override reference in AD_UserDef_Proc_Paramete…

nmicoud:
…r to use ChosenMultipleSelectionTable as report parameter


CarlosRuiz-globalqss (approved):
tested successfully

PR #1296, opened 2022-04-20
IDEMPIERE-5132 Enable filling Purchase Order in Material Receipt - fix issue with drop ship

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5132


hengsin (approved):
tested

PR #1297, opened 2022-04-20
IDEMPIERE-5269 Allow save a record in a read only search field

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5269

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
@@ -342,8 +342,10 @@ public class WSearchEditor extends WEditor implements ContextMenuListener, Value
 			boolean isQuickFormComp = false;
 			if (getComponent().getAttribute(QuickGridTabRowRenderer.IS_QUICK_FORM_COMPONENT) != null)
 				isQuickFormComp = (boolean) getComponent().getAttribute(QuickGridTabRowRenderer.IS_QUICK_FORM_COMPONENT);
-			
-			if ((getComponent().getText() == null || getComponent().getText().length() == 0) && !isQuickFormComp)
+
+			if (   (getComponent().getText() == null || getComponent().getText().length() == 0)
+				&& !isQuickFormComp
+				&& (gridField != null && gridField.isEditable(true)))
 			{
 				// open Info window similar to swing client
 				if (infoPanel != null)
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
```diff
@@ -342,8 +342,10 @@ else if ((Events.ON_OK.equals(e.getName()))) {
 			boolean isQuickFormComp = false;
 			if (getComponent().getAttribute(QuickGridTabRowRenderer.IS_QUICK_FORM_COMPONENT) != null)
 				isQuickFormComp = (boolean) getComponent().getAttribute(QuickGridTabRowRenderer.IS_QUICK_FORM_COMPONENT);
-			
-			if ((getComponent().getText() == null || getComponent().getText().length() == 0) && !isQuickFormComp)
+
+			if (   (getComponent().getText() == null || getComponent().getText().length() == 0)
+				&& !isQuickFormComp
+				&& (gridField != null && gridField.isEditable(true)))
```
hengsin, line 348: simpler to replace "&& (gridField != null && gridField.isEditable(true))" with "&& isReadWrite()"

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
CarlosRuiz-globalqss: I don't know why but in my tests isReadWrite() was returning true even when the field was read-only.

My test was:
- create a Window Customization for role _GardenWorld Admin not Advanced_ for _Sales Order > Order > Business Partner_ field and make it read-only
- with that role enter to _Sales Order_, navigate to the _Business Partner_ field using tab key and push the Enter key
- at that point isReadWrite was returning me "true" (bug maybe?)

Regards,

Carlos Ruiz

hengsin: Hi @CarlosRuiz-globalqss ,

I try with your test case above but couldn't reproduce the same issue (i.e info window not open upon enter with "&& isReadWrite()" condition.

Regards,
Low

CarlosRuiz-globalqss:
Thanks @hengsin - tested, it worked this time, maybe something wrong in my eclipse in the previous test.

PR #1298, opened 2022-04-21
IDEMPIERE-5014 Add stream interface to Query

kriegfrj:
https://idempiere.atlassian.net/browse/IDEMPIERE-5014

Carlos requested a proof-of-concept - this is it.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/Query.java
+++ b/org.adempiere.base/src/org/compiere/model/Query.java
@@ -28,7 +28,12 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Properties;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Consumer;
 import java.util.logging.Level;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.adempiere.exceptions.DBException;
 import org.compiere.util.CLogger;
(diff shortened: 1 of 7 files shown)


hengsin (changes requested):
looks good, just need to add missing javadoc and update the existing one to reflect the changes make.

kriegfrj:
Made changes to the Javadoc as requested and also added test cases thanks to @hengsin 's pointer as to where to put them.

Also added the AssertJ dependency which I made use of in my tests.

hengsin:
hi @kriegfrj ,

The added dependency is fine but you need to add the location to "maven.locations.xml", and update the timestamp (yymmdd) based target name inside "org.idempiere.p2.targetplatform.target"  and "org.idempiere.p2.targetplatform.tpd" ( for e.g to idempiere-220428 ).

Regards,
Low

kriegfrj:
> hi @kriegfrj ,
> 
> The added dependency is fine but you need to add the location to "maven.locations.xml", and update the timestamp (yymmdd) based target name inside "org.idempiere.p2.targetplatform.target" and "org.idempiere.p2.targetplatform.tpd" ( for e.g to idempiere-220428 ).
> 
> Regards, Low

Thanks @hengsin , done, done and done.

I take it that `org.idempiere.p2.targetplatform.target` is in some way derived from the other two files - are there any instructions on how to generate it? Thanks!

hengsin:
[- install plugin from https://github.com/eclipse-cbi/targetplatform-dsl
org.idempiere.p2.targetplatform.target is generated from org.idempiere.p2.targetplatform.tpd using Eclipse plugin from  . the maven.locations.xml content](https://wiki.idempiere.org/en/Building_iDempiere_by_tycho#update_.5Bidempiere-source.5D.2Forg.idempiere.p2.targetplatform.2Forg.idempiere.p2.targetplatform.target)

hengsin (approved):
looks good

PR #1300, opened 2022-04-23
IDEMPIERE-5093 Scheduler cron pattern scheduling is always using serv…

hengsin:
…er time zone

Fix unit test failure


CarlosRuiz-globalqss (approved):
unit test pass

PR #1302, opened 2022-04-25
IDEMPIERE-5274 : Report customization panel : increase width of print…

nmicoud:
… format name


nmicoud:
Hi,
The change forces the width to 400px (not sure if is the best way)

Regards,

Nicolas

hengsin:
[IDEMPIERE-5274.patch.txt](https://github.com/idempiere/idempiere/files/8561329/IDEMPIERE-5274.patch.txt)

hi @nmicoud , instead of setting a fixed width for the name component, the attach patch works fine for me. please check it out.

nmicoud:
hi @hengsin 
perfect! 
thanks

hengsin (approved):
tested

PR #1303, opened 2022-04-26
IDEMPIERE-5276 : NPE on MInfoWindow.get

nmicoud:


hengsin (approved):
trivial fix, good to go

PR #1305, opened 2022-04-26
IDEMPIERE-5277 - Fixed list on Report dialogue not filtering inactive…

d-ruiz:
… print formats

https://idempiere.atlassian.net/browse/IDEMPIERE-5277


hengsin (approved):
trivial fix, good to go

PR #1307, opened 2022-04-27
IDEMPIERE-5278 : Impossible to log in with panel from plugin automati…

nmicoud:
…cally opened at login


nmicoud:
See solution in ticket

PR #1308, opened 2022-04-28
IDEMPIERE-5280 : Avoid Msg.getMsg: NOT found log when confirming an a…

nmicoud:
…ction on a document

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/WDocActionPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/WDocActionPanel.java
@@ -379,7 +379,7 @@ public class WDocActionPanel extends Window implements EventListener<Event>, Dia
 				String docAction = lstDocAction.getSelectedItem().getLabel();
 				MessageFormat mf = new MessageFormat(Msg.getMsg(Env.getAD_Language(Env.getCtx()), "ConfirmOnDocAction"));
 				Object[] arguments = new Object[]{docAction};
-				FDialog.ask(0, this, mf.format(arguments), new Callback<Boolean>() {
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/WDocActionPanel.java
```diff
@@ -379,7 +379,7 @@ public void onOk(final Callback<Boolean> callback) {
 				String docAction = lstDocAction.getSelectedItem().getLabel();
 				MessageFormat mf = new MessageFormat(Msg.getMsg(Env.getAD_Language(Env.getCtx()), "ConfirmOnDocAction"));
 				Object[] arguments = new Object[]{docAction};
-				FDialog.ask(0, this, mf.format(arguments), new Callback<Boolean>() {
+				FDialog.ask(0, this, "", mf.format(arguments), new Callback<Boolean>() {
```
hengsin, line 382: not the right fix, should replace the first argument, 0 with gridTab.getWindowNo() instead.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/WDocActionPanel.java
nmicoud: done

hengsin:
[pr1308.patch.txt](https://github.com/idempiere/idempiere/files/8582215/pr1308.patch.txt)
Fixed vertical alignment of message.

nmicoud:
patch integrated ; I also apply it at another place

hengsin (approved):
tested

PR #1309, opened 2022-04-29
IDEMPIERE-5279 Improve PO to avoid calling cache reset unnecessarily

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5279

https://mattermost.idempiere.org/idempiere/pl/89kadc1skidwxbe5khdafrtzmh

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -2538,11 +2538,14 @@ public abstract class PO
 			m_newValues = new Object[size];
 			m_createNew = false;
 		}
-		if (!newRecord) {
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - this pull request implements what we talked in mattermost.

However, I noticed that C_Order table for example doesn't have a cache but still is calling the reset cache because there is a DirectKeyNamePairCache in MLookup referencing this table (and I think any table with a lookup field)

I'm not sure if is OK to call reset for the lookup cache on a specific Record ID, or if that can/must be avoided.

Regards,

Carlos Ruiz

hengsin:
Hi @CarlosRuiz-globalqss , I think the changes is sufficient for now.

hengsin (approved):
looks good

PR #1313, opened 2022-05-05
IDEMPIERE-5260 Enhance Discount Schema for Fixed Pricing

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5260


CarlosRuiz-globalqss (changes requested):
Hi @hengsin - the manual tests passed sucessfully.

But the unit test MDiscountSchemaTest.testFixedPriceDiscount is failing, it is returning 3 as the price, I suspect it is because the discount schema is cached first on the testPercentageDiscount.

CarlosRuiz-globalqss (approved):
unit test pass

manual tests OK

PR #1314, opened 2022-05-05
IDEMPIERE-4723 - refactor MArchive.get to use the Query class instead…

d-ruiz:
… of PreparedStatement, following best practices

https://idempiere.atlassian.net/browse/IDEMPIERE-4723

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MArchive.java
+++ b/org.adempiere.base/src/org/compiere/model/MArchive.java
@@ -25,6 +25,7 @@ import java.nio.file.Path;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Properties;
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/MArchive.java
```diff
@@ -59,28 +61,16 @@ public class MArchive extends X_AD_Archive {
 	 * @return archives
 	 */
 	public static MArchive[] get(Properties ctx, String whereClause) {
-		ArrayList<MArchive> list = new ArrayList<MArchive>();
-		StringBuilder sql = new StringBuilder("SELECT * FROM AD_Archive WHERE AD_Client_ID=?");
-		if (whereClause != null && whereClause.length() > 0)
+		List<MArchive> list = new ArrayList<MArchive>();
```
hengsin, line 64: this should be simplified to List<MArchive> list = new Query ... below.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MArchive.java
d-ruiz: Done.

hengsin (approved):
looks good

PR #1315, opened 2022-05-05
IDEMPIERE-5063 Improve Unit Tests

CarlosRuiz-globalqss:
- revert -> Close DB connections after each unit test class
- implement a different approach to close the transactions on tearDown even when not active

https://idempiere.atlassian.net/browse/IDEMPIERE-5063

### DIFF
--- a/org.idempiere.test/src/org/idempiere/test/AbstractTestCase.java
+++ b/org.idempiere.test/src/org/idempiere/test/AbstractTestCase.java
@@ -225,6 +225,5 @@ public abstract class AbstractTestCase {
 	 * shutdown for class
 	 */
 	static void shutdown() {
-		Adempiere.stop();
 	}
 }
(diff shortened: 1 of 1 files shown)


org.idempiere.test/src/org/idempiere/test/AbstractTestCase.java
```diff
@@ -226,6 +226,5 @@ protected String getTrxName() {
 	 */
 	static void shutdown() {
 		Adempiere.stop();
```
hengsin, line 228: I think this can be removed too, seems just a waste of cpu time that we are calling startup for each test class.

hengsin (changes requested):

org.idempiere.test/src/org/idempiere/test/AbstractTestCase.java
CarlosRuiz-globalqss: Done.
In my eclipse tests the time spent decreased from 3:23 to 3:07

hengsin (approved):
looks good

PR #1316, opened 2022-05-05
IDEMPIERE-5284 Create method EMail.sendEx to have access to the full Exception when sending EMails (FHCA-3587)

CarlosRuiz-globalqss:
- improved also the dumpMessage, is just called when the level is finest, but it was logging on fine level, also the if must be out of the loop

https://idempiere.atlassian.net/browse/IDEMPIERE-5284

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/EMail.java
+++ b/org.adempiere.base/src/org/compiere/util/EMail.java
@@ -247,7 +247,32 @@ public final class EMail implements Serializable
 	 *	Send Mail direct
 	 *	@return OK or error message
 	 */
-	public String send ()
+	public String send()
+	{
+		String msg;
+		try {
+			msg = send(false);
+		} catch (Exception e) {
+			msg = e.getLocalizedMessage();
+		}
+		return msg;
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/util/EMail.java
```diff
@@ -560,13 +593,13 @@ public boolean isSentOK()
 	 */
 	private void dumpMessage()
 	{
-		if (m_msg == null)
+		if (m_msg == null || !log.isLoggable(Level.FINEST))
```
hengsin, line 596: this doesn't seems right, if m_msg is null, it will only trigger NPE in the code block below.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/util/EMail.java
CarlosRuiz-globalqss: > this doesn't seems right, if m_msg is null, it will only trigger NPE in the code block below.

The condition is OR.
If m_msg is null will execute the return below.

PR #1322, opened 2022-05-11
IDEMPIERE-5282 Memory Leak ZK Detached Dom Objects are cleared only a…

hengsin:
…fter ZK session is destroyed

https://idempiere.atlassian.net/browse/IDEMPIERE-5282

- Fix global variable pollution: use closure and replace var with let
- Fix memory leak of Grid and ListBox


org.adempiere.ui.zk/WEB-INF/src/web/js/org/idempiere/commons/report.js
```diff
@@ -3,38 +3,38 @@ if (typeof window.idempiere === 'undefined')
...
 window.idempiere.showColumnMenu = function(doc, e, columnName, row) {
-	var d = idempiere.getMenu (doc, e.target.getAttribute ("componentId"), e.target.getAttribute ("foreignColumnName"), e.target.getAttribute ("value"));
+	let d = idempiere.getMenu (doc, e.target.getAttribute ("componentId"), e.target.getAttribute ("foreignColumnName"), e.target.getAttribute ("value"));
 	
-	var posx = 0;
-	var posy = 0;
-	if (!e) var e = window.event;
+	let posx = 0;
+	let posy = 0;
+	if (!e) let e = window.event;
```
igorpojzl, line 37: I got warning about **let** here. Should this be:
`if (!e) e = window.event;` ?

igorpojzl (approved):
Ok Looks good. Validated in production deployment.

hengsin:
tested by @igorpojzl

PR #1323, opened 2022-05-11
IDEMPIERE-5262 Implement readonly protection for DB.getSQLValueEx call

hengsin:
- Avoid unnecessary commit call.


CarlosRuiz-globalqss (approved):
tested

PR #1324, opened 2022-05-11

CarlosRuiz-globalqss:
Found also two minor performance improvements:
- avoid counting users when statistics not allowed
- change the query where to 1=0, better for the optimizer


hengsin (approved):
looks good

PR #1325, opened 2022-05-11
IDEMPIERE-5267 Make Cache Max Size configurable via system property

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5267

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/CCache.java
+++ b/org.adempiere.base/src/org/compiere/util/CCache.java
@@ -43,7 +43,7 @@ public class CCache<K,V> implements CacheInterface, Map<K, V>, Serializable
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = -2268565219001179841L;
+	private static final long serialVersionUID = 4960404895430292476L;
 
 	protected Map<K, V> cache = null;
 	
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/util/CCache.java
```diff
@@ -130,7 +155,11 @@ public CCache (String tableName, String name, int initialCapacity, int expireMin
 		m_name = name;
 		m_tableName = tableName;
 		setExpireMinutes(expireMinutes);
-		m_maxSize = maxSize; 
+		int propMaxSize = getCacheMaxSize(name);
+		if (propMaxSize >= 0)
+			m_maxSize = propMaxSize;
+		else
+			m_maxSize = maxSize;
```
hengsin, line 162: instead of ignoring the maxSize method parameter, I think we should override `public CCache (String tableName, String name, int initialCapacity, int expireMinutes, boolean distributed)` instead and uses either maxSize from getCacheMaxSize(name) or CacheMgt.MAX_SIZE (if getCacheMaxSize(name) return -1).

hengsin (changes requested):

org.adempiere.base/src/org/compiere/util/CCache.java
CarlosRuiz-globalqss: Hi @hengsin - I don't understand your suggestion, can you please clarify?

Something good about the suggested approach here is that we can change **any** cache size without needing to change the code to use a different method.

hengsin (approved):
looks good

PR #1329, opened 2022-05-13
IDEMPIERE-4269 shipment (on availability) should be created only if products are available

lHeidbreder:
These are the changes for the - now quite old - ticket #4269 concerning products in transit or production


org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
```diff
@@ -290,7 +291,17 @@ private String generate (PreparedStatement pstmt)
 							logInfo.append(" (set to 0)");
 						}
 						//	Adjust On Hand
-						onHand = onHand.subtract(unconfirmedShippedQty);
+						StringBuilder where3 = new StringBuilder ("EXISTS (SELECT * FROM M_InOut io WHERE io.M_InOut_ID=M_InOutLine.M_InOut_ID AND io.IsSOTrx = 'Y' AND io.DocStatus IN ('DR','IN','IP','WC')")
```
hengsin, line 294: Shouldn't include DR and IN

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
```diff
@@ -290,7 +291,17 @@ private String generate (PreparedStatement pstmt)
...
+							    .append	(" AND io.M_Warehouse_ID=").append(p_M_Warehouse_ID).append (")");;
+						iols = MInOutLine.getOfProduct(getCtx(), 
+							line.getM_Product_ID(), where3.toString(), get_TrxName());
+						for (int j = 0; j < iols.length; j++) 
+							// don't count lines of our m_shipment, they are already decreased in storage
+							if (m_shipment == null || iols[j].get_ValueAsInt("M_InOut_ID") != m_shipment.get_ID())
+								totalunconfirmedShippedQty = totalunconfirmedShippedQty.add(iols[j].getMovementQty());
+						logInfo = new StringBuilder("TotalUnconfirmed Qty=").append(totalunconfirmedShippedQty)
```
hengsin, line 302: this is inefficient as you can simply do a sum of movementqty for all m_inout of same product, warehouse and id != current shipment id.

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
```diff
@@ -173,7 +173,7 @@ protected String doIt () throws Exception
 			if (p_C_BPartner_ID != 0)
 				m_sql.append(" AND o.C_BPartner_ID=?");					//	#3
 		}
-		m_sql.append(" ORDER BY M_Warehouse_ID, PriorityRule, M_Shipper_ID, C_BPartner_ID, C_BPartner_Location_ID, C_Order_ID");
+		m_sql.append(" ORDER BY M_Warehouse_ID, PriorityRule, M_Shipper_ID, DateOrdered, C_BPartner_ID, C_BPartner_Location_ID, C_Order_ID");
```
hengsin, line 176: this change is not explain in the ticket. if it is for a separate issue, create another ticket for it.

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
mschnbeck: Hi @hengsin,
when we created this patch two years ago we took a look at the other statements dealing with unconfirmed inouts and took the same list of doc states. Perhaps both states are wrong there, too. Can you have a look at it?

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
mschnbeck: you are right, it was another patch. @hengsin, what do you think: will it be a generally accepted rule to take orders with an earlier date a higher priority than the ordering by the bpartner ID? Then we'll create another ticket.

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
lHeidbreder: You are absolutely correct. If I understood you corretly, this is closer to what you'd have in mind?

hengsin: yes, this is better but I think we need to verify that the generate sql is indeed efficient (i.e cost analysis through pgadmin).

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
hengsin: hi @mschnbeck , can you point me where to check for that ?

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
hengsin: 2 seq scan and that's going to be slow for big db, need new index.

hengsin: 2 seq scan and that's going to be slow for big db, need new index.

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
hengsin: hi @mschnbeck , my mistake, the status clause is fine. However, with your changes, the existing unconfirmedShippedQty (and the query to get it) is make redundant and should be removed from the code.

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
lHeidbreder: Hi @hengsin, I have to apologize, I made a mistake. For my test I used a development database containing next to no entries. After retrying the test on a larger database, it looked like this:
grafik

If you still see the need for an additional index, I'd like some hints on what these indexes should encompass. Thank you for your patience

org.adempiere.base.process/src/org/compiere/process/InOutGenerate.java
mschnbeck: Hi @hengsin, I don't think that the unconfirmedShippedQty is redundant. This is the quantity we already have on shipments not yet completed for _this_ order and therefore don't have to create a shipment again. Thus to calculate how much remains to be shipped. The totalunconfirmedShippedQty however is the quantity which seems to be stocked but really isn't available anymore, because it's on not yet completed shipments for this or other orders.

What could be optmized is, not to look for the totalunconfirmedShippedQty if the toDeliver is already zero or set to zero.

hengsin:
[4269_1329.patch.txt](https://github.com/idempiere/idempiere/files/8859336/4269_1329.patch.txt)

hi @lHeidbreder , please merge the patch above that fix NPE cause by "append(m_shipment.get_ID())".

hengsin:
There's another issue with the this PR, the IsUnconfirmedInOut parameter, going by current label, tooltip and existing treatment in code, it is means to decide whether to generate shipment line for order line with in progress shipment line and it is not means to decide the way to calculate on hand quantity.  To remedy this, you would have to either change the label & tooltip text of IsUnconfirmedInOut or add a new process parameter for it.

hengsin:
Hi @CarlosRuiz-globalqss ,

Do you see value to add a flag to the Generate Shipment process that say ""Subtract In Progress Shipment From On Hand" ?
If yes, we can adapt this PR to that.

Otherwise, I think we have to close this PR as I think it is not appropriate to attach the "Subtract In Progress Shipment From On Hand" behavior the existing IsUnconfirmedInOut flag.

Regards,
Low

mschnbeck:
Hi @hengsin, hi @CarlosRuiz-globalqss,

in my opinion closing this PR is no option because the current behaviour simply is wrong. I can't guess a situation where it is a good idea to include stock which is already on the road to another customer when calculating the available stock. Of course it should not be dependent on the IsUnconfirmedInOut flag because this one is there to help when a newly created InOut should include older InOut which are not already delivered for _this_ customer.

So the intention is not to introduce a new flag then the behaviour should be standard, because it's an error correction.

Regards,
Martin

CarlosRuiz-globalqss:
Moving the functional conversation to the ticket ...

lHeidbreder:
Hi @hengsin,
all changes have been tied to a new parameter, as per our discussion on Jira. We would like a decision on whether you want to accept the changes like this at all, as customers of ours - the main reason we are bringing these changes forth - are waiting to finally migrate to the latest version of iDempiere

Regards,
Lukas

CarlosRuiz-globalqss:
> Do you see value to add a flag to the Generate Shipment process that say ""Subtract In Progress Shipment From On Hand" ?
> If yes, we can adapt this PR to that.

Hi @hengsin - as this has been done a preference with a parameter I think there is no problem to add it to core - it's adding good value.

hengsin (changes requested):
[1329_subtractonhand_patch.txt](https://github.com/idempiere/idempiere/files/9010769/1329_subtractonhand_patch.txt)
- decouple p_SubtractOnHand from p_IsUnconfirmedInOut

PR #1331, opened 2022-05-17
IDEMPIERE-5291 - Process Parameter Panel - Hidden Parameters Should N…

PeterTakacs300:
…ot Take Empty Space

JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5291

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
@@ -145,6 +145,7 @@ public class ProcessParameterPanel extends Panel implements
 	private ArrayList<GridField> m_mFields = new ArrayList<GridField>();
 	private ArrayList<GridField> m_mFields2 = new ArrayList<GridField>();
 	private ArrayList<Space> m_separators = new ArrayList<Space>();
+	private ArrayList<Div> m_Divs = new ArrayList<Div>();
 	//
 	private Grid centerPanel = null;
 	private Map<String, List<Row>> fieldGroupContents = new HashMap<String, List<Row>>();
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -951,6 +954,8 @@ private void dynamicDisplay() {
 				}
 			} else if (editor.isVisible()) {
 				editor.setVisible(false);
+				m_Divs.get(i).getParent().setVisible(false);
```
hengsin, line 957: While this looks fine, it is a rather strange way of doing thing here. I means if you want to make Row (i.e Div.getParent()) visible/invisible, why don't you just keep a list of Row instead of Div ?

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
PeterTakacs300: True, first I couldn't make it work with Row, but I've just realised I was making the List of Row incorrectly. I see, what I need is already there..

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -951,6 +952,8 @@ private void dynamicDisplay() {
 				}
 			} else if (editor.isVisible()) {
 				editor.setVisible(false);
+				rowList.get(i).setVisible(false);
+
```
hengsin, line 956: you can't use the existing rowList variable for this - that's for field group and it will be made empty when field group change. you need to add a new instance variable to hold all rows.

hengsin (changes requested):

hengsin (approved):
looks good

PR #1333, opened 2022-05-17
IDEMPIERE-5289 Apply Packin from Folder is stopping when sending email throws error (FHCA-3601)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5289


hengsin (approved):
looks good

PR #1334, opened 2022-05-18
IDEMPIERE-5014 Improve efficiency of Query for large result sets - us…

hengsin:
…ing Stream

- remove iterable() and restore the POIterator base iterate()
implementation.


CarlosRuiz-globalqss (approved):
according to what was suggested in ticket

PR #1335, opened 2022-05-18
IDEMPIERE-2058 - ** New Report ** - improve for handling Forms too

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-2058

Installed fix new report - we assumed that to adding new Copy Report is better

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkReportViewer.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkReportViewer.java
@@ -35,6 +35,7 @@ import javax.servlet.http.HttpServletRequest;
 import org.adempiere.base.upload.IUploadService;
 import org.adempiere.exceptions.DBException;
 import org.adempiere.pdf.Document;
+import org.adempiere.util.Callback;
 import org.adempiere.util.ContextRunnable;
 import org.adempiere.webui.ClientInfo;
 import org.adempiere.webui.Extensions;
@@ -1146,12 +1147,8 @@ public class ZkReportViewer extends Window implements EventListener<Event>, ITab
 		if (   MRole.getDefault().isTableAccess(MPrintFormat.Table_ID, false) 
 			&& Boolean.TRUE.equals(MRole.getDefault().getWindowAccess(pfAD_Window_ID)))
 		{
-			StringBuffer sb = new StringBuffer("*** ").append(Msg.getMsg(Env.getCtx(), "Operations")).append(" ***");
-			KeyNamePair pp = new KeyNamePair(0, sb.toString());
-			comboReport.appendItem(pp.getName(), null);
-			
-			sb = new StringBuffer("** ").append(Msg.getMsg(Env.getCtx(), "NewReport")).append(" **");
-			pp = new KeyNamePair(-1, sb.toString());
+			StringBuffer sb = new StringBuffer("** ").append(Msg.getMsg(Env.getCtx(), "NewReport")).append(" **");
+			KeyNamePair pp = new KeyNamePair(-1, sb.toString());
 			comboReport.appendItem(pp.getName(), pp.getKey());
 			sb = new StringBuffer("** ").append(Msg.getMsg(m_ctx, "CopyReport")).append(" **");
 			pp = new KeyNamePair(-2, sb.toString());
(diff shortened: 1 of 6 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkReportViewer.java
```diff
@@ -1147,8 +1147,16 @@ private void fillComboReport(int AD_PrintFormat_ID)
 			&& Boolean.TRUE.equals(MRole.getDefault().getWindowAccess(pfAD_Window_ID)))
 		{
 			StringBuffer sb = new StringBuffer("** ").append(Msg.getMsg(Env.getCtx(), "NewReport")).append(" **");
-			KeyNamePair pp = new KeyNamePair(-1, sb.toString());
+			StringBuffer sb = new StringBuffer("*** ").append(Msg.getMsg(Env.getCtx(), "Operations")).append(" ***");
```
CarlosRuiz-globalqss, line 1150: Eclipse shows error here:
Duplicate local variable sb

PeterTakacs300: Yes, I missed a row when committing

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkReportViewer.java
```diff
@@ -1146,9 +1146,16 @@ private void fillComboReport(int AD_PrintFormat_ID)
 		if (   MRole.getDefault().isTableAccess(MPrintFormat.Table_ID, false) 
 			&& Boolean.TRUE.equals(MRole.getDefault().getWindowAccess(pfAD_Window_ID)))
 		{
-			StringBuffer sb = new StringBuffer("** ").append(Msg.getMsg(Env.getCtx(), "NewReport")).append(" **");
-			KeyNamePair pp = new KeyNamePair(-1, sb.toString());
+			StringBuffer sb = new StringBuffer("*** ").append(Msg.getMsg(Env.getCtx(), "Operations")).append(" ***");
```
CarlosRuiz-globalqss, line 1149: Hi, I think the "** Operations **" is unnecessary, looking at the UI it seems better without that extra line.  I also can imagine the users asking - what is that option and why it doesn nothing when I click on it?

If you agree, can you please include this patch that avoids adding that extra menu option:
[pr1335_PR.txt](https://github.com/idempiere/idempiere/files/8749144/pr1335_PR.txt)

Regards,

Carlos Ruiz

norbertbede: Thanks, the reason of operations was make more hard to accidentally click to the new report, which cause many orphaned formats.  

Another options is add new alert popup (cancel/confirm) - with message like "You selected function Copy Report. The copy report create brand new Print Format, you can modify, update, change. This is usually for improving reporting to get additional information views". 

This avoid new popup avoid accidental report creation (this happening frequently. The message sound must be enough explainable - this is something i dont need.  usually this is for admins or advanced users)

CarlosRuiz-globalqss: I'm not a fan of these dialogs:

message

We would be adding an extra click forever to these two functions, as we don't have the functionality "don't show it again" as in other software.

I would propose better this:

* Change the text of _New Report_ to _Create new report with all columns_
* Change the text of _Copy Report_ to _Create new report as a copy of current print format_
* If you think the _Copy Report_ option must be more commonly used than the _New Report_, then we change the order and put Copy before New

Like this:
proposal


Here is the patch on top of the revision before adding the dialogs:
[pr1335_PR2.txt](https://github.com/idempiere/idempiere/files/8754568/pr1335_PR2.txt)

Regards,

Carlos Ruiz

norbertbede: hi,

The main point of popup with confirmation is avoid make unwanted copy - which clone existing reports and make junk in print-format database. For implementor (like cloudempiere) makes problem in support process because we don't know given print format is serious planned/serious or not, or accidental copy. Previously we assumed the ** [Operations] ** menu item solved this a bit, but yesterday i proposed to change it to 100% solution prevent this effect - the reason why we added this confirm popup.

Regarding to user comfort we assume the ratio Print <> New/Copy report is 1000:1 - we are not lost any comfort here.

**Meanwhile We create another improvement - which probably argue more our popup change**
The next concept is asking for print format name. So user must enter new print format name to created new or copy. 

dialogprintformat


Note 1: it is same concept as delete and enter name for confirm 
Note 2: still the print format name must be changed trough print format wizard or print format window.

If not acceptable we keep it for our purpose and close PR with your suggestion, consider first please.

Norbert

CarlosRuiz-globalqss: @norbertbede - what if we move the two options in the list box to buttons in the toolbar, the buttons can be same as the windows buttons:

_New button_ to create a new report
_Copy button_ to copy the button on report

Also, I would suggest that we can implement an Event handler when pushing these buttons, we have event handlers on some of the buttons on windows, and that way is easy to add externally a dialog before processing a button.

Another advantage is that you can setup permissions (restrict per role) those buttons easily.

WDYT?

norbertbede: Hi @CarlosRuiz-globalqss  

my ideas
- toolbar:  for me is going out of context a bit from user perspective - however looks standard, not sure i like it while context.
- permission: good but +1 more task for administering - practically IMO anybody should create own printformat - soon printformat sharing and publication status resolve disturbing each others and allow make temp print formats - until implementing, and later publish it for targeted group (same idea like for saved queries, saved filters on window)
- event handler: OK

looks i need postpone this PR - need time to rethink it :) it started as simple commit from our code, but looks the solution has lot of opportunities and questions. 

Thanks for feedback.
Norbert

CarlosRuiz-globalqss: Anyways, we are very close, I think the only point of discussion is about the dialog, so maybe something easier, just add the event handler when clicking the option in the print format list - and you can implement the dialog as external plugin.

hengsin: Some suggestions:
- Perhaps "Create New Print Format" and "Duplicate Current Print Format" is closer to what the 2 entry suppose to do ?
- For the prompt, we can make it optional by tenant level ad_sysconfig entry
-  Alternatively, it would be nice if we can add support for "Don't ask again" or "Remember my choice"

norbertbede: The fact is to have 100 accident report against 0 per tenant is the most important. 
However we can add Sysconfig as @hengsin  suggesting
Don't ask again feature would be nice, but wrong ratio for time/outcome at this moment. 

@CarlosRuiz-globalqss if we will add sysconfig then is this acceptable for you ? 
Maybe would be better to create a new dialog type, to get more nice design.  (similar i suggest in new ticket https://idempiere.atlassian.net/browse/IDEMPIERE-5319) so moving around today UX standards.

Norbert

CarlosRuiz-globalqss: I'm OK with the SysConfig, but I think the EventHandler is more robust.

With the EventHandler you can do
- the confirmation dialog easily, 
- or even the new confirmation dialog as in IDEMPIERE-5319, 
- or ask for the print format name, if you write the EventHandler in a way that you can read a returned String, 
- or you can add logic, maybe some tenant wants to be excluded from this, or some role, or some user, or some print formats

So, IMO the effort of adding the SysConfig and the EventHandler are almost the same, but the benefits of EventHandler are a lot bigger.

PeterTakacs300: Hi @CarlosRuiz-globalqss, I've consulted this issue (about the EventHandler) with @igorpojzl too. We are not sure what kind of EventHandler do you mean here. 

Is it from _AbstractEventHandler_ - we thought, probably not, since the event topics you can register would not be useful here -, or from the classes that implement the EventHandler interface - they use _org.osgi.service.event_ as parameter in _handleEvent_ method. 

Could you provide some more specific info or examples, please?

norbertbede: hi, @CarlosRuiz-globalqss  please follow up to Peter, we would like to close this issue. thnks

CarlosRuiz-globalqss: Hi @PeterTakacs300 and @norbertbede - after testing again this, and I see you changed the confirmation for an asking for the print format name, that looks good.

Sorry I didn't see it before.

The problem I see now is that the askForInput doesn't have a Cancel button.

Maybe it would be good to implement a FDialog.askForInputWithCancel with a Callback that manages the Object Map.Entry<Boolean, Object>

Something like:
`public static void askForInputWithCancel(int windowNo, WEditor weditor, String adMessage, String title, final Callback<Map.Entry<Boolean, Object>> callback)`

This popup must show the cancel button, and return true when the OK button is pressed, false otherwise.  And the object from the editor (in this case the String) in the second parameter of the Map.

Sounds good?

PeterTakacs300: Hi @CarlosRuiz-globalqss, I tried to implement your suggestion in the latest commit.

CarlosRuiz-globalqss (approved):
tested succesfully

PR #1336, opened 2022-05-18
IDEMPIERE-3294 - check empty String parameter

PeterTakacs300:
Reopened JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-3294

We also found new use cases which should be fixed, and those are when process is starting from Scheduler, Workflow, Dashboard or Web Service

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MPInstance.java
+++ b/org.adempiere.base/src/org/compiere/model/MPInstance.java
@@ -160,6 +160,7 @@ public class MPInstance extends X_AD_PInstance
 		final String whereClause = "AD_Process_ID=?";
 		List <MProcessPara> list = new Query(getCtx(), MProcessPara.Table_Name, whereClause, get_TrxName())
 		.setParameters(getAD_Process_ID())
+		.setOnlyActiveRecords(true)
 		.setOrderBy("SeqNo")
 		.list();
 
--- a/org.adempiere.base/src/org/compiere/model/MPInstance.java
+++ b/org.adempiere.base/src/org/compiere/model/MPInstance.java
@@ -100,15 +100,6 @@ public class MPInstance extends X_AD_PInstance
 		setAD_User_ID(Env.getAD_User_ID(process.getCtx()));
 		if (!save())		//	need to save for parameters
 			throw new IllegalArgumentException ("Cannot Save");
-		//	Set Parameter Base Info
-		MProcessPara[] para = process.getParameters();
-		for (int i = 0; i < para.length; i++)
-		{
-			MPInstancePara pip = new MPInstancePara (this, para[i].getSeqNo());
-			pip.setParameterName(para[i].getColumnName());
-			pip.setInfo(para[i].getName());
-			pip.saveEx();
-		}
 	}	//	MPInstance
 
 	/**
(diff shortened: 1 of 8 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -654,6 +654,14 @@ public boolean saveParameters() {
 			//Save only parameters which are set
 			if((result == null) && (result2 == null))
 				continue;
+			if(result instanceof String) {
+				if((result2 == null) && (Util.isEmpty(String.valueOf(result))))
```
hengsin, line 658: this can be simplify to Util.isEmpty((String)result). Note that the (String) cast is safe here since it is inside the "instanceof String" block.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -654,6 +654,14 @@ public boolean saveParameters() {
...
+					continue;
+				else if((result2 instanceof String)
+						&& (Util.isEmpty(String.valueOf(result)) 
+						&& Util.isEmpty(String.valueOf(result2))))
+					continue;
```
hengsin, line 663: this whole block can be simplify to if (Util.isEmpty((String)result) && Util.isEmpty((String)result2)) continue.

hengsin: `if(result instanceof String) {
    if (Util.isEmpty((String)result) && Util.isEmpty((String)result2))
        continue`

hengsin: sorry, my mistake, the above is not correct since result2 can be null. should be `Util.isEmpty((String)result) && (result2 == null || Util.isEmpty((String)result2))` instead.

hengsin (changes requested):

PeterTakacs300:
We've found new use cases which should be fixed, and those are when process is starting from Scheduler, Workflow, Dashboard or Web Service

org.adempiere.base/src/org/compiere/model/MPInstance.java
```diff
@@ -100,15 +100,6 @@ public MPInstance (MProcess process, int Record_ID)
...
 			throw new IllegalArgumentException ("Cannot Save");
-		//	Set Parameter Base Info
-		MProcessPara[] para = process.getParameters();
-		for (int i = 0; i < para.length; i++)
-		{
-			MPInstancePara pip = new MPInstancePara (this, para[i].getSeqNo());
-			pip.setParameterName(para[i].getColumnName());
-			pip.setInfo(para[i].getName());
-			pip.saveEx();
-		}
```
hengsin, line 111: This change will break existing API contract.

PeterTakacs300:
thanks to @igorpojzl 's idea, after we reverted the changes in the constructor, we decided to replace it with the `MPInstance (Properties ctx, int AD_Process_ID, int Record_ID)` constructor - so the previous, `MPInstance (MProcess process, int Record_ID)` could be potentially marked as deprecated

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MProcess.java
```diff
@@ -340,7 +340,7 @@ public boolean processIt (ProcessInfo pi, Trx trx, boolean managedTrx)
 	{
 		if (pi.getAD_PInstance_ID() == 0)
 		{
-			MPInstance pInstance = new MPInstance (this, pi.getRecord_ID());
+			MPInstance pInstance = new MPInstance (getCtx(), this.getAD_Process_ID(), pi.getRecord_ID());
```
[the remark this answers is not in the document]
PeterTakacs300: this change ensures that no parameters will be created, when they are empty

org.adempiere.base/src/org/compiere/model/MPInstance.java
```diff
@@ -151,6 +151,24 @@ public MPInstancePara[] getParameters()
...
+	 * 	Get Process Parameters
+	 *	@return processParameters array
+	 */
+	public MProcessPara[] getProcessParameters()
+	{
+		final String whereClause = "AD_Process_ID=?";
+		List <MProcessPara> list = new Query(getCtx(), MProcessPara.Table_Name, whereClause, null)
+		.setParameters(getAD_Process_ID())
+		.setOrderBy("SeqNo")
+		.list();
```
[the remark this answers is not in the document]
hengsin: You need to add setOnlyActiveRecords(true) call to filter out the inactive records.

hengsin (changes requested):

hengsin (approved):
looks good

PR #1337, opened 2022-05-18
[IDEMPIERE-5014] Update Javadoc for new stream() method

kriegfrj:
Added notes about performance and resource handling vs `iterate()`.


hengsin (approved):
looks good.

PR #1342, opened 2022-05-24
IDEMPIERE-5300 ErrorMsg doesn't show full message + LifeTime Msg conf…

iDempierePatch:
https://idempiere.atlassian.net/browse/IDEMPIERE-5300

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/StatusBar.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/StatusBar.java
@@ -27,7 +27,9 @@ import org.adempiere.webui.component.Tabpanel;
 import org.adempiere.webui.component.Window;
 import org.adempiere.webui.util.ZKUpdateUtil;
 import org.adempiere.webui.window.WTextEditorDialog;
+import org.compiere.model.MSysConfig;
 import org.compiere.process.ProcessInfoLog;
+import org.compiere.util.Env;
 import org.zkoss.zk.ui.Component;
 import org.zkoss.zk.ui.Execution;
 import org.zkoss.zk.ui.Executions;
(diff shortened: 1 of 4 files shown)


iDempierePatch:
Error Message - The popup exposes the complete message and life time can be configured through system variable configurator (ErrorMsgLifeTime)  
If the value equals '0', then the popup will not disappear and you will need to close it manually; otherwise set the variable as milleseconds

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/StatusBar.java
```diff
@@ -173,11 +175,15 @@ public void setStatusLine (String text, boolean error, ProcessInfoLog[] m_logs)
 			return;
 		
     	String labelText = buildLabelText(m_statusText);
+    	//iDempiereConsulting __05/10/2021 --- Error Message - The popup exposes the complete message and life time can be configured through system variable configurator (ErrorMsgLifeTime)  
+    	// If the value equals '0', then the popup will not disappear and you will need to close it manually; otherwise set the variable as milleseconds
+    	int duration = Integer.parseInt(MSysConfig.getValue("ErrorMsgLifeTime", "4500", Env.getAD_Client_ID(Env.getCtx())));   //MSysConfig.getIntValue("ErrorMsgLifeTime", 3500);
```
hengsin, line 180: for zk sysconfig, the common convention is to use all cap, separate by _ and starts with ZK_. Also, add the constant to MSysConfig (note existing constant is alphabetically order) instead of using the string literal here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/StatusBar.java
```diff
@@ -173,11 +175,15 @@ public void setStatusLine (String text, boolean error, ProcessInfoLog[] m_logs)
...
     	if (error) {
-    		Notification.show(buildNotificationText(m_statusText), "error", findTabpanel(this), "top_left", 3500, true);
+    		Notification.show(m_statusText, "error", findTabpanel(this), "top_left", duration, true);
```
hengsin, line 182: why replace "buildNotificationText(m_statusText)" with "m_statusText" ? this is not explain in the ticket.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/StatusBar.java
hengsin: And please remove the "//MSysConfig.getIntValue("ErrorMsgLifeTime", 3500);" comment.

hengsin: And you need to create a tenant level, default ad_sysconfig record (with official id) with value equal to the current hard coded duration.

hengsin (changes requested):

CarlosRuiz-globalqss:
Hi @hengsin - I implemented your suggestions here

hengsin (approved):
looks good

PR #1343, opened 2022-05-25
IDEMPIERE-4523 - Process Audit not mark AD_PInstance.IsProcessing on simple Reports

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4523

### DIFF
--- a/org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java
+++ b/org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java
@@ -167,14 +167,15 @@ public abstract class AbstractProcessCtl implements Runnable
 		 */
 		if (AD_Workflow_ID > 0)	
 		{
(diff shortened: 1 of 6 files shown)


org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java
```diff
@@ -172,8 +172,15 @@ public void run ()
...
+			pinstance.setIsProcessing(true);
 			pinstance.saveEx();
-			unlock();
+			try {
+				unlock();
+			}
+			finally {
+				pinstance.setIsProcessing(false);
+				pinstance.saveEx();
+			}
```
hengsin, line 183: this is not correct, isprocessing should be set before the start of workflow.

hengsin (changes requested):

hengsin (approved):
tested

PR #1345, opened 2022-05-26
IDEMPIERE-5302 - Drill Assistant Feature

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5302
Author: @igorpojzl


org.adempiere.base.callout/src/org/cloudempiere/model/CalloutProcessDrillRule.java
```diff
@@ -0,0 +1,48 @@
+package org.cloudempiere.model;
```
hengsin, line 1: missing license header

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/WProcessParameter.java
```diff
@@ -14,11 +14,14 @@
 
 package org.adempiere.webui.apps.form;
```
hengsin, line 15: missing license header

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkReportViewer.java
```diff
@@ -825,14 +828,14 @@ public void onEvent(Event event) throws Exception {
...
-							executeDrill(query, event.getTarget());
-						}
+					if (de.getData() != null && de.getData() instanceof DrillData) {
+						DrillData data = (DrillData) de.getData();
+//						Listitem item = comboDrill.getSelectedItem();
+//						if (item != null && item.getValue() != null && item.toString().trim().length() > 0)
+//						{
+//							query.setTableName(item.getValue().toString());
+							executeDrill(data, event.getTarget());
+//						}
```
hengsin, line 838: please remove code that's no longer needed instead of comment it out.

org.adempiere.ui/src/org/compiere/print/DrillReportCtl.java
```diff
@@ -0,0 +1,789 @@
+package org.compiere.print;
```
hengsin, line 1: missing license header

org.adempiere.ui/src/org/compiere/print/DrillReportCtl.java
```diff
@@ -0,0 +1,789 @@
...
+	public HashMap<Integer, KeyNamePair[]> getDrillTablePrintFormatMap() {
+		return drillTablePrintFormatMap;
+	}
+
+
+	public HashMap<Integer, KeyNamePair[]> getDrillProcessRulesPrintFormatMap() {
+		return drillProcessRulesPrintFormatMap;
+	}
+
+}
```
hengsin, line 789: again, two general rule: 1. javadoc for public constructor and method. 2. remove instead of comments out code.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/form/WProcessParameter.java
```diff
@@ -54,59 +57,76 @@ public ADForm getForm() {
...
 						}
 						if (!Util.isEmpty(para.getInfo_To())) {
-							String s = schedulerPara.getDescription();
+							String s = poPara.get_ValueAsString("Description");
 							if (Util.isEmpty(s))
 								s = para.getInfo_To();
 							else
 								s = s + ", " + para.getInfo_To();
-							schedulerPara.setDescription(s);
+							poPara.set_ValueOfColumn("Description", s);
```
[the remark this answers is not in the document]
PeterTakacs300: I'm not sure how do you mean this, since poPara can be an instance of MSchedulerPara or MProcessDrillRulePara here. 
Perhaps there could be checked, if the given column exists?

org.adempiere.base/src/org/compiere/model/MProcessDrillRulePara.java
```diff
@@ -0,0 +1,62 @@
+package org.compiere.model;
```
hengsin, line 1: missing license header

org.adempiere.base/src/org/compiere/model/MProcessDrillRule.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.sql.ResultSet;
+import java.util.List;
+import java.util.Properties;
+
+import org.compiere.util.Env;
+import org.idempiere.cache.ImmutableIntPOCache;
+import org.idempiere.cache.ImmutablePOSupport;
+
+public class MProcessDrillRule extends X_AD_Process_DrillRule implements ImmutablePOSupport {
```
hengsin, line 11: missing license header and javadoc

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/WDrillReport.java
```diff
@@ -0,0 +1,541 @@
...
+
+		Tabbox tabbox = new Tabbox();
+		Tabs tabs = new Tabs();
+		tabbox.appendChild(tabs);
+		Tabpanels tabpanels = new Tabpanels();
+		tabbox.appendChild(tabpanels);
+
+		Tab tab = new Tab(Msg.translate(Env.getCtx(), "AD_Process_DrillRule_ID"));
+		tabs.appendChild(tab);
+		tab = new Tab(Msg.translate(Env.getCtx(), "AD_Table_ID"));
```
hengsin, line 204: For translation from AD_Element, it is more efficient to use Msg.getElement(ctx, element) instead.

hengsin (changes requested):

hengsin:
I think the better approach is to launch the related report/process through process or process modal dialog instead of through the use reportctl api.

The dialog apporach will give the following benefits over the current PR:
- Simpler to support both silent and show help mode.
- Works for both process and report.
- Allow user to change process parameter and output format.
- Probably less duplicated code.

norbertbede:
@hengsin pls take a look, we did what you suggested above. hope we are near to final version.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessModalDialog.java
```diff
@@ -74,14 +74,23 @@ public ProcessModalDialog(int WindowNo, ProcessInfo pi, boolean autoStart)
 	 * @param WindowNo
 	 * @param pi
 	 */
-	public ProcessModalDialog(EventListener<Event> listener, int WindowNo, ProcessInfo pi)
+	public ProcessModalDialog(EventListener<Event> listener, int WindowNo, boolean isReRun, ProcessInfo pi)
```
hengsin, line 77: Please update javadoc and keep the existing public constructor so that it doesn't break backward compatibility (i.e add public ProcessModalDialog(EventListener<Event> listener, int WindowNo, ProcessInfo pi) {  this(listener, WindowNo, true, pi); } ). When changing public method or constructor, if it is simple to maintain existing source and binary compatibility, we would want to keep it.

hengsin (changes requested):

hengsin (changes requested):
1. Please remove all unuse import, variables and methods (You can check that through the Warnings show at the Problems view in Eclipse).
2. WDrillReport (Report Dialog):
    - Doesn't work if no print format have been created for Tenant yet (For e.g, Order Detail report in a new iDempiere database). This shouldn't be the case.
    - The dialog is not resizable and not maximizable.
    - For drill assistant open from column (for e.g Product column in Shipment Detail report), the "Table" tab is a very long list and took a long while to load. I don't think it is very useful and it is kind of frustrating (due to the wait) when all I'm interested at is the Drill Rule. If this is needed for the double click drill option, it must be lazy loaded.
    - The drill rule tab looks bad when there's no drill rule define. If we can't hide the "Drill Assitant" link when there's no drill rule define for a column, we should show an empty message (for e.g, No drill rule have been setup for column).
3. Double Click Drill:
    - It open the WDrillReport dialog and ignore the selected "Drill:" drop down item at toolbar.
    - That feel rather confusing. We should either keep the old double click drill behavior or remove the "Drill:" drop down from toolbar.
4. Drill Rule window:
    - Bug: the table field value sometime change to the table of selected Process/Report after save . For e.g, sometime it change from M_Product to RV_OrderDetail_Order Detail after save.
    - Since the Column field is filter by table, it should be a Drop Down list instead of Search. It can't be a long list and Drop Down is more convenient to use than Search.

hengsin:
[pr_1345_patch.txt](https://github.com/idempiere/idempiere/files/8907598/pr_1345_patch.txt)

- Fix an error in migration/iD10/postgresql/202205251513_IDEMPIERE-5302.sql

hengsin:
Some additional UI issue and questions.

Drill Rule window:
- Column should default to primary key column of selected table.
- Question: is there use case where column is not the primary key column of selected table ?
- Report View and Print Format column: this is not cover in https://wiki.idempiere.org/en/NF10_DrillAssistant, how should it work ?

PeterTakacs300:
Drill Assistant - Table tab: The IsShowInDrillOptions Field on AD_Table will have default NO, it means given table will not be listed in Drill Table tab. This avoids the long list and performance degradation.

When there's no drill rule: We assume to be consistent it is better to always render Drill Assistant, that means we added new message.

hengsin:
Hi @PeterTakacs300 , the thing is depends on configuration, it CAN be a long list. It looks straightforward to implement lazy loading for the "Table" tab and I see no reason why it shouldn't be implemented since it can be a potential usability issue.

org.adempiere.base/src/org/compiere/model/MProcessDrillRule.java
```diff
@@ -0,0 +1,212 @@
...
+
+	@Override
+	protected boolean beforeSave(boolean newRecord) {
+
+		if(newRecord || is_ValueChanged(MProcessDrillRule.COLUMNNAME_AD_Process_ID)) {
+			MProcess process = MProcess.get(getAD_Process_ID());
+			if(process != null && process.getAD_ReportView_ID() > 0) {
+				MReportView reportView = MReportView.get(process.getAD_ReportView_ID());
+				if(reportView != null && reportView.getAD_Table_ID() > 0)
+					setAD_Table_ID(reportView.getAD_Table_ID());
```
hengsin, line 176: Why are we setting the table field to the ad_table_id of report view here (same thing in the callout code above) ? This wasn't the case with https://wiki.idempiere.org/en/NF10_DrillAssistant where table field is set to M_Product. Please clarify if the example given in the wiki page is wrong and there's a different prefer usage flow here.

org.adempiere.base.callout/src/org/cloudempiere/model/CalloutProcessDrillRule.java
```diff
@@ -0,0 +1,90 @@
...
+		mTab.setValue(MProcessDrillRule.COLUMNNAME_AD_Table_ID, reportView.getAD_Table_ID());
+		
+		// set AD_Column_ID
+		MTable table = MTable.get(reportView.getAD_Table_ID());
+		String[] keyColumns = table.getKeyColumns();
+		MColumn column = null;
+		
+		if(keyColumns.length > 0) {
+			column = MColumn.get(ctx, table.getTableName(), table.getKeyColumns()[0]);
+			mTab.setValue(MProcessDrillRule.COLUMNNAME_AD_Column_ID, column.getAD_Column_ID());
```
hengsin, line 85: shouldn't this happens after user select a table instead ? please clarify what's the prefer usage flow of the window here.

PeterTakacs300:
> WDrillReport
> 
> * Doesn't work if no print format have been created for Tenant yet (For e.g, Order Detail report in a new iDempiere database). I don't see this have been addressed in the latest commit.

Hi @hengsin, changes related to this are implemented in WDrillRule in lines 320-352.

norbertbede:
Wiki review done: https://wiki.idempiere.org/en/NF10_DrillAssistant

org.adempiere.ui/src/org/compiere/print/DrillReportCtl.java
```diff
@@ -0,0 +1,789 @@
...
+	 * 	Run Process or Report
+	 *	@param process process
+	 *	@return summary
+	 *	@throws Exception
+	 */
+	protected ProcessInfo prepareProcessInfo(MProcess process,MProcessDrillRule processDrillRule, int ad_PrintFormat_ID) throws Exception
+	{
+		if (log.isLoggable(Level.INFO)) log.info(process.toString());
+
+		//	Process (see also MWFActivity.performWork
```
hengsin, line 459: this should be remove

org.adempiere.ui/src/org/compiere/print/DrillReportCtl.java
```diff
@@ -0,0 +1,789 @@
...
+		int AD_Table_ID = processDrillRule.getAD_Table_ID();
+		int Record_ID = 0;
+
+		ProcessInfo pi = new ProcessInfo (process.getName(), process.getAD_Process_ID(), AD_Table_ID, Record_ID);
```
hengsin, line 463: AD_Table_ID and Record_ID is not doing anything here, not needed.

hengsin (changes requested):
1. see new inline comments
2. please resolve the merge conflict in org.adempiere.ui.zk/WEB-INF/src/metainfo/zk/lang-addon.xml
3. setting isshowindrilloptions=N for all table is too broad a change from existing behavior. I would suggest to set it to Y for the following transaction tables: 'C_AllocationHdr','C_BankStatement','C_BankTransfer','C_Invoice','C_Order','C_Payment','C_Project','C_RfQ','GL_Journal','M_InOut','M_Inventory','M_Movement','M_Production','M_Requisition','M_RMA','R_Request'

hengsin:
Tested latest commit and found 2 new issue:

- NoDrillKeyParameterSet error message is not translated.
- The drill only work if the selected process/report has a process parameter that matches the drill column name. This requirement is not reflected at all in the Drill Rule setup window.

org.adempiere.base.callout/src/org/cloudempiere/model/CalloutProcessDrillRule.java
```diff
@@ -0,0 +1,90 @@
...
+		}
```
hengsin, line 86: don't think this is doing anything meaningful here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/DrillCommand.java
```diff
@@ -55,11 +55,12 @@ public boolean service(AuRequest request, boolean everError) {
 		String columnName = (String) data.get(0);
 		String tableName = MQuery.getZoomTableName(columnName);
 		String code = (String) data.get(1);
```
hengsin, line 57: for _ID column, it is better that code is integer instead of string. if you examine the query object for m_product_id, the where clause is m_product_id='130' instead of m_product_id=130.

hengsin:
[pr_1345_migration_patch.txt](https://github.com/idempiere/idempiere/files/9007549/pr_1345_migration_patch.txt)
- Fix migration script error (constraint not exists)

hengsin (changes requested):
There's one inconsistency in the Drill Assistant dialog:
- the dialog remains open after I've click on the Run Report link at "Drill Rule" tab but the behavior is different at the "Table" tab. At table tab, the dialog is dismiss after I've click on one of the report link.

hengsin (changes requested):
please revert the latest disposeParent commit which is totally unnecessary. WDrillReport is already a register event listener of ProcessModalDialog, you just need to add `if (DialogEvents.ON_WINDOW_CLOSE.equal(event.getName())) this.detach();` to the onEvent method of WDrillReport.

hengsin (changes requested):
[pr_1345_0630_patch.txt](https://github.com/idempiere/idempiere/files/9016504/pr_1345_0630_patch.txt)
please apply this patch and we should be good to go.

hengsin (approved):
tested

PR #1347, opened 2022-05-27
IDEMPIERE-5301 2022-05 Platform Update

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5301


CarlosRuiz-globalqss:
Hi @hengsin - my eclipse is 2021-12

When I imported the project, maven ... etc
The following error appears:
Plugin execution not covered by lifecycle configuration: org.eclipse.tycho:tycho-compiler-plugin:2.7.3:validate-classpath (execution: default-validate-classpath, phase: initialize)

I tried a clean materialization and same issue:
eclipseError

hengsin:
ok, will check, probably not compatible with 2021-12 . I'm using 2022-03.

hengsin:
Please use Eclipse 2022-03 instead.

CarlosRuiz-globalqss (approved):
* basic UI test passed (creating a POS order)
* all unit tests pass
* launched with idempiere-rest plugin and eclipse didn't complain

So, I think this is good to go.

The only issue is that it seems that it will create some headaches for all iDempiere developers.

Most developers maintain old versions of iDempiere, for example after running mvn on master, running mvn on release-9 shows this error:
```
[ERROR] Internal error: java.lang.IllegalArgumentException: bundleLocation not found: /home/carlos/.m2/repository/org/eclipse/orbit/bundles/org.junit/4.13.2-SNAPSHOT/org.junit-4.13.2-SNAPSHOT.jar -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalArgumentException: bundleLocation not found: /home/carlos/.m2/repository/org/eclipse/orbit/bundles/org.junit/4.13.2-SNAPSHOT/org.junit-4.13.2-SNAPSHOT.jar
```

I would suggest to document how to avoid this issue before pushing this ahead, and notify developers about the upcoming change, they need to use eclipse 2022-03 and manage the recommendation to fix the mvn issue on old projects.

PR #1348, opened 2022-05-27
IDEMPIERE-5093 Scheduler cron pattern scheduling is always using serv…

hengsin:
…er time zone

- Fix unit test.
- Fix server monitor error out if AD_ClientInfo.TimeZone has invalid
value.

https://idempiere.atlassian.net/browse/IDEMPIERE-5093


CarlosRuiz-globalqss (approved):
looks good

PR #1349, opened 2022-05-30
IDEMPIERE-5286 Fix ConcurrentModificationException in MPrintFormat.getAllItems Method

anozimada:
https://idempiere.atlassian.net/browse/IDEMPIERE-5286

note: I use different approach from the ticket author's suggestion.


hengsin (approved):
looks good

PR #1351, opened 2022-05-31
IDEMPIERE-4661 Fix Password fields show double context "Field Suggestions"

anozimada:
https://idempiere.atlassian.net/browse/IDEMPIERE-4661


CarlosRuiz-globalqss (approved):
tested

PR #1352, opened 2022-06-01
IDEMPIERE-4825 Fix Navigation arrows not consistent on detail tab when in single row mode

anozimada:
https://idempiere.atlassian.net/browse/IDEMPIERE-4825?focusedCommentId=48358


hengsin (approved):
looks good

PR #1355, opened 2022-06-02
IDEMPIERE-5307 - Report Summary Functions Not Working Properly

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5307

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -964,25 +964,15 @@ public class DataEngine
 											valueString = DisplayType.getDateFormat(pdc.getDisplayType(), m_language).format(value);
 										if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 											valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
-										if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
-											Serializable groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
-													pdc.getAD_PrintFormatItem_ID(), functions[f]);
-											if(groupValue instanceof Timestamp)
-												groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
-											if(groupValue != null)
-												valueString += ": " + groupValue;
-										}
 										pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
 											valueString, DisplayType.String, false, pdc.isPageBreak(), pdc.getFormatPattern()));
 									}
 									else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
 									{
-										Serializable groupValue = m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
-												pdc.getAD_PrintFormatItem_ID(), functions[f]);
-										if(groupValue instanceof Timestamp)
-											groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
 										pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-												groupValue, PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), 
+												m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
+														pdc.getAD_PrintFormatItem_ID(), functions[f]), 
+												PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), 
 												false, pdc.isPageBreak(), pdc.getFormatPattern()));
 									}
 								}	//	 for all columns
@@ -1195,25 +1185,15 @@ public class DataEngine
 									valueString = DisplayType.getDateFormat(pdc.getDisplayType(), m_language).format(value);
 								if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 									valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
-								if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
-									Serializable groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
-											pdc.getAD_PrintFormatItem_ID(), functions[f]);
-									if(groupValue instanceof Timestamp)
-										groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
-									if(groupValue != null)
-										valueString += ": " + groupValue;
-								}
 								pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
 									valueString, DisplayType.String, pdc.getFormatPattern()));
 							}
 							else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
 							{
-								Serializable groupValue = m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
-										pdc.getAD_PrintFormatItem_ID(), functions[f]);
-								if(groupValue instanceof Timestamp)
-									groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
 								pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-										groupValue,	PrintDataFunction.getFunctionDisplayType(functions[f],
+										m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
+												pdc.getAD_PrintFormatItem_ID(), functions[f]),	
+										PrintDataFunction.getFunctionDisplayType(functions[f],
 											pdc.getDisplayType()),pdc.getFormatPattern()));
 							}
 						}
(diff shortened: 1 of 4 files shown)


org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1371,6 +1405,19 @@ else if (token.equals("LINE"))
...
+	
+	/**
+	 * Returns Timestamp in format MM/DD/YYYY
+	 * @param Timestamp t
+	 * @return
+	 */
+	private String formatTimestamp(Timestamp t) {
+		if(t != null) {
+			String sTimestamp = t.toString();
+			return sTimestamp = sTimestamp.substring(5, 7) + "/" + sTimestamp.substring(8, 10) + "/" + sTimestamp.substring(0, 4);
```
hengsin, line 1417: Wrong to hardcode the formatting here, please use DisplayType.getDateFormat for date or timestamp format instead.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1288,7 +1322,7 @@ private void printRunningTotal (PrintData pd, int levelNo, int rowNo)
 				else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), PrintDataFunction.F_SUM))
 				{
 					pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-						m_group.getValue(PrintDataGroup.TOTAL, pdc.getAD_PrintFormatItem_ID(), PrintDataFunction.F_SUM),
+						m_group.getValue(PrintDataGroup.TOTAL, pdc.getAD_PrintFormatItem_ID(), PrintDataFunction.F_SUM).toString(),
```
hengsin, line 1325: why the need to add toString() here ?

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1185,14 +1196,25 @@ else if (pdc.getDisplayType() == DisplayType.DateTime)
 									valueString = DisplayType.getDateFormat(pdc.getDisplayType(), m_language).format(value);
 								if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 									valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
+								if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
+									Object groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
+											pdc.getAD_PrintFormatItem_ID(), functions[f]);
+									if(groupValue instanceof Timestamp)
+										groupValue = formatTimestamp((Timestamp)groupValue);
```
hengsin, line 1203: I think should format with pdc.getDisplayType here (i.e it could be just date or date+time). same for other place that call formatTimestamp.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/print/DataEngine.java
PeterTakacs300: hello, I've added toString() there, because the third parameter of the PrintDataElement constructor is Serializable and m_group.getValue(...) returns Object

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -964,14 +964,25 @@ private void loadPrintData (PrintData pd, MPrintFormat format)
 											valueString = DisplayType.getDateFormat(pdc.getDisplayType(), m_language).format(value);
 										if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 											valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
+										if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
+											Object groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
+													pdc.getAD_PrintFormatItem_ID(), functions[f]);
+											if(groupValue instanceof Timestamp)
+												groupValue = DisplayType.getDateFormat().format((Timestamp)groupValue);
```
hengsin, line 971: should be DisplayType.getDateFormat(pdc.getDisplayType()) instead, same goes for other places that has similar call.

org.adempiere.base/src/org/compiere/print/DataEngine.java
hengsin: should change getValue(...) to returnSerializable instead (same goes for addValue and getFunctionValue). toString() doesn't sounds right here.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/print/DataEngine.java
PeterTakacs300: I see, thank you

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -964,15 +964,25 @@ private void loadPrintData (PrintData pd, MPrintFormat format)
 											valueString = DisplayType.getDateFormat(pdc.getDisplayType(), m_language).format(value);
 										if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 											valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
+										if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
+											Serializable groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
+													pdc.getAD_PrintFormatItem_ID(), functions[f]);
+											if(groupValue instanceof Timestamp)
+												groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
+											if(groupValue != null)
+												valueString += ": " + groupValue;
```
hengsin, line 973: what's the use case for this ? there's existing timestamp check at line 963.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -964,15 +964,25 @@ private void loadPrintData (PrintData pd, MPrintFormat format)
...
+										Serializable groupValue = m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
+												pdc.getAD_PrintFormatItem_ID(), functions[f]);
+										if(groupValue instanceof Timestamp)
+											groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
 										pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-											m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
-												pdc.getAD_PrintFormatItem_ID(), functions[f]), 
-											PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), 
+												groupValue, PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), 
 												false, pdc.isPageBreak(), pdc.getFormatPattern()));
```
hengsin, line 986: the change for this block doesn't looks correct, it ignore pdc.getFormatPattern() and break the existing PrintDataElement.getValueDisplay(Language) method.

hengsin, line 986: the change for this block doesn't looks correct, it ignore pdc.getFormatPattern() and break the existing PrintDataElement.getValueDisplay(Language) method.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1185,15 +1195,25 @@ else if (pdc.getDisplayType() == DisplayType.DateTime)
...
 								if (format.getTableFormat().isPrintFunctionSymbols())		//	Translate Sum, etc.
 									valueString	+= PrintDataFunction.getFunctionSymbol(functions[f]);
+								if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f])) {
+									Serializable groupValue = m_group.getValue(pdc.getAD_PrintFormatItem_ID(), 
+											pdc.getAD_PrintFormatItem_ID(), functions[f]);
+									if(groupValue instanceof Timestamp)
+										groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
+									if(groupValue != null)
+										valueString += ": " + groupValue;
+								}
```
hengsin, line 1205: same as my question above for line 966/973

hengsin, line 1205: same as my question above, why the need of this after the timestamp check above.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1185,15 +1195,25 @@ else if (pdc.getDisplayType() == DisplayType.DateTime)
...
+										pdc.getAD_PrintFormatItem_ID(), functions[f]);
+								if(groupValue instanceof Timestamp)
+									groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
 								pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-									m_group.getValue(group_pdc.getAD_PrintFormatItem_ID(), 
-										pdc.getAD_PrintFormatItem_ID(), functions[f]),
-									PrintDataFunction.getFunctionDisplayType(functions[f],
+										groupValue,	PrintDataFunction.getFunctionDisplayType(functions[f],
 											pdc.getDisplayType()),pdc.getFormatPattern()));
 							}
```
hengsin, line 1218: same as my comments above for line 976/986

hengsin, line 1218: same as my comments above for line 976/986

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1222,15 +1242,26 @@ else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
...
 					{
+						Serializable groupValue = m_group.getValue(PrintDataGroup.TOTAL, 
+								pdc.getAD_PrintFormatItem_ID(), functions[f]);
+						if(groupValue instanceof Timestamp)
+							groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
 						pd.addNode(new PrintDataElement(pdc.getAD_PrintFormatItem_ID(), pdc.getColumnName(),
-							m_group.getValue(PrintDataGroup.TOTAL, 
-								pdc.getAD_PrintFormatItem_ID(), functions[f]),
-							PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), pdc.getFormatPattern()));
+								groupValue, PrintDataFunction.getFunctionDisplayType(functions[f], pdc.getDisplayType()), pdc.getFormatPattern()));
```
hengsin, line 1264: same as my comments above for line 976/986

hengsin, line 1264: same as my comments above for line 976/986

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -1222,15 +1242,26 @@ else if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
...
 							name = PrintDataFunction.getFunctionSymbol(functions[f]);	//	Symbol
+ 						if (m_group.isFunctionColumn(pdc.getAD_PrintFormatItem_ID(), functions[f]))
+						{
+ 							Serializable groupValue = m_group.getValue(PrintDataGroup.TOTAL, 
+									pdc.getAD_PrintFormatItem_ID(), functions[f]);
+							if(groupValue instanceof Timestamp)
+								groupValue = DisplayType.getDateFormat(pdc.getDisplayType()).format((Timestamp)groupValue);
+							if(groupValue != null)
+								name += ": " + groupValue;
+						}
```
hengsin, line 1253: why special treatment for timestamp only ? wouldn't that create inconsistency here as it is only symbol here for other data type ?

hengsin, line 1253: why special treatment for timestamp only ? wouldn't that create inconsistency here ?

hengsin (changes requested):

PeterTakacs300:
After discussion with my colleagues I removed the code related to the use case, when we select a column as grouping column, and also select a function (e.g. sum) for the same column. (In this case we can see only the symbol, and can’t see the result of the function.) - we agreed that this use case is unnecessary.

Regarding the Timestamp, we taught, that it would be useful if functions MIN and MAX worked for Timestamps, since before this, the functions worked correctly only for numeric values.

hengsin (approved):
looks good

PR #1356, opened 2022-06-03
IDEMPIERE-4543 Fix Post-it doesn't save/update from HTML tab

anozimada:
https://idempiere.atlassian.net/browse/IDEMPIERE-4543


hengsin (approved):
looks good

PR #1360, opened 2022-06-09
IDEMPIERE-5311 - Zoom Across Doesn't Allow Zoom from Different Windows of the Same Table

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5311


CarlosRuiz-globalqss:
@PeterTakacs300 - I have not tested your described scenario, but according to the description I'm wondering if the following patch solves the problem and also avoids duplicates in case of the same window:
[IDEMPIERE-5311_suggest.txt](https://github.com/idempiere/idempiere/files/8892073/IDEMPIERE-5311_suggest.txt)

PeterTakacs300:
Yes, if my tests were correct, it does solve both the problems, thank you.

CarlosRuiz-globalqss (approved):
tested by @PeterTakacs300

PR #1362, opened 2022-06-11
IDEMPIERE-5314 Add support to create local target platform mirror

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5314


CarlosRuiz-globalqss (approved):
tested according to README.md
harmless

PR #1363, opened 2022-06-13
IDEMPIERE-5315 Mismatch in queries WWFActivity.getWhereActivities and DPActivitiesModel.getWorkflowCount

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5315

- Found that DPActivities is deprecated, so deprecated too the DPActivitiesModel
- Also found that DefaultDesktop had a dependency to show the total of DPActivities, refactored to do the same with the new DPDocumentStatus
- Moved the query to MWFActivity to use where required
- Refactored JDBC to Query


CarlosRuiz-globalqss:
Tested by @d-ruiz

PR #1371, opened 2022-06-21
IDEMPIERE-5319 - Delete Record with Confirmation Logic

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5319

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
@@ -62,6 +62,7 @@ import org.adempiere.webui.component.ZkCssHelper;
 import org.adempiere.webui.editor.IProcessButton;
 import org.adempiere.webui.editor.WButtonEditor;
 import org.adempiere.webui.editor.WEditor;
+import org.adempiere.webui.editor.WStringEditor;
 import org.adempiere.webui.event.ActionEvent;
 import org.adempiere.webui.event.ActionListener;
 import org.adempiere.webui.event.DialogEvents;
@@ -110,6 +111,7 @@ import org.compiere.process.DocAction;
 import org.compiere.process.ProcessInfo;
 import org.compiere.process.ProcessInfoLog;
 import org.compiere.process.ProcessInfoUtil;
+import org.compiere.tools.FileUtil;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
@@ -202,6 +204,8 @@ public abstract class AbstractADWindowContent extends AbstractUIPart implements 
 	private int adWindowId;
 
 	private MImage image;
+	
+	private String deleteConfirmationLogic;
 
 	/**
 	 * Quick Form Status bar
@@ -2756,29 +2760,64 @@ public abstract class AbstractADWindowContent extends AbstractUIPart implements 
 			onDeleteSelected(postCallback);
 			return;
 		}
-
-        FDialog.ask(curWindowNo, null, "DeleteRecord?", new Callback<Boolean>() {
-
-			@Override
-			public void onCallback(Boolean result)
-			{
-				if (result)
+		
+		deleteConfirmationLogic = adTabbox.getSelectedGridTab().getDeleteConfirmationLogic();
+		if(Util.isEmpty(deleteConfirmationLogic)) {
+			FDialog.ask(curWindowNo, null, "DeleteRecord?", new Callback<Boolean>() {
+				@Override
+				public void onCallback(Boolean result)
 				{
-		        	//error will be catch in the dataStatusChanged event
-		            boolean success = adTabbox.getSelectedGridTab().dataDelete();
-		            adTabbox.getSelectedGridTab().dataRefreshAll(true, true);
-		    		adTabbox.getSelectedGridTab().refreshParentTabs();
-		    		if (!success)
-		    			showLastWarning();
-
-		            adTabbox.getSelectedTabpanel().dynamicDisplay(0);
-		            focusToActivePanel();
-		            MRecentItem.publishChangedEvent(Env.getAD_User_ID(ctx));		            
+					if (result)
+					{
+			        	//error will be catch in the dataStatusChanged event
+			            boolean success = adTabbox.getSelectedGridTab().dataDelete();
+			            adTabbox.getSelectedGridTab().dataRefreshAll(true, true);
+			    		adTabbox.getSelectedGridTab().refreshParentTabs();
+			    		if (!success)
+			    			showLastWarning();
+	
+			            adTabbox.getSelectedTabpanel().dynamicDisplay(0);
+			            focusToActivePanel();
+			            MRecentItem.publishChangedEvent(Env.getAD_User_ID(ctx));		           
+					}
+					if (postCallback != null)
+						postCallback.onCallback(result);
 				}
-				if (postCallback != null)
-					postCallback.onCallback(result);
-	        }
-		});
+			});
+		}
+		else {
+			int tableID = adTabbox.getSelectedGridTab().getAD_Table_ID();
+			int recordID = adTabbox.getSelectedGridTab().getRecord_ID();
+			deleteConfirmationLogic = FileUtil.parseTitle(ctx, deleteConfirmationLogic, tableID, recordID, curWindowNo, null);
+			
+			WEditor editor = new WStringEditor();
+			editor.fillHorizontal();
+			
+			FDialog.askForInputDeleteConfirmation(curWindowNo, editor, "DeleteRecordWithConfirm?", new Object[] {deleteConfirmationLogic}, null, 
+					deleteConfirmationLogic, new Callback<Object>() {
+				@Override
+				public void onCallback(Object result)
+				{
+					if(!(result instanceof String))
+						return;
+						
+					if(result.equals(deleteConfirmationLogic)) {
+						boolean success = adTabbox.getSelectedGridTab().dataDelete();
+			            adTabbox.getSelectedGridTab().dataRefreshAll(true, true);
+			    		adTabbox.getSelectedGridTab().refreshParentTabs();
+			    		if (!success)
+			    			showLastWarning();
+	
+			            adTabbox.getSelectedTabpanel().dynamicDisplay(0);
+			            focusToActivePanel();
+			            MRecentItem.publishChangedEvent(Env.getAD_User_ID(ctx));
+			            
+					}
+					if (postCallback != null)
+						postCallback.onCallback(result.equals(deleteConfirmationLogic));
+				}
+			});		
+		}
     }
 
     // Elaine 2008/12/01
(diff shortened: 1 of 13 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
```diff
@@ -2756,29 +2760,64 @@ private void onDeleteCallback(final Callback<Boolean> postCallback)
...
+		else {
+			int tableID = adTabbox.getSelectedGridTab().getAD_Table_ID();
+			int recordID = adTabbox.getSelectedGridTab().getRecord_ID();
+			deleteConfirmationLogic = FileUtil.parseTitle(ctx, deleteConfirmationLogic, tableID, recordID, curWindowNo, null);
+			
+			WEditor editor = new WStringEditor();
+			editor.fillHorizontal();
+			
+			FDialog.askForInputDeleteConfirmation(curWindowNo, editor, "DeleteRecordWithConfirm?", new Object[] {deleteConfirmationLogic}, null, 
+					deleteConfirmationLogic, new Callback<Object>() {
```
hengsin, line 2797: if result can only be of string type, you should just use Callback<String> instead.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
```diff
@@ -507,9 +507,18 @@ public boolean isVisible()
...
+     */
     public void setBackground(boolean error)
     {
-
+    	if(error) {
+    		setFieldStyle("border-color: red;");
+    	}
+    	else {
+    		setFieldStyle("");
+    	}
```
hengsin, line 521: the prefer approach in zk is to use WrongValueException instead.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -294,6 +321,12 @@ public int show(String message, String title, int buttons, String icon, WEditor
 
 		if ((buttons & NO) != 0)
 			btnNo.setVisible(true);
+		
+		if ((buttons & DELETE) != 0)
+			btnDelete.setVisible(true);
+
+		if ((buttons & CANCEL_TEXT) != 0)
+			btnCancelText.setVisible(true);
```
hengsin, line 329: I'm skeptical that this is necessary. For me, "to proceed with delete" is still a question to the user and a standard yes, no, cancel should work just fine too.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -307,11 +340,20 @@ public int show(String message, String title, int buttons, String icon, WEditor
...
+					}
+				});
+			}
 		}
-
+		
 		this.setTitle(title);
 		this.setPosition("center");
-		this.setClosable(true);
+		this.setClosable(Util.isEmpty(correctInput));
```
hengsin, line 356: shouldn't close just works the same as cancel/no here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FDialog.java
```diff
@@ -423,6 +423,33 @@ public void onCallback(Object result) {
...
+    	Callback<Object> msgCallback = null;
+    	if (callback != null)
+    	{
+    		msgCallback = new Callback<Object>() {
+				@Override
+				public void onCallback(Object result) {
+					callback.onCallback(result);
+				}
+			};
+    	}
```
hengsin, line 448: don't think there's a use case here that callback can be null

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FDialog.java
```diff
@@ -423,6 +423,33 @@ public void onCallback(Object result) {
...
+     * Confirmation dialog before deleting the records. 
+     * @param windowNo
+     * @param weditor
+     * @param adMessage
+     * @param adMessageArgs
+     * @param title
+     * @param correctInput
+     * @param callback
+     */
+    public static void askForInputDeleteConfirmation(int windowNo, WEditor weditor, String adMessage, Object[] adMessageArgs, String title, String correctInput, final Callback<Object> callback)
```
hengsin, line 437: instead of passing correctInput, a more general approach is to create a validation interface instead. that should provide support for different type of editor too.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
PeterTakacs300: Our team leading by @norbertbede tries to make iDempiere UX more attractive and follow todays standards, this ticket/improvement was also intended to this. WrongValueException renders another popup, so the user must close it before he/she can correct the input value, this means one extra step compared to the case when we indicate wrong value with red borders of the input field. WrongValueException is the preferred approach in zk, but it doesn't fulfil our vision. 

Maybe a message should be rendered inline, below input field, this could be a compromise. I add a version of it in the next commit. We think rendering multiple popups is not user friendly

Screenshot 2022-06-22 at 14 16 36
Screenshot 2022-06-22 at 14 16 54
.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
PeterTakacs300: We think using the new "DELETE" and "CANCEL_TEXT" buttons would be more user friendly and would follow today's standards. This UI would be more clear to Users - e.g. they could know the purpose of the dialog (to delete records) without even reading the question.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
CarlosRuiz-globalqss: > WrongValueException renders another popup, so the user must close it before he/she can correct the input value, this means one extra step compared to the case when we indicate wrong value with red borders of the input field

This is not correct, the popup that is rendered doesn't need to be closed by the user, it is automatically closed when the field that triggers the problem is changed.

You can try for example clearing the "Role" field on login and clicking the OK button, the popup shows "Fill mandatory fields: Role" - but it disappears automatically when the role is filled.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
CarlosRuiz-globalqss: According to some comments from @norbertbede the intention of this is not to be used just on delete cases, this can be also used in case other dangerous operations in the system, for example before running a process that is dangerous.

Because of that I agree with Heng Sin a normal Yes/No dialog would suffice, and indeed that makes it consistent with the actual delete dialog that will appear in all the other windows that don't have a "Delete Confirmation Logic" configured.

I think this also suggests that the method FDialog.askForInputDeleteConfirmation must be changed to askForInputTextConfirmation (or something like that).


About text buttons, I think it's better to make the UI consistent, if you want text buttons, then it is better to implement a way in FDialog to manage text buttons everywhere (maybe a Tenant SysConfig?) - again here the user will see some dialogs with icon buttons deleting in some windows, and text buttons in other windows, IMO that's bad user experience.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/Messagebox.java
```diff
@@ -156,6 +170,17 @@ private void init()
...
 		btnNo.setId("btnNo");
 		
+		String cancelTextLabel = Util.cleanAmp(Msg.getMsg(ctx, "Cancel"));
+		btnCancelText = ButtonFactory.createButton(cancelTextLabel, null, cancelTextLabel);
+		btnCancelText.addEventListener(Events.ON_CLICK, this);
+		btnCancelText.setId("btnCancelText");
+		
+		String deleteLabel = Util.cleanAmp(Msg.getMsg(ctx, "delete"));
+		btnDelete = ButtonFactory.createButton(deleteLabel, null, deleteLabel);
+		btnDelete.setStyle("color: white; background: red;");
```
CarlosRuiz-globalqss, line 180: Setting the btnDelete and spanError in red hardcoded is not a good idea, it is better to do it in CSS.

But I think this comment would be obsolete, if we opt for the OK/Cancel button, then the Delete button is not necessary, and if we move to WrongValueException then the spanError is also unnecessary (making this comment useless).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
```diff
@@ -2756,29 +2760,65 @@ private void onDeleteCallback(final Callback<Boolean> postCallback)
...
-				if (postCallback != null)
-					postCallback.onCallback(result);
-	        }
-		});
+			});
+		}
+		else {
+			int tableID = adTabbox.getSelectedGridTab().getAD_Table_ID();
+			int recordID = adTabbox.getSelectedGridTab().getRecord_ID();
+			deleteConfirmationLogic = FileUtil.parseTitle(ctx, deleteConfirmationLogic, tableID, recordID, curWindowNo, null);
```
CarlosRuiz-globalqss, line 2791: Testing delete logic in Freight Category window:

I followed the three examples on the ticket description

1) ConfirmDeleteLogic = @Date@

-> didn't work, worked correctly in this way @#Date@

2) ConfirmDeleteLogic = @M_Product.Value@

-> didn't work, worked correctly in this way @Value@

3) ConfirmDeleteLogic = “@DeleteMe@“ - Dialog will ask for Zmazat in Slovak Language.

-> didn't work - I assume this can be a message, tested with @delete.selected@ and didn't work

PeterTakacs300:
Current state:

Screenshot 2022-06-24 at 15 54 58
Screenshot 2022-06-24 at 15 55 08

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
PeterTakacs300: Yes, in the first two use cases @#Date@ and @Value@ are correct, I added the 3rd use case in the latest commit, so @delete.selected@ should also work now.

norbertbede:
We decide for now as @CarlosRuiz-globalqss suggest buttons to stay consistent, we agree for is better approach.
However long term this button styling and tooltip must be reimplemented in new jira issue - allow change behaviour from button => text - same for tooltip. IMO this design is doesn't fit latest UX best practices.

CarlosRuiz-globalqss:
Hi @PeterTakacs300 - please solve the conflicts created in Messagebox after integrating IDEMPIERE-2058

CarlosRuiz-globalqss (approved):
tested successfully

PR #1375, opened 2022-06-27
IDEMPIERE-5331 Create Production from Order (all Lines)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5331

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/OrderCreateProduction.java
+++ b/org.adempiere.base.process/src/org/compiere/process/OrderCreateProduction.java
@@ -1,0 +1,164 @@
+/***********************************************************************
+ * This file is part of iDempiere ERP Open Source                      *
+ * http://www.idempiere.org                                            *
+ *                                                                     *
+ * Copyright (C) Contributors                                          *
+ *                                                                     *
+ * This program is free software; you can redistribute it and/or       *
+ * modify it under the terms of the GNU General Public License         *
+ * as published by the Free Software Foundation; either version 2      *
+ * of the License, or (at your option) any later version.              *
+ *                                                                     *
+ * This program is distributed in the hope that it will be useful,     *
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of      *
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the        *
+ * GNU General Public License for more details.                        *
+ *                                                                     *
+ * You should have received a copy of the GNU General Public License   *
+ * along with this program; if not, write to the Free Software         *
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,          *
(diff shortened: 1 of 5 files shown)


org.adempiere.base.process/src/org/compiere/process/OrderCreateProduction.java
```diff
@@ -0,0 +1,164 @@
...
+			production.setProductionQty(line.getQtyOrdered().subtract(line.getQtyDelivered()));
+			production.setDatePromised(line.getDatePromised());
+			if (product.getM_Locator_ID() > 0)
+				production.setM_Locator_ID(product.getM_Locator_ID());
+			production.setC_OrderLine_ID(line.getC_OrderLine_ID());
+
+			int locator = product.getM_Locator_ID();
+			if (locator == 0)
+				locator = MWarehouse.get(getCtx(), line.getM_Warehouse_ID()).getDefaultLocator().get_ID();
+			production.setM_Locator_ID(locator);
```
hengsin, line 108: this make line 101 & 102 above redundant

CarlosRuiz-globalqss: Thanks @hengsin - I copied the code from OrderLineCreateProduction and didn't notice the redundant assignment there.

So, with the last commit I refactored also that class and additionally removed the unused parameters

hengsin (changes requested):
1. Tested order with PChair. Production document created with 9 production line which is not correct.
2. Failed the following scenario
- Create and Complete Sales Order with PChair
- Run "Create Production from Order"
- Void the production document created.
- Goes back order, run "Create Production from Order Line". The process error out with "Production has already been created: xxx"

CarlosRuiz-globalqss:
Thanks @hengsin 

> Tested order with PChair. Production document created with 9 production line which is not correct.

This is not changed in this ticket, it's the actual behavior of create production, if you create a production with PChair manually, it's created with 9 lines, I see it's doing BOM explosion for the legs.
I think to discuss if that's correct or not - or if it must be made configurable, requires a different ticket.

>   Failed the following scenario
>    Create and Complete Sales Order with PChair
>    Run "Create Production from Order"
>    Void the production document created.
>    Goes back order, run "Create Production from Order Line". The process error out with "Production has already been created: xxx"

Also not introduced in this ticket, but is a good opportunity to solve it too.  Committed the fix for void and reverse.

hengsin:
Our treatment of phantom bom item is wrong, see https://blogs.sap.com/2008/04/24/what-is-a-phantom-item-how-is-it-used/

Anyway, that can be addressed in a separate ticket.

hengsin:
My mistake, that BOM explosion is actually correct.

hengsin (approved):
tested

PR #1376, opened 2022-06-28
IDEMPIERE-5332 Refactoring of Archive Viewer

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5332

### DIFF
--- a/org.adempiere.ui/src/org/compiere/apps/form/Archive.java
+++ b/org.adempiere.ui/src/org/compiere/apps/form/Archive.java
@@ -13,7 +13,13 @@
  *****************************************************************************/
 package org.compiere.apps.form;
 
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
 import java.util.logging.Level;
 
 import org.compiere.model.MArchive;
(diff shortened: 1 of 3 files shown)


org.adempiere.ui/src/org/compiere/apps/form/Archive.java
```diff
@@ -28,43 +34,66 @@
...
-//		int AD_Role_ID = Env.getAD_Role_ID(Env.getCtx());
 		
 		boolean trl = !Env.isBaseLanguage(Env.getCtx(), "AD_Process");
 		String lang = Env.getAD_Language(Env.getCtx());
-		// TODO: ASP - implement process and window access ASP control
 		String sql = "SELECT DISTINCT p.AD_Process_ID,"
 				+ (trl ? "trl.Name" : "p.Name ")
 			+ " FROM AD_Process p INNER JOIN AD_Process_Access pa ON (p.AD_Process_ID=pa.AD_Process_ID) "
 			+ (trl ? "LEFT JOIN AD_Process_Trl trl on (trl.AD_Process_ID=p.AD_Process_ID and trl.AD_Language=" + DB.TO_STRING(lang) + ")" : "") 
-			+ " WHERE "+role.getIncludedRolesWhereClause("pa.AD_Role_ID", null) // metas: use included roles
```
CarlosRuiz-globalqss, line 57: I have not tested yet - but removing the role.getIncludedRolesWhereClause from the where is not changing the access for users with included roles?

CarlosRuiz-globalqss: Thanks @hengsin - understood

Tests with Archive feature in the UI went fine.

What's the goal of replacing the role.getIncludedRolesWhereClause here?
The question is because I see the same code in getTableData, and that method is also used in cmd_query, so, wondering if is a performance improvement or the goal is something different

I'm approving this as all tests passed

CarlosRuiz-globalqss (approved):

PR #1377, opened 2022-06-28
IDEMPIERE-5322 - Document Rounding Rules

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5322


PeterTakacs300:
Implemented to C_Invoice and C_Order - there are known issues with C_Order Posting, so if it would be necessary we could fix that further.

PeterTakacs300:
We agree with the "charge based solution” according to the ticket-discussion.

PR #1380, opened 2022-06-29
IDEMPIERE-5322 Document Rounding Rules

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5322


CarlosRuiz-globalqss:
As mentioned in ticket, given the potential differences in needs and approach to solve the same problem, I think is better to have this as a plugin.  The existing plugin can be an initial seed to solve more complex cases.

PR #1392, opened 2022-07-12
IDEMPIERE-5342 : Ability to hide the title of the chart when displaye…

nmicoud:
…d on Home page

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/IChartRendererService.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/IChartRendererService.java
@@ -51,16 +51,6 @@ public interface IChartRendererService {
 	 * @param width
 	 * @param height
 	 * @param chartModel
-	 * @return true if render successfully
-	 */
-	public boolean renderChart(Component parent, int width, int height, ChartModel chartModel);
-
-	/**
-	 * render chart for AD_Chart
-	 * @param parent
-	 * @param width
-	 * @param height
-	 * @param chartModel
 	 * @param showTitle
 	 * @return true if render successfully
 	 */
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/IChartRendererService.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/IChartRendererService.java
@@ -54,4 +54,15 @@ public interface IChartRendererService {
 	 * @return true if render successfully
 	 */
 	public boolean renderChart(Component parent, int width, int height, ChartModel chartModel);
+
+	/**
+	 * render chart for AD_Chart
+	 * @param parent
+	 * @param width
+	 * @param height
+	 * @param chartModel
+	 * @param showTitle
+	 * @return true if render successfully
+	 */
+	public boolean renderChart(Component parent, int width, int height, ChartModel chartModel, boolean showTitle);
 }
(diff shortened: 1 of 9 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/IChartRendererService.java
```diff
@@ -54,4 +54,15 @@ public interface IChartRendererService {
...
+	/**
+	 * render chart for AD_Chart
+	 * @param parent
+	 * @param width
+	 * @param height
+	 * @param chartModel
+	 * @param showTitle
+	 * @return true if render successfully
+	 */
+	public boolean renderChart(Component parent, int width, int height, ChartModel chartModel, boolean showTitle);
```
hengsin, line 67: I did suggest using default to maintain backward compatibility. if we ignore backward compatibility, we should use default for the old method since it should always just delegate to the new method.

nmicoud: Hi,

I've added IChartRendererService.renderChart(Component, int, int, ChartModel, boolean) to preserve IChartRendererService.renderChart(Component, int, int, ChartModel) as it may be used by plugins.

nb: the ChartRendererServiceImpl.renderChart(Component, int, int, ChartModel) is not used anymore by the standard

Do you want to remove IChartRendererService.renderChart(Component, int, int, ChartModel) ?

If that's not what you expect, can you elaborate more please? (I think I'm lost here)

hengsin: to maintain backward binary compatibility:
replace 
`public boolean renderChart(Component parent, int width, int height, ChartModel chartModel, boolean showTitle)`
with
`default boolean renderChart(Component parent, int width, int height, ChartModel chartModel, boolean showTitle) {
return renderChart(Component parent, int width, int height, ChartModel chartModel);
}`
This approach allow existing plugin that only implements the `public boolean renderChart(Component parent, int width, int height, ChartModel chartModel)` method continue to work without any code changes or recompilation.

hengsin: alternatively, if we don't have to keep backward binary compatibility:
replace
`public boolean renderChart(Component parent, int width, int height, ChartModel chartModel);`
with
`default boolean renderChart(Component parent, int width, int height, ChartModel chartModel) {
return renderChart(parent, width, height, chartModel, true);
}`

With this, plugin only needs to override/implement the `public boolean renderChart(Component parent, int width, int height, ChartModel chartModel, boolean showTitle)` method.

nmicoud: OK, I understand now ; thanks for the explanation.
Rethinking, the backward compatibility is not needed as the changes will only impact v10 (and at the end, is just adding a parameter).
Does it sounds correct?

nb: I was not able to test it as I'm facing compilation issue (https://groups.google.com/g/idempiere/c/m7BcqplytoQ)

hengsin: Yes, not a must if it is for v10 only.

nmicoud: OK, so the last commit should resolve the issue

hengsin (approved):
tested

PR #1401, opened 2022-07-20
IDEMPIERE-4592 Bank Statement Loader zoom in the Load Bank Statement process throws an exception when the field is empty

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4592


d-ruiz (approved):
Tested succesfully

PR #1408, opened 2022-07-21
IDEMPIERE-5349: Adding AlwaysUpdatableLogic on Column and Field.

dpansheriya:
Added Always updatable logic on Field and column


migration/iD10/oracle/202207180103_IDEMPIERE-5349.sql
```diff
@@ -0,0 +1,185 @@
...
..., c.valuemax AS ValueMax, fgt.name AS fieldgroup, vr.code AS validationcode, f.included_tab_id AS Included_Tab_ID, fg.fieldgrouptype AS FieldGroupType, fg.iscollapsedbydefault AS IsCollapsedByDefault, COALESCE(f.infofactoryclass, c.infofactoryclass) AS InfoFactoryClass, c.isautocomplete AS IsAutocomplete, COALESCE(f.isallowcopy, c.isallowcopy) AS IsAllowCopy, f.isdisplayedgrid AS IsDisplayedGrid, f.seqnogrid AS SeqNoGrid, c.seqnoselection AS SeqNoSelection, f.xposition AS XPosition, f.columnspan AS ColumnSpan, f.numlines AS NumLines, COALESCE(f.istoolbarbutton, c.istoolbarbutton) AS IsToolbarButton, c.formatpattern AS FormatPattern, f.isadvancedfield AS IsAdvancedField, f.isdefaultfocus AS IsDefaultFocus, c.ad_chart_id AS AD_Chart_ID, f.ad_labelstyle_id AS AD_LabelStyle_ID, f.ad_fieldstyle_id AS AD_FieldStyle_ID, c.pa_dashboardcontent_id AS PA_DashboardContent_ID, COALESCE(trl.placeholder, ct.placeholder) AS Placeholder, c.ishtml AS IsHtml, f.isquickform AS IsQuickForm, COALESCE(f.ad_val_rule_lookup_id, c.ad_val_rule_lookup_id) AS AD_Val_Rule_Lookup_ID, vrf.code AS validationcodelookup, COALESCE(f.alwaysupdatablelogic, c.alwaysupdatablelogic) AS alwaysupdatablelogic FROM ad_field f
```
CarlosRuiz-globalqss, line 132: Lines with length > 2499 cause error SP2-0027 in oracle.  It must be splitted

org.adempiere.base/src/org/compiere/model/GridField.java
```diff
@@ -444,8 +444,30 @@ public boolean isEditable (Properties ctx, boolean checkContext,boolean isGrid)
 		}
 
 		//  Fields always updateable
-		if (m_vo.IsAlwaysUpdateable)      //  Zoom
-			return true;
+		if (m_vo.IsAlwaysUpdateable)
```
CarlosRuiz-globalqss, line 447: Bringing comment from @hengsin in ticket:

> AlwaysUpdatableLogic is only check and override IsAlwaysUpdateable when IsAlwaysUpdateable is true. This is inconsistent with how current code deals with IsReadonly and ReadonlyLogic. I don’t think such inconsistency is a good thing to have

CarlosRuiz-globalqss (changes requested):

migration/iD10/oracle/202207180103_IDEMPIERE-5349.sql
dpansheriya: This is resolved

org.adempiere.base/src/org/compiere/model/GridField.java
dpansheriya: This also resolved by making isAlwaysUpdatable flag has more preference than updatable logic

hengsin (changes requested):
hi @dpansheriya , you need to rename the migration script to current date and include the ad_infowindow_id column that we have added to ad_field_v & ad_field_vt.

dpansheriya:
> hi @dpansheriya , you need to rename the migration script to current date and include the ad_infowindow_id column that we have added to ad_field_v & ad_field_vt.

Please find latest commit. For Ad_field_v view, I think earlier approach where we have view definition committed and we were updating it was good approach. As we always getting merge issue so it was good to apply changes.

hengsin (approved):
tested

PR #1411, opened 2022-07-26
IDEMPIERE-5153 Cost Adjument Line Before Save: Update current cost price

hengsin:
- Fix always getting current cost from primary accounting schema.


hengsin:
tested by Michael Powacht (https://idempiere.atlassian.net/browse/IDEMPIERE-5153?focusedCommentId=48579)

PR #1412, opened 2022-07-26
IDEMPIERE-5356 - Add Printing Support to ProcessInfoDialog

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5356

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/ReportEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/ReportEngine.java
@@ -2403,6 +2403,28 @@ queued-job-count = 0  (class javax.print.attribute.standard.QueuedJobCount)
 	}
 
 	/**
+	 * Get DOC_TABLE_ID array
+	 * @return
+	 */
+	public static int[] getDocTableIDs() {
+		return DOC_TABLE_ID;
+	}
+	
+	/**
+	 * Get Report Engine Type from Table_ID
+	 * @param tableID
+	 * @return Report Engine Type 
+	 * -1 if Report Engine Type was not found
+	 */
+	public static int getReportEngineType(int tableID) {
+		for(int i = 0; i < DOC_TABLE_ID.length; i++) {
+			if(DOC_TABLE_ID[i] == tableID)
+				return i;
+		}
+		return -1;
+	}
+	
+	/**
 	 * build css for table from mapCssInfo
 	 * @param doc
 	 */
--- a/org.adempiere.base/src/org/compiere/print/ReportEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/ReportEngine.java
@@ -2401,14 +2401,6 @@ queued-job-count = 0  (class javax.print.attribute.standard.QueuedJobCount)
 	public void setIsReplaceTabContent(boolean m_isReplaceTabContent) {
 		this.m_isReplaceTabContent = m_isReplaceTabContent;
 	}
-
-	/**
-	 * Get DOC_TABLE_ID array
-	 * @return
-	 */
-	public static int[] getDocTableIDs() {
-		return DOC_TABLE_ID;
-	}
 	
 	/**
 	 * Get Report Engine Type from Table_ID
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
@@ -14,25 +14,53 @@
 
 package org.adempiere.webui.component;
 
+import static org.compiere.model.SystemIDs.PROCESS_C_INVOICE_GENERATERMA_MANUAL;
+import static org.compiere.model.SystemIDs.PROCESS_C_INVOICE_GENERATE_MANUAL;
+import static org.compiere.model.SystemIDs.PROCESS_M_INOUT_GENERATERMA_MANUAL;
+import static org.compiere.model.SystemIDs.PROCESS_M_INOUT_GENERATE_MANUAL;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_C_DUNNING;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_C_INVOICE;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_C_ORDER;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_C_PROJECT;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_C_RFQRESPONSE;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_DD_ORDER;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_M_INOUT;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_M_INVENTORY;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_M_MOVEMENT;
+import static org.compiere.model.SystemIDs.PROCESS_RPT_PP_ORDER;
+
+import java.io.File;
+import java.io.FileInputStream;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Level;
 
 import org.adempiere.webui.ISupportMask;
 import org.adempiere.webui.LayoutUtils;
+import org.adempiere.webui.apps.AEnv;
+import org.adempiere.webui.apps.form.WGenForm;
 import org.adempiere.webui.event.DialogEvents;
 import org.adempiere.webui.factory.ButtonFactory;
+import org.adempiere.webui.session.SessionManager;
 import org.adempiere.webui.util.ZKUpdateUtil;
+import org.adempiere.webui.window.SimplePDFViewer;
+import org.compiere.print.ReportEngine;
 import org.compiere.process.ProcessInfo;
 import org.compiere.process.ProcessInfoLog;
 import org.compiere.process.ProcessInfoUtil;
+import org.compiere.util.CLogger;
 import org.compiere.util.DisplayType;
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
 import org.zkoss.zhtml.Text;
+import org.zkoss.zk.au.out.AuEcho;
 import org.zkoss.zk.ui.Component;
 import org.zkoss.zk.ui.Executions;
 import org.zkoss.zk.ui.event.Event;
 import org.zkoss.zk.ui.event.EventListener;
 import org.zkoss.zk.ui.event.Events;
+import org.zkoss.zk.ui.util.Clients;
 import org.zkoss.zul.Hbox;
 import org.zkoss.zul.Image;
 import org.zkoss.zul.Separator;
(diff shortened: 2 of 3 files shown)


PeterTakacs300:
Hi @hengsin , I'm not sure how you ment _"printing should depends solely on whether the linked document is printable"_. For now, we tried similar approach as in ReportCtl.start method.

hengsin:
I means whether the linked document is with a document Report Engine Type. In other words, instead of hard coding the process info id, you check whether the ad_table_id of the process log record is associated with a document Report Engine Type (Shipment, Invoice, Order, etc). Another alternative is to add a report engine type field to the process log and use that to decide whether printing is available.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +245,64 @@ public void onEvent(Event event) throws Exception {
...
+	/**
+	 * Is Printable
+	 * @param pi
+	 * @return
+	 */
+	public boolean isPrintable(ProcessInfo pi) {
+		for(ProcessInfoLog log : m_logs)
+			for(int tableID : ReportEngine.getDocTableIDs())
+				if(log.getAD_Table_ID() == tableID) {
+					reportEngineType = ReportEngine.getReportEngineType(tableID);
```
hengsin, line 302: should check reportEngineType is valie (i.e not -1) here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +245,64 @@ public void onEvent(Event event) throws Exception {
...
+		// Loop through all items
+		List<File> pdfList = new ArrayList<File>();
+		for (int i = 0; i < m_logs.length; i++)
+		{
+			int RecordID = m_logs[i].getRecord_ID();
+			ReportEngine re = null;
+			
+			re = ReportEngine.get (Env.getCtx(), reportEngineType, RecordID);
+			
+			pdfList.add(re.getPDF());
```
hengsin, line 267: This code block only work if all m_logs record is of same ad_table_id and with record_id > 0. I don't think it is a good idea to assume that's always the case.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -51,9 +64,17 @@ public class ProcessInfoDialog extends Window implements EventListener<Event> {
 
 	private static final String MESSAGE_PANEL_STYLE = "text-align:left; word-break: break-all; overflow: auto; max-height: 250pt; min-width: 230pt; max-width: 450pt;";
 
+	/**	Logger			*/
+	private static final CLogger log = CLogger.getCLogger(WGenForm.class);
```
hengsin, line 68: WGenForm.class is wrong here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +245,64 @@ public void onEvent(Event event) throws Exception {
 		if (event.getTarget() == btnOk) {
 			this.detach();
 		}
+		if(event.getTarget() == btnPrint) {
+			Clients.showBusy("Processing...");
```
hengsin, line 249: not translated.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +245,64 @@ public void onEvent(Event event) throws Exception {
...
+			int RecordID = m_logs[i].getRecord_ID();
+			ReportEngine re = null;
+			
+			re = ReportEngine.get (Env.getCtx(), reportEngineType, RecordID);
+			
+			pdfList.add(re.getPDF());				
+		}
+		if (pdfList.size() > 1) {
+			try {
+				File outFile = File.createTempFile(printTmpFileName, ".pdf");
```
hengsin, line 271: printTmpFileName is not being set any where ...

org.adempiere.base/src/org/compiere/print/ReportEngine.java
```diff
@@ -2402,6 +2402,28 @@ public void setIsReplaceTabContent(boolean m_isReplaceTabContent) {
 		this.m_isReplaceTabContent = m_isReplaceTabContent;
 	}
 
+	/**
+	 * Get DOC_TABLE_ID array
+	 * @return
+	 */
+	public static int[] getDocTableIDs() {
```
hengsin, line 2409: this public method is not needed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
hengsin: this loop of ReportEngine.getDocTableIDs() is not good, it should be simplify to just `if (ReportEngine.getReportEngineType(tableID) >= 0) return true;`

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +242,63 @@ public void onEvent(Event event) throws Exception {
...
+	public void onPrint() {
+		// Loop through all items
+		List<File> pdfList = new ArrayList<File>();
+		for (int i = 0; i < m_logs.length; i++)
+		{
+			int recordID = m_logs[i].getRecord_ID();
+			ReportEngine re = null;
+			
+			if(recordID > 0) {
+				re = ReportEngine.get (Env.getCtx(), ReportEngine.getReportEngineType(m_logs[i].getAD_Table_ID()), recordID);
```
hengsin, line 263: the safe thing to do here is `int reportEngineType = ReportEngine.getReportEngineType(m_logs[i].getAD_Table_ID()); if (reportEngineType >= 0) { ...}`

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +242,64 @@ public void onEvent(Event event) throws Exception {
...
+		if (pdfList.size() > 1) {
+			try {
+				File outFile = File.createTempFile(getTitle(), ".pdf");					
+				AEnv.mergePdf(pdfList, outFile);
+
+				Clients.clearBusy();
+				Window win = new SimplePDFViewer(getTitle(), new FileInputStream(outFile));
+				SessionManager.getAppDesktop().showWindow(win, "center");
+			} catch (Exception e) {
+				log.log(Level.SEVERE, e.getLocalizedMessage(), e);
```
hengsin, line 277: Better just rethrow as runtime exception instead. Otherwise it is silent error to user.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +242,64 @@ public void onEvent(Event event) throws Exception {
...
+			}
+		} else if (pdfList.size() > 0) {
+			Clients.clearBusy();
+			try {
+				Window win = new SimplePDFViewer(getTitle(), new FileInputStream(pdfList.get(0)));
+				SessionManager.getAppDesktop().showWindow(win, "center");
+			} catch (Exception e)
+			{
+				log.log(Level.SEVERE, e.getLocalizedMessage(), e);
```
hengsin, line 286: Better just rethrow as runtime exception instead. Otherwise it is silent error to user.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +242,64 @@ public void onEvent(Event event) throws Exception {
 		if (event.getTarget() == btnOk) {
 			this.detach();
 		}
+		if(event.getTarget() == btnPrint) {
+			Clients.showBusy(Msg.getMsg(Env.getCtx(), "Processing"));
+			Clients.response(new AuEcho(ProcessInfoDialog.this, "onPrint", null));
```
hengsin, line 247: the alternative is to use `Executions.schedule(this.getDesktop(), e -> onPrint(), new Event("onPrint"));`. The benefit is you can declare onPrint as private and didn't add a "fake" public method to ProcessInfoDialog (fake as in although it is public but it is not means to be call directly by other class).

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +237,64 @@ public void onEvent(Event event) throws Exception {
...
+		if (pdfList.size() > 1) {
+			try {
+				File outFile = FileUtil.createTempFile(getTitle(), ".pdf");					
+				AEnv.mergePdf(pdfList, outFile);
+
+				Clients.clearBusy();
+				Window win = new SimplePDFViewer(getTitle(), new FileInputStream(outFile));
+				SessionManager.getAppDesktop().showWindow(win, "center");
+			} catch (Exception e) {
+				throw new RuntimeException(e.getLocalizedMessage());
```
hengsin, line 272: best practise is to use throw new RuntimeException(e.getLocalizedMessage(), e); instead so the exception stack trace is not lost. same goes for the throw below.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +237,64 @@ public void onEvent(Event event) throws Exception {
...
+				re = ReportEngine.get (Env.getCtx(), reportEngineType, recordID);
+				pdfList.add(re.getPDF());
+			}
+		}
+		if (pdfList.size() > 1) {
+			try {
+				File outFile = FileUtil.createTempFile(getTitle(), ".pdf");					
+				AEnv.mergePdf(pdfList, outFile);
+
+				Clients.clearBusy();
```
hengsin, line 268: this should goes before "try{", same as the Clients.clearBusy() call below.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +237,64 @@ public void onEvent(Event event) throws Exception {
...
+			Clients.showBusy(Msg.getMsg(Env.getCtx(), "Processing"));
+			Executions.schedule(this.getDesktop(), e -> onPrint(), new Event("onPrint"));
+		}
+	}
+	
+	/**
+	 * On Print
+	 */
+	private void onPrint() {
+		// Loop through all items
```
hengsin, line 250: hmm ... sorry for the mistake in previous review, Clients.clearBusy(); should be the first line in this method, not the 2 place below. No matter what happens below, we need to make sure the busy mask is clear.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/component/ProcessInfoDialog.java
```diff
@@ -216,6 +237,64 @@ public void onEvent(Event event) throws Exception {
...
+				throw new RuntimeException(e.getLocalizedMessage(), e);
+			}
+		}
+		this.detach();
+	}
+	
+	/**
+	 * Is Printable
+	 * @param pi
+	 * @return
```
hengsin, line 290: this is minor but @return is empty here.

hengsin (changes requested):

PR #1413, opened 2022-07-27
IDEMPIERE-5358 Refactor CreateFrom

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5358


org.adempiere.ui/src/org/compiere/grid/CreateFromInvoice.java
```diff
@@ -82,10 +79,11 @@ public boolean dynInit() throws Exception
 	}   //  dynInit
 
 	/**
-	 * Load PBartner dependent Order/Invoice/Shipment Field.
+	 * Load BPartner related Shipment records.
 	 * @param C_BPartner_ID
+	 * @return list of shipment records
 	 */
-	protected ArrayList<KeyNamePair> loadShipmentData (int C_BPartner_ID)
```
CarlosRuiz-globalqss, line 88: haven't tested, just quick checking the changes.

this is minor - as it can be easily solved in plugins, just wondering as changing the name of a protected method can break plugins that override this method

CarlosRuiz-globalqss (approved):

PR #1418, opened 2022-08-02
IDEMPIERE-5362 - Attribute Dialog UX improvments 2022 aug

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5362

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
@@ -216,7 +216,9 @@ public class WPAttributeDialog extends Window implements EventListener<Event>
 	private void init () throws Exception
 	{
 		mainLayout.setParent(this);
-		mainLayout.setHeight("500px");
+		ZKUpdateUtil.setHflex(mainLayout, "1");
+		ZKUpdateUtil.setVflex(mainLayout, "min");
+		mainLayout.setStyle("max-height: 600px;");
 		
 		North north = new North();
 		north.setSclass("dialog-content");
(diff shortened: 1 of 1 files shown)


PeterTakacs300:
Before:
Screenshot 2022-08-02 at 10 59 17

After:
Screenshot 2022-08-02 at 10 54 58

Problem:
Screenshot 2022-08-02 at 10 55 51

I temporarily hardcoded the height to 500px, I would need help with setting max-height to the dialog, so there would not be the empty space seen on the picture above, when we have less attributes.

hengsin:
hi @PeterTakacs300 , you have to use css or setstyle to set min or max height (https://www.w3schools.com/cssref/pr_dim_min-height.asp)

PeterTakacs300:
Current state

Max height:
Screenshot 2022-08-03 at 08 59 49

Shrinking when there are less attributes:
Screenshot 2022-08-03 at 09 00 09

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WPAttributeDialog.java
```diff
@@ -215,13 +218,22 @@ private void init () throws Exception
 		mainLayout.setParent(this);
 		ZKUpdateUtil.setHflex(mainLayout, "1");
 		ZKUpdateUtil.setVflex(mainLayout, "min");
+		mainLayout.setStyle("max-height: 600px;");
```
hengsin, line 221: This wouldn't works well on mobile with smaller height. one option is to do it with condition, i.e `if (ClientInfo.maxHeight(600) mainLayout.setStyle("max-height: 100%;"); else mainLayout.setStyle("max-height: 600px;");`

As a safety net, we can setMaxizable(true) for the dialog. This give user a way to make the dialog usable if there's problem with default sizing set by the application.

PeterTakacs300:
https://user-images.githubusercontent.com/93127072/182789155-f007b4b9-f3d2-4dde-bbaf-0dc38f34b73b.mp4

hengsin (changes requested):
please update ticket with the details of other changes.

norbertbede:
done

CarlosRuiz-globalqss (changes requested):

CarlosRuiz-globalqss (approved):

CarlosRuiz-globalqss:
Tested, I find this OK.  @hengsin  ?

hengsin:
ok for me.

hengsin (approved):
tested

PR #1421, opened 2022-08-03
IDEMPIERE-4723 - Deprecate and replace FDialog

d-ruiz:
Discussion in the ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-4723

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
@@ -61,19 +61,32 @@ public class Dialog {
      * @param message additional message
      * @return The translated AD_Message appended with the additional message
      */
-    private static StringBuffer constructMessage(String adMessage, String message) {
+    private static String constructMessage(String adMessage, String message, Object ... args) {
 		StringBuffer out = new StringBuffer();
 
 		if (!Util.isEmpty(adMessage)) {
-			out.append(Msg.getMsg(Env.getCtx(), adMessage));
+			out.append(getADMessage(adMessage, args));
 		}
 
 		if (!Util.isEmpty(message)) {
-			out.append("<br>").append(message);
+			if (out.length() > 0)
+				out.append("<br>");
+			out.append(message);
 		}
 
-		return out;
+		return out.toString();
 	}
+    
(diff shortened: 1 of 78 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
```diff
@@ -0,0 +1,496 @@
...
+import org.compiere.util.Env;
+import org.compiere.util.Msg;
+import org.compiere.util.Trace;
+import org.compiere.util.Util;
+import org.zkoss.zk.ui.Desktop;
+import org.zkoss.zk.ui.Executions;
+import org.zkoss.zk.ui.event.Event;
+import org.zkoss.zk.ui.event.EventListener;
+
+public class Dialog {
```
hengsin, line 53: I think the Dialog name is fine. However I think it should have been declare as final with a private constructor since it is not means to be extended or instantiated.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FDialog.java
```diff
@@ -51,7 +51,7 @@
  * @date    Feb 25, 2007
  * @version $Revision: 0.10 $
  */
-
+@Deprecated(since="10", forRemoval=true)
```
hengsin, line 54: I think it help to point out that this is replace by the Dialog class.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
d-ruiz: Done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/FDialog.java
d-ruiz: Done

hengsin (changes requested):
[1421_warnings.patch.txt](https://github.com/idempiere/idempiere/files/9264673/1421_warnings.patch.txt)
- remove never use import.

Patch applied

hengsin (approved):
looks good

PR #1422, opened 2022-08-04
[pipo] Throw useful exception when table not found

kriegfrj:
With this commit, Pack In will throw an informative error message if it can't find a table (including the table name).

Without this fix, you simply get an NPE which isn't very helpful for debugging.

Implements https://idempiere.atlassian.net/browse/IDEMPIERE-5366


CarlosRuiz-globalqss (approved):
looks good

PR #1423, opened 2022-08-04
IDEMPIERE-5367 Logic expression IN doesn't works with quoted text

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5367


CarlosRuiz-globalqss (commented):
Hi @hengsin - this unit test fails:
[1423_fail.txt](https://github.com/idempiere/idempiere/files/9259035/1423_fail.txt)

expr = @LineType@="C"&@CalculationType@='A,R,S'

LineType = C
CalculationType = A
must return false - is returning true

LineType = C
CalculationType = A,R,S
must return true - is returning false

So, we would not be able to compare strings that have a comma inside?

hengsin:
no, that us expected to work. will fix.

hengsin:
hi @CarlosRuiz-globalqss , fix committed, please re-test.

CarlosRuiz-globalqss (approved):

PR #1424, opened 2022-08-04
IDEMPIERE-5368 Pack In doesn't close resources properly on failure

kriegfrj:
The input stream was not explicitly closed on exit (whether a success or failure). This leaves a resource leak and for added misery for Windows users meant that you can't pack in more than once without restarting the server.

Fixes [IDEMPIERE-5368](https://idempiere.atlassian.net/jira/software/c/projects/IDEMPIERE/issues/IDEMPIERE-5368)

_Edit: changed link to the correct issue in JIRA_

### DIFF
--- a/org.adempiere.pipo/src/org/adempiere/pipo2/PackIn.java
+++ b/org.adempiere.pipo/src/org/adempiere/pipo2/PackIn.java
@@ -131,8 +131,7 @@ public class PackIn {
 			if (log.isLoggable(Level.INFO)) log.info("importXML:" + msg);
 			return msg;
 		}
-		try {
-			FileInputStream input = new FileInputStream(in);
+		try (FileInputStream input = new FileInputStream(in)) {
 			return importXML(input, ctx, trxName);
 		} catch (Exception e) {
 			log.log(Level.SEVERE, "importXML:", e);


hengsin (approved):
trivial fix, good to go

hengsin (changes requested):
While the fix looks trivial, can you elaborate further how this suppose to fix IDEMPIERE-5270 ?

CarlosRuiz-globalqss:
> While the fix looks trivial, can you elaborate further how this suppose to fix IDEMPIERE-5270 ?

it's a mistake in the commit message, it fixes IDEMPIERE-5368

BTW @kriegfrj - we use to set the commit subject equals to the ticket name and description

hengsin:
If it is a typo, this need to be closed since it is now being link to the wrong ticket in Jira.

kriegfrj:
> > While the fix looks trivial, can you elaborate further how this suppose to fix IDEMPIERE-5270 ?
> 
> it's a mistake in the commit message, it fixes IDEMPIERE-5368
> 

So sorry - that was very careless of me. Fixed now. Forgive me.

> BTW @kriegfrj - we use to set the commit subject equals to the ticket name and description

Thanks for the "heads up", and for your patience! May I humbly request: I usually try hard to follow the contribution guidelines in any OSS project I work on, and I just double-checked and this requirement/suggestion was not listed in https://github.com/idempiere/idempiere/blob/master/CONTRIBUTING.md. 

(Btw: In most of the other OSS projects I've worked on, the ticket/issue name describes the problem whereas the PR title describes the solution (which is often not known at the time that the problem is reported). But I'm happy to "go with the flow".)

I did have one suggestion though - characters in the first line of a commit message are precious as they are often used in a one-line summary with a fixed number of characters (I believe GitHub normally recommends a max of 70 chars). The "IDEMPIERE-" at the start of each commit message is pretty redundant and doesn't really add any useful information (we know that the commits are related to iDempiere because they're in the iDempiere source tree). We could reclaim some of those precious characters for more useful/specific information if we left off the "IDEMPIERE-" bit and just put the ticket number. I realise this might not work though if there is automatic ticket linkage back to JIRA, but if that could be overcome then I think that this might be worth considering.

hengsin (approved):
trivial fix, good to go

CarlosRuiz-globalqss:
Thanks @kriegfrj 

Maybe we thought it was implied here:
https://wiki.idempiere.org/en/Fork_and_Branch_Git_Workflow

Thanks for reviewing, I'll take care later about CONTRIBUTING.md

About the IDEMPIERE- at the beginning, yes - we use that to link JIRA with github.

kriegfrj:
> Thanks @kriegfrj
> 
> Maybe we thought it was implied here: https://wiki.idempiere.org/en/Fork_and_Branch_Git_Workflow
>

Yes, that is perfect! All that's missing then is a link to _that_ Wiki page from CONTRIBUTING.md (unless I missed it, I couldn't see the link there, though I saw a link to other Wiki pages).

> Thanks for reviewing, I'll take care later about CONTRIBUTING.md
> 
> About the IDEMPIERE- at the beginning, yes - we use that to link JIRA with github.

👍

Thank you too for cherry-picking into release-9, that's helpful.

PR #1427, opened 2022-08-04
IDEMPIERE-5355 Improve default trx name (FHCA-3765)

CarlosRuiz-globalqss:
Improve performance using StackWalker instead of Throwable->getStackTrace

https://idempiere.atlassian.net/browse/IDEMPIERE-5355


hengsin (approved):
looks good

PR #1431, opened 2022-08-09
IDEMPIERE-5373 - Attach result file from Import CSV Process when runn…

d-ruiz:
…ing on Background

https://idempiere.atlassian.net/browse/IDEMPIERE-5373


hengsin (approved):
trivial fix, good to go

PR #1434, opened 2022-08-10
IDEMPIERE-5378 MFA must allow to select the mechanism when the user has several registered (FHCA-3808)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5378


hengsin (approved):
looks good

PR #1437, opened 2022-08-11
IDEMPIERE-5380 StackOverflowError when try to login in application with substitute user (FHCA-3828)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5380

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRole.java
+++ b/org.adempiere.base/src/org/compiere/model/MRole.java
@@ -2897,9 +2897,9 @@ public final class MRole extends X_AD_Role implements ImmutablePOSupport
 	 */
 	private void loadSubstitutedRoles(boolean reload)
 	{
-		if (this.m_parent != null)
+		if (this.m_parent != null || isMasterRole())
 		{
-			// load only if this is logged role (no parent roles) 
+			// load only if this is logged role (no parent or master roles) 
 			return;
 		}
 		//


org.adempiere.base/src/org/compiere/model/MRole.java
```diff
@@ -2897,9 +2897,9 @@ private void loadChildRoles(boolean reload)
 	 */
 	private void loadSubstitutedRoles(boolean reload)
 	{
-		if (this.m_parent != null)
+		if (this.m_parent != null || isMasterRole())
```
hengsin, line 2900: isn't the error is cause by user and substitute user is using the same master role instead of the fact the user is using a master role ?

CarlosRuiz-globalqss: My finding was that loadSubstitutedRoles is being called multiple times, and it becomes recursive.

loadSubstitutedRoles just need to be called one time, because is user dependent, not role dependent, so the loadSubstitutedRoles must work just on the login role, not on children (which was already excluded) or master roles (which was not)

BTW, I think probably we have discussed this before, but the design of substitute roles seems wrong, however, is out of the scope of the ticket, would be worthy to discuss it in another ticket?

hengsin: hmm .. I think the fix should be along the line where if the substitute role = current role or have been loaded as child/included role, we should not try to include it again.

CarlosRuiz-globalqss: I tried that as well, changing the whereClause in loadSubstitutedRoles, with the following patch:
[badTry.txt](https://github.com/idempiere/idempiere/files/9324393/badTry.txt)
But, it didn't work.

Also, there is where I found that the substitute roles is somewhat wrongly designed, because the substitute is related to the user, and not to the login role.
From what I see it is just wrong to add options from a role that is not selected on Login (which is the actual behavior of substitute roles).

I think what substitute role must do is to allow the user to login with the roles from the user being substituted.  With the actual behaviour, the substitute user is presented with a different menu than the substituted user when using exactly the same login role (I find that concerning).
The other point to use substituted roles is when doing a workflow approval.

But again, that's beyond the current ticket fix, the fix here is just to solve a bad recursion scenario.

hengsin (approved):
looks ok

PR #1438, opened 2022-08-11
IDEMPIERE-5381 System users cannot revoke MFA trusted devices in GardenWorld (FHCA-3824)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5381


hengsin (approved):
looks good

PR #1439, opened 2022-08-11
IDEMPIERE-5382 Allow just one MFA registration per mechanism (FHCA-3822)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5382


hengsin (approved):
looks good

PR #1448, opened 2022-08-22
IDEMPIERE-5389 - Dashboard row layout

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5389

Improvements:

- row (horizontal) layout for the dashboard
- responsiveness to screen size
- SysConfig: DASHBOARD_LAYOUT_ORIENTATION
- ability to load data from AD_StatusLine and render it on dashboard

### DIFF
--- a/migration/iD10/oracle/202208181113_IDEMPIERE-5389.sql
+++ b/migration/iD10/oracle/202208181113_IDEMPIERE-5389.sql
@@ -1,0 +1,50 @@
+-- IDEMPIERE-5389
+SELECT register_migration_script('202208181113_IDEMPIERE-5389.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Aug 16, 2022, 4:52:58 PM CEST
+INSERT INTO AD_SysConfig (AD_SysConfig_ID,AD_Client_ID,AD_Org_ID,Created,Updated,CreatedBy,UpdatedBy,IsActive,Name,Value,Description,EntityType,ConfigurationLevel,AD_SysConfig_UU) VALUES (200203,0,0,TO_TIMESTAMP('2022-08-16 16:52:58','YYYY-MM-DD HH24:MI:SS'),TO_TIMESTAMP('2022-08-16 16:52:58','YYYY-MM-DD HH24:MI:SS'),100,100,'Y','DASHBOARD_LAYOUT_ORIENTATION','R','C - columns (vertical layout) / R - rows (horizontal layout)','D','S','4de26350-c352-4552-87be-44abe331f821')
+;
+
+-- Aug 18, 2022, 11:13:06 AM CEST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,FKConstraintType,IsHtml) VALUES (215099,0,'Status Line',50010,'AD_StatusLine_ID',22,'N','N','N','N','N',0,'N',19,0,0,'Y',TO_TIMESTAMP('2022-08-18 11:13:06','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2022-08-18 11:13:06','YYYY-MM-DD HH24:MI:SS'),100,202635,'Y','N','D','N','N','N','Y','6318c76d-e4e1-4846-a1e8-57d4c5a52df3','N',0,'N','N','N','N')
+;
+
+-- Aug 18, 2022, 11:13:15 AM CEST
+UPDATE AD_Column SET FKConstraintName='ADStatusLine_PADashboardConten', FKConstraintType='N',Updated=TO_TIMESTAMP('2022-08-18 11:13:15','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=215099
+;
+
+-- Aug 18, 2022, 11:13:15 AM CEST
+ALTER TABLE PA_DashboardContent ADD AD_StatusLine_ID NUMBER(10) DEFAULT NULL 
+;
+
+-- Aug 18, 2022, 11:13:15 AM CEST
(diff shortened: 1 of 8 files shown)


org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
```diff
@@ -267,3 +267,113 @@
...
+	}
+	.dashboard-row-4 .z-hlayout-inner {
+		width: 100%;
+		display: flex;
+	}
+	.dashboard-row-3 .z-hlayout-inner {
+		width: 100%;
+		display: flex;
+	}
+}
```
hengsin, line 379: This basically a hard coded layout. Wouldn't works well for the generic need of the project and it is not going to works fine if the dashboard still support dynamic reordering of dashboard widget by user.

PeterTakacs300: Do you mean by hardcoded that we've defined 10 classes (support 10 dashboards in 1 row)? If we would change it to 12, it would be the exact limitation Bootstrap has. We assumed the admin never defines more than 10 gadgets in 1 row.

According to my tests the dynamic reordering of dashboard widget by user worked for me. Probably we need to specify the use case, when does it not work fine?

We focus to keep same principles on server side, and on client side we achieve the better responsiveness by client-side css.

At the same time we think, that the css was not written the best way. If we change the exact number of defined classes to dynamic approach, would this solve the problem? 

For clarifying - does the java improvement seems to be good? Or you haven't checked it yet? 

Could you explain more for resolving the issues, please?

norbertbede: Hi @hengsin , we are consulted here the css question with our web designer. We modify the design which also covering span.

Firstly we define a new sysconfig - define "dashboard (row/col) max gadgets"  . let say we set typical boostrap 12 figure. Dashboard Preference will have "GadgetSpan" field, where the admin enter the number which specify how many "positions" cols this gadget taken.  let say set 6 then that means in case 12 max gadgets 50% and additional 2 will have 3+3  (I assume this  can be implemented for column based layout too probably also with support without specify span field)

We will rename css classes, because we had wrong approach - however 12 classes for 12 cols must be defined. + 12 for media small device. (this because we are able client side redraw the layout). we are can't find better solution. 

Another options would be flexbox https://css-tricks.com/snippets/css/a-guide-to-flexbox/  or replace fully grid by javascript lib which is for flexible layout generation. (the second is out of our knowledge for now.)

if you have another idea let us know.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -579,6 +797,10 @@ public void onEvent(AfterSizeEvent event) throws Exception {
 	public void onEvent(Event event) throws Exception {
 		Component comp = event.getTarget();
         String eventName = event.getName();
+        String layoutOrientation = MSysConfig.getValue(MSysConfig.DASHBOARD_LAYOUT_ORIENTATION, Env.getAD_Client_ID(Env.getCtx()));
+        
+        if(!layoutOrientation.equals(DASHBOARD_LAYOUT_ROWS) && !layoutOrientation.equals(DASHBOARD_LAYOUT_COLUMNS))
+        	layoutOrientation = DASHBOARD_LAYOUT_ROWS;
```
hengsin, line 803: for backward compatibility, this should fall back to columns instead.

org.adempiere.base/src/org/compiere/model/MDashboardPreference.java
```diff
@@ -41,41 +41,61 @@ public static int getForSessionColumnCount(boolean isShowInDashboard, int AD_Use
...
 		List<MDashboardPreference> list = getForSessionQuery(isShowInDashboard, AD_User_ID, AD_Role_ID).list();
 		return list.toArray(new MDashboardPreference[list.size()]);
 	}
 	
 	public static Query getForSessionQuery(boolean isShowInDashboard, int AD_User_ID, int AD_Role_ID)
+	{
+		return getForSessionQuery(isShowInDashboard, AD_User_ID, AD_Role_ID, -1);
+	}
+	
+	public static Query getForSessionQuery(boolean isShowInDashboard, int AD_User_ID, int AD_Role_ID, int lineNo)
```
hengsin, line 63: this added method is not use anywhere, what's the purpose of adding this ?

hengsin:
[pr1448.patch.txt](https://github.com/idempiere/idempiere/files/9439412/pr1448.patch.txt)
- change DASHBOARD_LAYOUT_ORIENTATION configuration level from System to Client
- minor refinement
- reduce duplication of code

hengsin:
hi @CarlosRuiz-globalqss , this tested ok for me, do you have time to review this ?

migration/iD10/oracle/202208181113_IDEMPIERE-5389.sql
```diff
@@ -0,0 +1,46 @@
...
+;
+
+-- Aug 23, 2022, 4:20:40 PM CEST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Description,AD_Table_ID,ColumnName,DefaultValue,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,FKConstraintType,IsHtml) VALUES (215100,0,'Flex Grow','This defines the ability for a flex item to grow if necessary. It accepts a unitless value that serves as a proportion. It dictates what amount of the available space inside the flex container the item should take up.',200013,'FlexGrow','1',10,'N','N','N','N','N',0,'N',11,0,0,'Y',TO_TIMESTAMP('2022-08-24 14:01:08','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2022-08-24 14:01:08','YYYY-MM-DD HH24:MI:SS'),100,203636,'Y','N','D','N','N','N','Y','ed69cf80-6c51-4d48-8ae1-363e3e98591d','Y',0,'N','N','N','N')
```
CarlosRuiz-globalqss, line 16: This lines has a wrong match vs the postgresql version.

This patch fixes the issue:
[FixOracleMigration.txt](https://github.com/idempiere/idempiere/files/9445340/FixOracleMigration.txt)

CarlosRuiz-globalqss (changes requested):
please apply patch to fix oracle script

CarlosRuiz-globalqss:
Also, please fix the documentation page that is in draft:
https://wiki.idempiere.org/en/NF10_Dashboard_Row_Orientation

CarlosRuiz-globalqss (approved):
Thanks @PeterTakacs300 and @norbertbede 

From my side this is ready too.

Just completing the documentation is pending.

CarlosRuiz-globalqss:
Ah, sorry, just noticed the documentation is ready

PR #1452, opened 2022-08-24
IDEMPIERE-5361 : Jasper reports with Charts are not working

nmicoud:
Hi,
I've followed a suggestion from @anozimada (downgrade jfreechart to 1.0.19) and in my tests, it seems to work BUT I probably didn't make the change in the correct way (as there is an error in Eclipse (Plugin execution not covered by lifecycle configuration: org.apache.maven.plugins:maven-antrun-plugin:1.8:run (execution: default, phase: validate))

That would be great if someone with better technical knowledge can have a look and/or work on it as this is far from my abilities (I'm playing the sorcerer's apprentice) ; I would be happy to test.

Thanks,

Nicolas


hengsin:
[pr1452.patch.txt](https://github.com/idempiere/idempiere/files/9424833/pr1452.patch.txt)
- remove warning
- update target platform version

nmicoud:
Hi,
Patch integrated

hengsin (approved):
tested

PR #1455, opened 2022-08-27
IDEMPIERE-5397 Refactoring of Product Attribute Grid

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5397


CarlosRuiz-globalqss (commented):
Hi @hengsin - I found some issues with the change on the form

With this patch:
[IDEMPIERE-5397_PR.txt](https://github.com/idempiere/idempiere/files/9452246/IDEMPIERE-5397_PR.txt)

I solved the issues, and restored a behavior more close to what was before in the form, the most important changes:
* show the table complete by default, not collapsed
* show also the cells where there are no products

This last change is important when analyzing products (for example in garment industry) to see if the matrix is complete.

I changed the code directly removing some code, didn't do a refactor, but maybe would be better to review the method filterAttributeValues which is now very basic or useless, and is not really filtering anything.

I did tests with the old behavior and everything is OK.
Also did tests with filter and grouping - and it seems OK too.

Also, tested your 2Pack to add a button to the sales order (nice!) - and it still worked fine.

But, the unit test is broken, because of the changes, and I didn't understand 100% the test case.

Can you please review and integrate the patch if you agree with the changes?

hengsin:
Hi @CarlosRuiz-globalqss , the patch completely change the intended behaviour of the change which is to show only attributes combination with corresponding product. If you think that's not good, please close the PR and I will abandon the PR and the corresponding ticket, thanks.

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss , the patch completely change the intended behaviour of the change which is to show only attributes combination with corresponding product. If you think that's not good, please close the PR and I will abandon the PR and the corresponding ticket, thanks.

Thanks @hengsin - the way how the form behaves is a required use case for some situations, but I think there is big value in your proposal for other situations.

Especially as you added an even better use in sales order lines.

I think the best would be to have both options, maybe the easiest would be with a flag to toggle showing the whole grid of just products.

If you like I can try to continue from where you left and I'll try to add the flag (however I cannot promise a date, these days have been frantic with too many parallel stuff running on my side).

hengsin:
Hi @CarlosRuiz-globalqss , alright, I will leave this as it is and you can continue when you have time. Also, I don't understand at all why you have remove all the filter attribute code. Anyway, it is a form that probably no one is using and I don't want to spent more time on this.

PR #1456, opened 2022-08-28
IDEMPIERE-5399 Data type Time is mostly useless

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5399


hengsin (approved):
looks good

PR #1460, opened 2022-09-02
IDEMPIERE-5403 : Misleading error message on TimeUtil.getDay(int, int…

nmicoud:
…, int)

https://idempiere.atlassian.net/browse/IDEMPIERE-5403


CarlosRuiz-globalqss (approved):
minor

PR #1461, opened 2022-09-05
IDEMPIERE-5405 - Drill Assistant - potential Cross Tenant Issue on Table tab

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5405


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/WDrillReport.java
```diff
@@ -387,7 +387,7 @@ private Table getDrillTableBox(KeyNamePair drillTable, int tabIndex, int groupIn
 	
 	private KeyNamePair findTablePrintFormat(KeyNamePair drillTable) {
 		
-		Integer printFormatID = new Query(Env.getCtx(), MPrintFormat.Table_Name, " AD_Table_ID = ? ", null).setParameters(drillTable.getKey()).firstId();
+		Integer printFormatID = new Query(Env.getCtx(), MPrintFormat.Table_Name, " AD_Table_ID = ? AND AD_Client_ID = ? ", null).setParameters(drillTable.getKey(), Env.getAD_Client_ID(Env.getCtx())).firstId();
```
CarlosRuiz-globalqss, line 390: would be simpler to use the .setClient_ID() of Query

But, I'm wondering if print formats from System tenant must be included too?
If that's the case then the query must be " AND AD_Client_ID IN (0,?) "

CarlosRuiz-globalqss (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/WDrillReport.java
CarlosRuiz-globalqss: Something additional here, usually when getting a first record you must add an ORDER BY clause.

In this case we also probably would like to put first the default print format, so maybe an ORDER BY like:

IsDefault DESC, Created DESC

norbertbede:
hi @CarlosRuiz-globalqss can we close this please ?

CarlosRuiz-globalqss (approved):

PR #1462, opened 2022-09-05
IDEMPIERE-5406 - Define max-height to Charts in Dashboard Row Layout

PeterTakacs300:
JRIA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5406

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
+++ b/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
@@ -216,8 +216,16 @@
 	width: 100%;
 }
 
-.dashboard-row .chart-gadget {
-	max-height: 300px; 
+@media screen and (min-width: 768px) {
+	.dashboard-row .chart-gadget {
+		max-height: 300px; 
+	}
+}
+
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
```diff
@@ -216,6 +216,10 @@
 	width: 100%;
 }
 
+.dashboard-row .chart-gadget {
+	max-height: 300px;
```
hengsin, line 220: for hardcoded height, it is better to define it by screen size, i.e "@media" notation with max or min height.

PeterTakacs300: Media query added, now everything larger than 768px (tablet to desktop) has suggested max-height 300px, and smaller devices will have 150px. It is hard to test for smaller device because I have charts with JFreeChart, for large devices we were able to test with billboard plug-in, it should work fine there.

hengsin (changes requested):
looks fine but again, for theme changes, you need to update the version timestamp in lang-addon.xml

hengsin (approved):
looks good

PR #1463, opened 2022-09-05
IDEMPIERE-5402 Replace Jfree Chart with Billboard

hengsin:
- replace jfreechart with https://github.com/naver/billboard.js

https://idempiere.atlassian.net/browse/IDEMPIERE-5402


CarlosRuiz-globalqss (approved):
Basic tests
Dashboard shows the same info with a new format
It requires importing two new projects in eclipse (or rematerializing)

PR #1468, opened 2022-09-07
IDEMPIERE-5389 - improve Status Line as Dashboard Content height

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5389

**Before:**
IDEMPIERE-5389 - Padding Multiplied

**After:**
IDEMPIERE-5389 - Padding Fixed

You can see, that the Performance Widget is partially hidden, also as chart doesn't fit widget height (next image)

IDEMPIERE-5389 - Chart Issue

For better testing we provide Pack Out for Status Line: https://idempiere.atlassian.net/browse/IDEMPIERE-5410

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
+++ b/org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
@@ -224,6 +224,8 @@
 	display: flex;
  	justify-content: center;
  	align-items: center;
+ 	margin: 20px 0px 20px 0px;
+ 	height: -webkit-fill-available;
 }
 
 .help-content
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/gadget.css.dsp
```diff
@@ -224,6 +224,8 @@
 	display: flex;
  	justify-content: center;
  	align-items: center;
+ 	margin: 20px 0px 20px 0px;
+ 	height: -webkit-fill-available;
```
hengsin, line 228: this wouldn't work with firefox. what are you trying to achieve here ? have you try flex-grow instead ?

PeterTakacs300: Fixed

PeterTakacs300:
We must refactor how we set same height for widget, and moved to css instead ZK (caused issue with charts and Performance gadget).

The status line css is not perfect, but looks good - we will continue to tune it.

Column layout:
Screenshot 2022-09-07 at 13 57 37

Row layout:
Screenshot 2022-09-07 at 13 58 48

hengsin (approved):
looks ok

PR #1470, opened 2022-09-08
IDEMPIERE-5400 - Gauge Indicator Style and Color Improvement

PeterTakacs300:
JIRA Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5400

Added option to show/hide title similar as Chart - removes bottom title

Style Changes:
Before:
Screenshot 2022-09-08 at 12 25 11

After:
Screenshot 2022-09-08 at 12 24 06


hengsin:
[pr1470.patch.txt](https://github.com/idempiere/idempiere/files/9527865/pr1470.patch.txt)
- update version to avoid issue with browser cache.

Need the same for theme too but I think it is better to address that at the pull request for IDEMPIERE-5412

hengsin (changes requested):
please apply the attach patch

hengsin (approved):
tested

PR #1472, opened 2022-09-09
IDEMPIERE-5410 - Demo Data for Dashboard

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5410

The PR contains also the css classes for KPI gadgets creates through Status Lines


CarlosRuiz-globalqss (changes requested):
Hi @PeterTakacs300 

It would be better to separate the css commit in a different ticket or at least pull request.
Also, when changing/adding css you need to update org.adempiere.ui.zk/WEB-INF/src/metainfo/zk/lang-addon.xml

Regards,

Carlos Ruiz

migration/iD10/oracle/202209081507_IDEMPIERE-5410.sql
```diff
@@ -0,0 +1,226 @@
...
+-- Sep 8, 2022, 3:27:35 PM CEST
+INSERT INTO PA_DashboardPreference (AD_Client_ID,AD_Org_ID,AD_Role_ID,AD_User_ID,ColumnNo,Created,CreatedBy,IsActive,IsCollapsedByDefault,IsShowInDashboard,Line,PA_DashboardPreference_ID,Updated,UpdatedBy,PA_DashboardContent_ID,PA_DashboardPreference_UU,FlexGrow) VALUES (11,0,102,100,1,TO_TIMESTAMP('2022-09-08 15:27:35','YYYY-MM-DD HH24:MI:SS'),100,'Y','N','Y',0,200003,TO_TIMESTAMP('2022-09-08 15:27:35','YYYY-MM-DD HH24:MI:SS'),100,200008,'3763992f-794f-44b9-ba5f-fc4cdad0384f',1)
+;
+
+-- Sep 8, 2022, 3:28:40 PM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=0, IsShowInDashboard='Y', Line=2,Updated=TO_TIMESTAMP('2022-09-08 15:28:40','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=200003
+;
+
+-- Sep 8, 2022, 3:28:40 PM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=1, IsShowInDashboard='Y', Line=2,Updated=TO_TIMESTAMP('2022-09-08 15:28:40','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=1000004
```
CarlosRuiz-globalqss, line 144: there are several lines here referencing non-official IDs

migration/iD10/oracle/202209090852_IDEMPIERE-5410.sql
```diff
@@ -0,0 +1,410 @@
...
+-- Sep 9, 2022, 8:52:46 AM CEST
+INSERT INTO AD_Chart (AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Chart_ID,ChartType,Name,WinHeight,IsDisplayLegend,ChartOrientation,IsTimeSeries,TimeUnit,TimeScope,DomainLabel,RangeLabel,AD_Chart_UU,EntityType) VALUES (0,0,'Y',TO_TIMESTAMP('2022-09-09 08:52:45','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2022-09-09 08:52:45','YYYY-MM-DD HH24:MI:SS'),100,200003,'AC','Orders last 30 days',300,'Y','V','Y','W',0,'Day','Value','f046ce9b-7cef-4085-b479-e47c690f2585','D')
+;
+
+-- Sep 9, 2022, 8:52:46 AM CEST
+INSERT INTO AD_ChartDatasource (AD_Client_ID,AD_Org_ID,AD_Table_ID,Created,CreatedBy,DateColumn,EntityType,IsActive,KeyColumn,Name,AD_ChartDatasource_ID,FromClause,Updated,UpdatedBy,AD_Chart_ID,ValueColumn,TimeOffset,AD_ChartDatasource_UU) VALUES (0,0,259,TO_TIMESTAMP('2022-09-09 08:52:46','YYYY-MM-DD HH24:MI:SS'),100,'o.dateordered','D','Y','o.C_Order_ID','current',200003,'C_Order o',TO_TIMESTAMP('2022-09-09 08:52:46','YYYY-MM-DD HH24:MI:SS'),100,200003,'sum(o.grandtotal)',0,'7f05953d-8591-42d2-b6a3-a27a3381f797')
+;
+
+-- Sep 9, 2022, 8:52:47 AM CEST
+DELETE FROM AD_Chart WHERE AD_Chart_ID=200003
```
CarlosRuiz-globalqss, line 41: this line is a mismatch vs the postgresql version

Note that normally I use these scripts to check problems with the migration scripts:
https://github.com/globalqss/idempiere-stuff/tree/stuff/verifyPullRequest

migration/iD10/oracle/202209090852_IDEMPIERE-5410.sql
```diff
@@ -0,0 +1,410 @@
...
+;
+
+-- Sep 9, 2022, 9:28:44 AM CEST
+INSERT INTO PA_DashboardPreference (AD_Client_ID,AD_Org_ID,AD_Role_ID,AD_User_ID,ColumnNo,Created,CreatedBy,IsActive,IsCollapsedByDefault,IsShowInDashboard,Line,PA_DashboardPreference_ID,Updated,UpdatedBy,PA_DashboardContent_ID,PA_DashboardPreference_UU,FlexGrow) VALUES (11,0,102,100,1,TO_TIMESTAMP('2022-09-09 09:28:44','YYYY-MM-DD HH24:MI:SS'),100,'Y','N','Y',0,200009,TO_TIMESTAMP('2022-09-09 09:28:44','YYYY-MM-DD HH24:MI:SS'),100,200015,'2ce19b47-d9f5-4afe-bd9a-6fdeb05dd3ed',1)
+;
+
+-- Sep 9, 2022, 9:29:10 AM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=0, IsShowInDashboard='Y', Line=0,Updated=TO_TIMESTAMP('2022-09-09 09:29:10','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=1000006
```
CarlosRuiz-globalqss, line 100: several lines here referencing non-official IDs

migration/iD10/oracle/202209091013_IDEMPIERE-5410.sql
```diff
@@ -0,0 +1,152 @@
...
+-- Sep 9, 2022, 11:05:00 AM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=1, IsShowInDashboard='Y', Line=5,Updated=TO_TIMESTAMP('2022-09-09 11:05:00','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=200010
+;
+
+-- Sep 9, 2022, 11:05:12 AM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=0, IsShowInDashboard='Y', Line=7,Updated=TO_TIMESTAMP('2022-09-09 11:05:12','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=200008
+;
+
+-- Sep 9, 2022, 11:05:12 AM CEST
+UPDATE PA_DashboardPreference SET ColumnNo=1, IsShowInDashboard='Y', Line=7,Updated=TO_TIMESTAMP('2022-09-09 11:05:12','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE PA_DashboardPreference_ID=1000003
```
CarlosRuiz-globalqss, line 130: several lines here referencing non-official IDs

CarlosRuiz-globalqss (changes requested):

PeterTakacs300:
Hi @CarlosRuiz-globalqss, I've removed the lines with non-offical IDs and registered IDs for those 8 Dashboard Preferences. 

Also removed the migration script with the Orders from this PR, because it caused errors - when completing orders with 'log migration scripts' on, it generates some invalid lines. 
This way a lot of new charts will be left with no data, so the orders should be added later. We don't know a correct way to log an order completion into migration scripts yet.
Because of that, first we would like you to review without the orders.

migration/iD10/oracle/202209091013_IDEMPIERE-5410.sql
```diff
@@ -0,0 +1,140 @@
...
+;
+
+-- Sep 9, 2022, 10:17:20 AM CEST
+INSERT INTO AD_StatusLine (AD_StatusLine_ID,AD_Client_ID,AD_Org_ID,AD_StatusLine_UU,Created,CreatedBy,EntityType,IsActive,Name,Updated,UpdatedBy,AD_Message_ID,SQLStatement) VALUES (200022,0,0,'53f41a48-b5fc-4016-b8b6-5c450efff3ab',TO_TIMESTAMP('2022-09-09 10:17:20','YYYY-MM-DD HH24:MI:SS'),100,'D','Y','KPI_Gross Profit Margin YTD',TO_TIMESTAMP('2022-09-09 10:17:20','YYYY-MM-DD HH24:MI:SS'),100,200783,'SELECT 
+  ROUND (SUM(((revenue-COGS)/revenue)*100) FILTER (WHERE kpidata.year = date_trunc(''year'', now())),2)::numeric as revenue_YTD ,
```
CarlosRuiz-globalqss, line 35: we need to include here SQL that works both in oracle and postgresql

otherwise the demo won't work in oracle

norbertbede: we forgot about oracle and kpi status line. we have no experience with oracle.
i understand this requirement, no idea for now how to make compatible. (we need understand oracle syntax)
hope i understood the problem - the kpi status lines makes data incompactible.

can be kpi definitions different in PG and oracle ? that is acceptable or we must use common/compatible  syntax ? 
n

hengsin: hi @norbertbede , preferably we should use common syntax to reduce maintenance issue. Some observation and suggestion:
- trunc can be use in place of date_trunc. trunc(timestamp, 'YEAR') and trunc(timestamp, 'MONTH')
- ::numeric, there's no common numeric type between oracle and postgresql. the solution here is either to remove the use of ::numeric and if it is a must, introduce a `@NUMERIC_DB_TYPE@` context variable and replace ::numeric with `cast(... as @NUMERIC_DB_TYPE@)`. Alternatively, I think with Oracle dialect mode, cast(... as Number) should work as the translate layer will convert Number to Numeric.
- there's no equivalent of generate_series in Oracle. While we can create similar function in Oracle, the sql syntax of using it will be slightly different from PostgreSQL.

Another option is to distribute this as postgresql only plugin and free your self from the Core constraint of having to support both Oracle and PostgreSQL.

CarlosRuiz-globalqss: @hengsin is right, additionally:

* most of the time we can simply remove the ::numeric (and any ::casting) and postgresql works fine, it is required in very few cases
* I see also in the queries SUM ... FILTER - that's postgresql only syntax, and can be expressed in a generic way as SUM ... CASE WHEN ...
* generate_series has no equivalent, in the case of the query here, as it is generating 12 records it could be replaced in a generic way with a JOIN with a subquery table with 12 SELECTS joined by UNION

PeterTakacs300: Hi @CarlosRuiz-globalqss, @hengsin 

In the latest commit I removed the ::numeric casts, replaced date_trunc with trunc function and replaced FILTER with CASE WHEN statement, we would like you to review if this would be ok. 

The remaining question is that what would be the best way of replacing/removing generate_series?

norbertbede: would be great use similar approach as on truncate().
as @CarlosRuiz-globalqss suggest
=> there's no equivalent of generate_series in Oracle. While we can create similar function in Oracle, the sql syntax of using it will be slightly different from PostgreSQL.
can you help with this ?

hengsin:
[pr1472.kpi-generate-series.sql.txt](https://github.com/idempiere/idempiere/files/9652130/pr1472.kpi-generate-series.sql.txt)
- example of generate_series equivalent that works for both oracle and postgresql

CarlosRuiz-globalqss:
Comments after running the latests scripts:

1. This is inserting PA_DashboardPreference records which must exists almost in all databases, and because of that the dashboard is being duplicated.
I haven't checked, but I think is better not to insert records in this table, they will be automatically populated when entering into the app.

1. Login as SuperUser into System is showing lots of "tenant" dashboards that are not relevant for System roles, for example "Sales YTD", "Opportunities by Campaign", etc.  AFAIR we don't show that in System

1. the query
```SELECT AVG(TotalLines) FROM C_Order WHERE AD_CLient_ID = @#AD_Client_ID@ AND issotrx='Y' AND trunc(DateOrdered, 'D') BETWEEN trunc(now() - INTERVAL '1 year', 'D') AND trunc(now(), 'D')```
shows error in oracle:
```ORA-30089: missing or invalid <datetime field>```
Solution 1st: replace ```now()``` with ```getdate()```
Solution 2nd: instead of ```getdate() - INTERVAL '1 year'``` use ```add_months(getdate(),-12)```

norbertbede:
hi @CarlosRuiz-globalqss 

1. if we are not inserting, then dashboard layout is absolutely random. that is the reason why we register these preferences (logged then set awaited layout)
2. On System we are doesn't log anything so created automatically when you logged in. 
3. This can be fixed tomorrow, we oversight this.

hengsin:
It seems we lack a setting to say a dashboard is applicable to either system tenant, non-system tenant or both.

CarlosRuiz-globalqss:
or - said in other words - do we want to push these new dashboards to every non GardenWorld tenant out there?

hengsin:
+1 that the new dashboard should be created at gardenworld instead of system. if we want to make it available for new tenant, I think we need to add some sort of template mechanism here. another option is to release this as 2pack package that can be installed to any tenant.

norbertbede:
hi, actually, we can change script to gardenworld, partially resolve the issue for specific user
Dashboard preferences are on user level - in other words, no templating as hengsin wrote. So if you login with user GardenUser you will see same mess as for system.

Our vision long termi is clear. only multi-dashboard resolve it, which support multi-tenancy and dasbhoard content reusability. Partially we have done, as i wrote we are working on it continually. There is another ticket around.

We can fix point 3 and move dashboard contents to the GW for now, or we can simple drop this PR. 
n

CarlosRuiz-globalqss:
> if we are not inserting, then dashboard layout is absolutely random. that is the reason why we register these preferences (logged then set awaited layout)

There is something missing then, the design of the dashboard is that you create the dashboard and it will be accommodated automatically for the users login into the app.  It doesn't make sense that you need to go through every user and set the preference.

AFAIR that was achieved with the ColumnNo and Line in the PA_DashboardContent table, the configuration of those two variables did the trick to distribute correctly the dashboards for new users.
Does that still work?

hengsin:
hi @norbertbede , if your are working on a multi dashboard implementation, it make more sense to move that first since that should make many of the changes here invalid.

norbertbede:
yes. agree looks cumbersome. and we are working on dashboard, however need more time. first we tune it. then start to commit. huge change. you can close it.

PR #1474, opened 2022-09-09
IDEMPIERE-5414 - Add Help icon to Dashboard Content

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5414


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -335,20 +338,40 @@ private Panel newGadgetPanel(MDashboardPreference dp, MDashboardContent dc) {
...
+			help.setImage(ThemeManager.getThemeResource(IMAGES_CONTEXT_HELP_PNG));
+		Popup popup = new Popup();
+		popup.setSclass("dashboard-content-help");
+		popup.setPopup(popup);
+		Text t = new Text(text);
+		popup.appendChild(t);
+		help.setTooltip(popup);
+		help.addEventListener(Events.ON_CLICK, (Event event) -> {
+			popup.setPage(help.getPage());
+			popup.open(help, "after_start");
```
hengsin, line 373: need to add LayoutUtils.autoDetachOnClose(popup) call. without which, can lead to memory leak.

PeterTakacs300: Didn't know that, thank you

hengsin:
hi @PeterTakacs300 , the latest changes works fine but I notice 2 minor issue here:
- alignment between text and the question mark icon seems a bit off. I think the icon is probably 1 or 2 pixel lower than where it should be at.
- should hide the icon when the widget's label and description is the same.

PeterTakacs300:
Hi @hengsin. The problem is, that is seems like the button is centered correctly, and the image within the button is off. Maybe other component would be more correct instead of ToolbarButton? I haven't find another solution yet.

hengsin:
hi @PeterTakacs300, I've try using margin-bottom and it seems to help. Beside toolbarbutton, you can try to use A as well.

hengsin (approved):
looks good

PR #1477, opened 2022-09-10
IDEMPIERE-5416 Alt+A is used for Create+New - for usability would be good to use it also for New

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5416

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
@@ -2721,21 +2721,18 @@ public abstract class AbstractADWindowContent extends AbstractUIPart implements 
 			@Override
 			public void onCallback(Boolean result)
 			{
-				if(result)
-		    	{
-		    		adTabbox.getSelectedGridTab().dataRefreshAll(true, true);
-		    		adTabbox.getSelectedGridTab().refreshParentTabs();
-		    		IADTabpanel dirtyTabpanel = (IADTabpanel) Executions.getCurrent().removeAttribute("adtabpane.saved");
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
```diff
@@ -2721,21 +2721,18 @@ public void onSaveCreate()
...
+				IADTabpanel dirtyTabpanel = (IADTabpanel) Executions.getCurrent().removeAttribute("adtabpane.saved");
+				if (dirtyTabpanel != null && dirtyTabpanel.getGridTab().isDetail()) {
+					try {
+						adTabbox.getSelectedTabpanel().getDetailPane().onNew();
+					} catch (Exception e) {
+						throw new RuntimeException(e);
+					}
+				} else {
+					onNew();
+				}
```
hengsin, line 2735: why remove the `if(result)` checking ?

CarlosRuiz-globalqss: result is the return of the onSave operation.
The dataSave is returning ERROR when the record has not been modified, and there is nothing to save, and consequently the onSave is returning false, so the onNew is not called in this case.  I tested it and it seems to work without problems.

hengsin: Not working well for the following:
- Open Product Window
- Create new record, Enter Weeder for both Search Key and Name
- Press Alt+A to save

CarlosRuiz-globalqss: Thanks @hengsin - I added a new commit that works better.

hengsin (approved):
tested

PR #1480, opened 2022-09-13
IDEMPIERE-5412 - css classes for Status Line KPI gadgets

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5412


hengsin:
hi @PeterTakacs300 , this is making thing quite messy for us to review. It makes more sense to me that we should have a new ticket for all the kpi gadget and corresponding css classes. Your "demo data" ticket should really include just the "demo data" scripts.

hengsin:
This can't be reviewed separately from the corresponding KPI gadgets.

PR #1483, opened 2022-09-13
IDEMPIERE-5354 Manage use case for microsoft OAuth2 preferred_username (FHCA-3757)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5354


CarlosRuiz-globalqss:
This is to be tested directly in FH

PR #1492, opened 2022-09-20
IDEMPIERE-5396 Replace CreateFrom form with Info Window Process

etantg:
Backport 'IDEMPIERE-5396 Replace CreateFrom form with Info Window Process' to release-9

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRMA.java
+++ b/org.adempiere.base/src/org/compiere/model/MRMA.java
@@ -170,7 +170,7 @@ public class MRMA extends X_M_RMA implements DocAction
        else
        {
            String sqlStmt = "SELECT C_Invoice_ID FROM C_Invoice WHERE C_Order_ID=?";
-           invId = DB.getSQLValueEx(null, sqlStmt, shipment.getC_Order_ID());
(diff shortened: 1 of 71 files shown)


org.adempiere.base/src/org/compiere/model/MRMA.java
```diff
@@ -47,7 +47,7 @@ public class MRMA extends X_M_RMA implements DocAction
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 6196067840638153414L;
+	private static final long serialVersionUID = -8352164928046804628L;
```
CarlosRuiz-globalqss, line 50: The line 173 of MRMA differs from the master commit:

in the master commit is get_TrxName()

in this commit is left using no transaction (null)

is that OK?

CarlosRuiz-globalqss:
@hengsin - apart from my comment above - which I don't know if is valid or harmless, I did a minor test and everything looks OK.

For me is good to go.

org.adempiere.base/src/org/compiere/model/MRMA.java
etantg: Done, please check. Thanks.

PR #1494, opened 2022-09-22
IDEMPIERE-5429 Rest API Web Service for composite not support JSON

uthadehikaru:
This is easy to solve and minor bug fix. i tested for my project and it could handle it.


hengsin:
please create the pull request against master instead of release-9.

PR #1495, opened 2022-09-22
IDEMPIERE-2205 - Ask for secret input from within a process

d-ruiz:
https://idempiere.atlassian.net/browse/IDEMPIERE-2205

## Checklist:

- [X] New and existing unit tests pass locally with my changes
- [X] I have made corresponding changes to the documentation as follows:
- - [X] Improvement (improves and existing functionality): This documentation is needed if the improvement changes the way the user interacts with the system or the outcome of a process/task changes. If it is just, for instance, a performance improvement, documentation might not be needed.: **https://wiki.idempiere.org/en/NF10_Ask_For_Secret_Input_Within_Process**
- [X] The changed/added documentation is in the project wiki (not privately-hosted pdf files or links pointing to a company website) and is complete and self-explanatory.

_The second commit is only cleaning some System.outs and fixing a java doc header. I couldn't merge into one, I think it can be done when squashing it into the core_

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
@@ -396,7 +396,6 @@ public final class Dialog {
     }
     
     public static void askForInput(int windowNo, String adMessage, String title, final Callback<String> callback, boolean isInputProtected) {
-    	System.out.println("Secretisimo : " + isInputProtected);
     	Callback<String> msgCallback = null;
     	if (callback != null) 
     	{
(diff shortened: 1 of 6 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/Dialog.java
```diff
@@ -387,7 +387,15 @@ public static void askForInput(int windowNo, String adMessage, final Callback<St
     	askForInput(windowNo, adMessage, "", callback);
     }
     
+    public static void askForSecretInput(int windowNo, String adMessage, final Callback<String> callback) {
+    	askForInput(windowNo, adMessage, "", callback, true);
```
hengsin, line 391: I think it is much cleaner that we just call the version of showDialog that takes editor and pass it the wpasswordeditor. The new showdiaog method looks a bit strange now as the isInputProtected parameter is just ignore if editor is not null. The additional isInputProtected parameter looks unnecessary too since the method accept editor parameter and you can just pass it a wpasswordeditor and has full control over it.

d-ruiz: You're right. I tried that at first and WPasswordEditor didn't have a usable constructor, I had to add it anyway so closing the pull request to create a new one with the requested changes

PR #1497, opened 2022-09-23
IDEMPIERE-5429 Rest API Web Service for composite not support JSON

uthadehikaru:
committing json format support for composite on idempiere master branch


hengsin (approved):
trivial fix, good to go

PR #926, opened 2021-10-15
IDEMPIERE-5013 Implement HikariCP as a replacement for c3p0

siepkes:
Hi all!

I order to support PostgreSQL reconnecting after failover when using client side node selection I replaced C3P0 with HikariCP. HikariCP is an actively maintained, Apache licensed connection pool with substantially better performance and better resilience to failure (DB disconnects, etc.) then C3P0. Read more about it here: https://github.com/brettwooldridge/HikariCP (including comparison and benchmarks with C3P0). During the development of HikariCP a lot of care was spend on making HikariCP behave as correctly and predictable as possible.

Let me know what you think!

Aside from HikariCP the following changes were made:

Cleaned up the `getCachedConnection` method. With HikariCP there is no need to retry to obtain a connection since getting an connection will block until a free connection is available or until a timeout is reached (default 30 seconds) at which point an `SQLException` is thrown. This also removed calling `Runtime.getRuntime().runFinalization();`. HikariCP is currently configured to detect / log leaks when a connection hasn't returned to the pool for longer then 5 minutes.

Initialization of the datasource must happen in the `getDataSource()` method because at object construction not all JDBC config is known. However this method could (as far as I could tell) be called concurrently from multiple threads but had no mechanism to prevent initializing the DB pool multiple times. The variable in which the pool itself was stored (`m_ds`) also was not marked volatile or immutable which could lead to visibility issues. Instead of lazy initialization of the pool in the `getDataSource()` method the pool could probably better be initialized at object construction. However I wasn't able to achieve that without breakage therefor I made the initialization mechanism work correctly with concurrent invocations.

The `MaxStatementsPerConnection` option was removed because HikariCP doesn't support it.

https://idempiere.atlassian.net/browse/IDEMPIERE-4984

PR #1499, opened 2022-09-26
Fixed user defined JDBC URL override not working.

siepkes:
Fixes logic error in the way a user defined `jdbcUrl` in `hikaricp.properties` is handled.

Additionally changed the default max pool size to the previous limit of 90 connections. PR #926 originally also had this but this apparently fell out. Testing showed that during first time server startup a flood of connection can be made and a connection pool of 30 can be overwhelmed in some cases (depending on the speed of your setup). This results in callers waiting on a connection which will time-out.


hengsin (approved):
looks good

PR #1500, opened 2022-09-26
IDEMPIERE-5432 - Performance Measure Calculation - Improve SQL Definition

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5432

# Pull Request Checklist

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MMeasureCalc.java
+++ b/org.adempiere.base/src/org/compiere/model/MMeasureCalc.java
@@ -39,7 +39,7 @@ public class MMeasureCalc extends X_PA_MeasureCalc implements ImmutablePOSupport
 	/**
 	 * 
 	 */
-	private static final long serialVersionUID = 3143013490477454559L;
+	private static final long serialVersionUID = -1334100963468705584L;
 
 	/**
 	 * 	Get MMeasureCalc from Cache (immutable)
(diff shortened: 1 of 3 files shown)


PeterTakacs300:
In the migration scripts I've removed 'WHERE' from the beginning of WhereClause fields in the 6 existing Performance Measure Calculation records in System.

org.adempiere.base/src/org/compiere/model/MMeasureCalc.java
```diff
@@ -143,7 +143,7 @@ public String getSqlPI (MGoalRestriction[] restrictions,
 	{
 		StringBuilder sb = new StringBuilder(getSelectClause())
 			.append(" ")
-			.append(getWhereClause());
+			.append("WHERE " + getWhereClause());
```
hengsin, line 146: It is unlikely we will accept this - a small convenience that will break backward compatibility.

norbertbede: this is breaking change. we migrate GW data, we can add to release notes.
not sure too much people used goals at all - because performance and bugs. i assume this will have small impact. wdyt ?

hengsin: I don't think it is a good idea to break backward compatibility for such small gain. We can wait for @CarlosRuiz-globalqss  opinion on this but it is NO for me.

norbertbede: what we see as problem:
we have multiple "engines" to generate data for charts, goals, status lines each of them has a bit another sql definition logic and parsing. this is one of the example where in charts we haven't  WHERE term in where field and goal has. then developer/consultant must remember different writing practice.  We would like a bit standardise with this change.

there are more issues (limitation to write modern/effective sql queries). but this is out of this PR, i will sum-up this on another forum.

CarlosRuiz-globalqss (changes requested):
see my comments about patch

hengsin (approved):
tested by carlos

PR #1502, opened 2022-09-27
IDEMPIERE-5414 - Improve Dashboard Help Icon Behaviour

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5414

# Pull Request Checklist

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
@@ -359,21 +359,29 @@ public class DashboardController implements EventListener<Event> {
 	
 	private void renderHelpButton(Caption caption, String text) {
 		ToolBarButton help = new ToolBarButton();
-		caption.appendChild(help);
+		help.setVisible(false);
 		if (ThemeManager.isUseFontIconForImage())
 			help.setIconSclass("z-icon-Help");
 		else
 			help.setImage(ThemeManager.getThemeResource(IMAGES_CONTEXT_HELP_PNG));
+		caption.appendChild(help);
 		Popup popup = new Popup();
-		popup.setSclass("dashboard-content-help");
 		popup.setPopup(popup);
 		Text t = new Text(text);
+		popup.setSclass("dashboard-content-help");
 		popup.appendChild(t);
 		help.setTooltip(popup);
-		help.addEventListener(Events.ON_CLICK, (Event event) -> {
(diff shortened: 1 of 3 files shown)


PeterTakacs300:
We think this behaviour would be more interactive, what is your opinion?

https://user-images.githubusercontent.com/93127072/192455257-09e47023-6b13-4549-b31f-326c381bf4df.mp4

hengsin:
It seems so much work just to show  simple tooltip text. Why not just set it as tooltip text for the caption component instead ?

PeterTakacs300:
We decided to implement the question mark icon to make a more concrete, better UX. And then the popup is probably more customisable than a simple tooltip, can show better structured text for longer help-texts if needed.

hengsin:
hi @PeterTakacs300 , it does looks and work just fine but it is kind of weird UI that you mouse over a button to get the popup text and it gives you the impression that it is clickable and you if you click on it, nothings happens.

PeterTakacs300:
@hengsin wdyt?



https://user-images.githubusercontent.com/93127072/193545062-c569de2a-44f5-4f67-8bf6-c30c7b78d4f2.mp4

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/DashboardController.java
```diff
@@ -358,7 +359,8 @@ private Panel newGadgetPanel(MDashboardPreference dp, MDashboardContent dc) {
 	}
 	
 	private void renderHelpButton(Caption caption, String text) {
-		ToolBarButton help = new ToolBarButton();
+		A help = new A();
```
hengsin, line 362: since you just need mouse over event, shouldn't a label works just fine here ? An A element is still a clickable element and that's not desirable here (you might have issue with :active and :visited state too)

PeterTakacs300: If you mean the org.adempiere.webui.component.Label class - it does not have the setIconSclass() method. Would you suggest to implement the setIconSclass() and setImage() methods, or extend the LabelImageElement abstract class from Label?

Not sure if, and how would these work, do you have any suggestions?

hengsin: ah ok, my mistake, label wouldn't work here.

hengsin (approved):

PR #1503, opened 2022-09-27
IDEMPIERE-5433 Separate invoices for different BPs (fixes release9)

kriegfrj:
Fixes https://idempiere.atlassian.net/browse/IDEMPIERE-5433. Now a new invoice will be created for a new ship-to BPartner as well as for a new invoice-to BP Location, so that you don't inadvertently mix ship-to BPs on the same invoice. Sorting by BPartner first, then Bill_Location_ID takes care of most of this, but as noted in the JIRA ticket there are some edge cases that are not covered.

While looking at this code, I also noticed that if you have two invoices with the same Bill_Location_ID but different Bill_User_ID, they will still consolidate onto the same invoice. But if you have two different contacts, it would probably be best if they each got their own invoice even if they are for the same location.

The only thing that is misisng from this PR (I think) is a specific regression test. All the existing tests pass, but I couldn't find a test specifically for InvoiceGenerate where I could add a regression test for this issue.  If someone can point me in the direction of the correct place to put such a test, I can consider adding one.

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/InvoiceGenerate.java
+++ b/org.adempiere.base.process/src/org/compiere/process/InvoiceGenerate.java
@@ -91,6 +91,8 @@ public class InvoiceGenerate extends SvrProcess
 	private BigDecimal p_MinimumAmtInvSched = null;
 	/**	Per Invoice Savepoint */
 	private Savepoint m_savepoint = null;
+	/** BPartner on the last order processed */
+	private int m_bpartnerID = 0;
 
 	/**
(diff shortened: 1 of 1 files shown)


kriegfrj:
Sorry I realised I based my PR off of release-9 instead of master...

kriegfrj:
> Sorry I realised I based my PR off of release-9 instead of master...

Sorry, I'll have to fix this later.

CarlosRuiz-globalqss:
Replaced by https://github.com/idempiere/idempiere/pull/1504

PR #1504, opened 2022-09-28
IDEMPIERE-5433 Better invoice consolidation rules

kriegfrj:
As #1503 but fixes the master branch.


hengsin:
Hi @kriegfrj , you can always add a new test class for this, for e.g org.idempiere.test.process.InvoiceGenerateTest

CarlosRuiz-globalqss:
Hi @kriegfrj - I think you don't need to add m_bpartner_id variable, you can consider the following patch:
[IDEMPIERE-5433.txt](https://github.com/idempiere/idempiere/files/9682879/IDEMPIERE-5433.txt)

hengsin:
I haven't test this but it seems there's something not right here in both approach, C_Invoice.C_BPartner_ID derived from C_Order.Bill_BPartner_ID, not C_Order.C_BPartner_ID (see line 356 of MInvoice).

CarlosRuiz-globalqss:
Thanks @hengsin - sorry, my fault using the wrong column (copy/paste issue).

This is my proposed patch to avoid creating the unneeded variable:
[IDEMPIERE-5433_PR2.txt](https://github.com/idempiere/idempiere/files/9685800/IDEMPIERE-5433_PR2.txt)

CarlosRuiz-globalqss:
Hi @kriegfrj - have you checked/tested the patch provided?

kriegfrj:
> Hi @kriegfrj - have you checked/tested the patch provided?

Sorry Carlos - I have been out-of-office for a couple of weeks.

I can see that you addressed the first of my concerns that I flagged in https://github.com/idempiere/idempiere/pull/1504#issuecomment-1263566416 . However, the revised patch still doesn't address the other issues that I have raised. In short, your PR doesn't actually change the current behaviour - which also means it doesn't solve the problem that I highlighted in [JIRA](https://idempiere.atlassian.net/browse/IDEMPIERE-5433):

1. Checking if the `Bill_BPartner_ID` has changed on each iteration is redundant. This is because the `if` statement already checks if `Bill_Location_ID` has changed - the `Bill_Location_ID` is pretty much guaranteed to change if the `Bill_BPartner_ID` changes (though the reverse is not the case), so it is sufficient to check the `Bill_Location_ID` only.
2. If you have two orders with the same bill-to BP (BPI), but different ship-to BPs (BPS1 & BPS2), the consolidation behaviour varies in a non-deterministic way depending on how the orders happen to be sorted. If the two orders happen to appear next to each other in the sort, then they will be consolidated onto the same invoice - but if they do not appear next to each other in the sort, then they will be consolidated onto different invoices.

I think it makes sense that different invoices should be raised for different orders if the orders' ship-to BPs are different. This seems to be the intended behaviour (which is why the query sorts the orders by ship-to BP before it sorts by bill-to BP), and that is also the behaviour that makes sense in my internal use case. But the current implementation doesn't achieve this consistently - there are some edge cases where orders with different ship-to BPs could appear on the same invoice if they have the same bill-to BP. So make this consistent, you need to check on each iteration whether the *ship-to* BP has changed too, and start a new invoice if it has. That is why it is necessary to store the last *ship to* BP as a field to compare against the ship-to BP for the current record - you can't eliminate it. You could store the last order instead (from which you can extract the last ship-to BP), but you have to store something in addition to the last invoice because the `MInvoice` doesn't record the ship-to BP.

In short, I think that my PR is still necessary.

kriegfrj:
What if I renamed `m_bpartner_id` to `m_shiptobp_id` to make it clear what the function of that field is and why it is different from `m_invoice.getC_BPartner_ID()`?

CarlosRuiz-globalqss:
Hi @kriegfrj - I created a unit test to check your issue and now is clear that your patch is the correct, can you please integrate this patch?
[IDEMPIERE-5433_PR3.txt](https://github.com/idempiere/idempiere/files/9861981/IDEMPIERE-5433_PR3.txt)

Regards,

Carlos Ruiz

kriegfrj:
> Hi @kriegfrj - I created a unit test to check your issue and now is clear that your patch is the correct, can you please integrate this patch? [IDEMPIERE-5433_PR3.txt](https://github.com/idempiere/idempiere/files/9861981/IDEMPIERE-5433_PR3.txt)
> 
> Regards,
> 
> Carlos Ruiz

Done!

CarlosRuiz-globalqss (approved):
Hi @hengsin - I think the unit test clarify the scenario that is being fixed.

Is your concern above solved?

hengsin:
hi @CarlosRuiz-globalqss , yes, looks ok to me.

kriegfrj:
Thanks! Any chance of a cherry-pick into the release-9 branch? :smile:

CarlosRuiz-globalqss:
> Thanks! Any chance of a cherry-pick into the release-9 branch? smile

sure - I'll try to integrate it in next merge

PR #1515, opened 2022-10-05
Minor - solve two new eclipse 2022-09 warnings

CarlosRuiz-globalqss:
In eclipse 2022-09 two new warnings are shown:
Project 'db' has no explicit encoding set
Project 'migration' has no explicit encoding


hengsin (approved):
trivial fix, good to go

PR #1521, opened 2022-10-08
IDEMPIERE-5445 : Fin Report - Percentage lines - Rounding

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-5445

# Pull Request Checklist

## Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules
- - [X] New feature (non-breaking change which adds functionality): I have created the New Feature page in the project wiki explaining the functionality and how to use it. If relevant, I have committed sample data to the core seed to have usable examples in GardenWorld.

### DIFF
--- a/org.adempiere.base/src/org/compiere/report/FinReport.java
+++ b/org.adempiere.base/src/org/compiere/report/FinReport.java
@@ -780,7 +780,8 @@ public class FinReport extends SvrProcess
 						}
 						// end fix bug [ 1563664 ]
 						if (m_lines[line].isCalculationTypePercent()) {
-							sb.append(" *100");
+							if (m_lines[line].isMultiplyPercentageBy100())
+								sb.append(" *100");
 
 							if (m_lines[line].getRoundingPrecision() > 0) 
 								sb.append(", ").append(m_lines[line].getRoundingPrecision()).append(")");
@@ -815,6 +816,10 @@ public class FinReport extends SvrProcess
 					{
 						if (col > 0)
 							sb.append(",");
+						
+						if (m_lines[line].isCalculationTypePercent() && m_lines[line].getRoundingPrecision() > 0)
+							sb.append("ROUND(");
+						
 						sb.append ("COALESCE(r1.Col_").append (col).append(",0)");
 						// fix bug [ 1563664 ] Errors in values shown in Financial Reports
 						// Carlos Ruiz - globalqss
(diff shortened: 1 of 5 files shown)


nmicoud:
I've added a 2Pack in the ticket to ease testing of the feature

org.adempiere.base/src/org/compiere/report/FinReport.java
```diff
@@ -763,6 +763,10 @@ else if (m_lines[line].isCalculationTypeSubtract())
 						if (col > 0)
 							sb.append(",");
 						sb.append ("Col_").append (col).append("=");
+
+						if (m_lines[line].isCalculationTypePercent() && m_lines[line].getRoundingPrecision() > 0)
```
hengsin, line 767: thinking further, if this is only for type percent, calling it rounding precision is misleading, should be named as PercentPrecision instead. Have you test what's the impact if the check for type percent is remove ?

nmicoud: Ok, I'll rename the column.

Not sure to understand "_the impact if the check for type percent is remove_" : can you elaborate more please?
I don't see why this check would be removed :-/


I discussed this morning with the customer and he asks me about another change ; I'm describing it here, but maybe is better in the ticket or forum. If so, tell me.

He asked me if it would be possible to have another field to set the 'multiplier' for percentage lines. At the moment, if the line isCalculationTypePercent, then there is a 
`sb.append(" *100");`

Let's say I have 
line 1 with value = 50
line 2 with value = 25
line 3 is the percentage line 2 / line 1 ; I'll get 50,00

He would like to have (in some cases 0.5). 
So question is about adding a field to set the multiplier for percentage lines (default would be 100)

wdyt?

Thanks,

Nicolas

hengsin: I means without the percent type check, the rounding will be applied to both amount and percentage values. I don't know whether that's desirable for your use case but just checking it out here. Or perhaps it could be a combination of PercentagePrecision and AmountPrecision.

I don't see the need to have arbitrary multiplier for percentage, it should be either 100 or nothing. For me, a IsMultiplyPercentageBy100 make more sense here.

Btw, I think your patch doesn't cover the Oracle part (there's a if (DB.isPostgreSQL()) check a few lines above).

nmicoud: My use case is just to update percent lines.
And as you said in the forum, amounts should be managed apart (or maybe in another ticket).
That's why we decide to call the column RoundingPrecision as it may be re-used later for another purpose.
But the current display logic shows it only for percent lines (it was missing in the first commit).

I may add a validation on MReportLine.beforeSave to make sure the field will be filled only for this purpose? (which should be removed/updated if the situation evolves later)

Last commit also adds IsMultiplyPercentageBy100 (default Y) and adapt the code.

And you're right, I forgot the Oracle part ; this if fixed now

If everything is ok, I'll rename the wiki page I created last week and rename it to something like "New features for FinReport Percentage lines")

hengsin: I would prefer to call it PercentRounding if it is only for percentage value. Anway, let see what's @CarlosRuiz-globalqss  take on this.

CarlosRuiz-globalqss:
Hi @nmicoud - can you please take a look to https://github.com/idempiere/idempiere/pull/1524

Following the idea of PA_ReportColumn.Factor, I added Multiplier and RoundFactor to columns and lines.
The migration script also deprecates the Factor column and uses Multiplier instead

CarlosRuiz-globalqss:
replaced with https://github.com/idempiere/idempiere/pull/1524

PR #1524, opened 2022-10-11
IDEMPIERE-5445 Implement multiplier and rounding for financial reports

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5445

# Pull Request Checklist

## Checklist:

### DIFF
--- a/migration/i9/oracle/202210111720_IDEMPIERE-5445.sql
+++ b/migration/i9/oracle/202210111720_IDEMPIERE-5445.sql
@@ -1,0 +1,238 @@
+-- IDEMPIERE-5445 Fin Report - Percentage lines - Rounding
+SELECT register_migration_script('202210111720_IDEMPIERE-5445.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Oct 11, 2022, 5:20:53 PM CEST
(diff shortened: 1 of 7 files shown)


nmicoud (approved):
Hi 

I've made some tests and I didn't find any issue.

nb: I may not be the best tester here as I'm not a specialist of Financial Reports as I'm not using them outside of this particular customer ; and I'm not the one who sets them up.

Thanks

Nicolas

CarlosRuiz-globalqss:
Thanks @nmicoud - can you please help with the wiki documentation?

I would like to know the opinion from Steven Sackett, I think this is not 100% aligned with what he recommended on forums.
Will ask him in the JIRA ticket.

nmicoud:
OK, I'll write a wiki page (and delete the previous one) - trying to do this this week

adaxa-steven:
pls see attached which tries to explain my point better.... 
[percent display error.pdf](https://github.com/idempiere/idempiere/files/9764383/percent.display.error.pdf)

also noticed that Nicolas's original problem may be handled by std functionality provided he is happy to have percentages displayed to the currency precision

nmicoud:
Hi Steven,
My original issue (which was in fact a missing functionnality) was to be able to round percentage lines. (and only those kind of lines).
I haven't tested the pattern functionnality, and from what I see in you document, it should work in my case as I have only 2 decimals on currencies.
But if you need more than 2 decimal, a patch is needed to "force" to round percentage lines and let other ones "untouched"

adaxa-steven:
also noted ... in the daily build in the report 'steven test' the value for COGS is displayed as negative with the 'amount type' set to either 'null' or to 'balance expected sign' . I assumed that a debit amount in an account 5 which has 'account type' = Expense and 'account sign' = natural would print just as a number without a negative sign.  I would then define a calculation of Margin = Revenue - COGS but I had to sum them to get the right number.  I would have expected to sum them if I was using 'accounted sign'.
When I set the amount type for COGS to 'Balance - accounted sign ' it still printed as negative even though it is a debit value.

adaxa-steven:
please see attached testing results from https://test-feature.idempiere.org:3243/webui/
[carlos percent test.pdf](https://github.com/idempiere/idempiere/files/9775856/carlos.percent.test.pdf)

nmicoud:
hi,
Sorry, I won't have time to write the documentation this week - hope to be able to do it next week (but i can't guarantee as i will be on holidays after)

CarlosRuiz-globalqss:
> please see attached testing results from https://test-feature.idempiere.org:3243/webui/ [carlos percent test.pdf](https://github.com/idempiere/idempiere/files/9775856/carlos.percent.test.pdf)

Thanks @adaxa-steven for your test case, answering the questions from your PDF:

> ... why not ordered by Key? Is there an option somewhere?

That's the info window automatically created when there is no info window for the table associated, by default is always ordered by the first column discovered, in this case the Tenant column.

The option to make it work better is to create an info window for that table.

> I have the feeling the process should not change the actual number stored if it a % regardless of whar the multiplier and the rounding factor says ... it should just leave it as some number like 35.092713456789 and then there should be a ‘format pattern’ on the line that ‘overwrites’ whatever format pattern may be on the column definition

You did it right, the configuration of 0.001 and then 1000 for the multipliers does the trick in the right way.

The numbers are stored in the database without rounding (because you didn't define rounding), so in database they look as:
1234.06789
800.99999
433.06790
35.09271276801473215500000

The problem is that you cannot define different "Format Pattern" for the same column, that's not used directly from the financial report for anything, the Format Pattern is just copied into the print format item, and you cannot have different pattern for different lines in the print format.

A potential solution could be to define a third column and in the print format define that the value is printed just for certain line.

Look how I did it on top of your test case:

ExamplePercentInOtherColumn

CarlosRuiz-globalqss:
Tested successfully by @nmicoud

PR #1530, opened 2022-10-15
IDEMPIERE-5399 Data type Time is mostly useless

CarlosRuiz-globalqss:
Fix issue reported by Nicolas when first value is 01:00
https://idempiere.atlassian.net/browse/IDEMPIERE-5399?focusedCommentId=48793


hengsin (approved):
trivial fix, good to go

nmicoud:
With some delay, ok for me too ; thanks a lot !

PR #1545, opened 2022-10-27
IDEMPIERE-5408 Allow or enforce login with specific tenant (FHCA-3823)

CarlosRuiz-globalqss:
Fix issue with EMail MFA

https://idempiere.atlassian.net/browse/IDEMPIERE-5408

* When user login with EMail MFA the MFA panel is not shown and the Processing message just keep spinning
* Also changed the message about TenantNotFound by FailedLogin to avoid a brute force attack to discover which tenants are valid in the installation


d-ruiz (approved):
Looks good

PR #1526, opened 2022-10-12
IDEMPIERE-5443 - Date Range Filter Component

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5443

# Pull Request Checklist

- [x] My code follows the [code guidelines](https://wiki.idempiere.org/en/Contributing_to_Trunk) of this project
- [x] My code follows the best practices of this project
- [x] I have performed a self-review of my own code
- [x] My code is easy to understand and review. 
- [x] In hard-to-understand areas, I have commented my code.
- [x] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have added unit tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings
- [ ] I have made corresponding changes to the documentation as follows:
- - [ ] New feature (non-breaking change which adds functionality): I have created the New Feature page in the project wiki explaining the functionality and how to use it. If relevant, I have committed sample data to the core seed to have usable examples in GardenWorld.
- [ ] The changed/added documentation is in the project wiki (not privately-hosted pdf files or links pointing to a company website) and is complete and self-explanatory.

PR #1552, opened 2022-11-03
IDEMPIERE-5470 - Date Range Fieldtype (Reference List)

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5470

# Pull Request Checklist


PeterTakacs300:
The commit contains:

- new Date Range Picker reference
- new Date Range Editor implemented for Process Parameters (demonstration video below)
- refactoring of DateRangePicker.java
- - renaming some variables, methods, small refactor of the code structure to make it more understandable for developers

The Date Range Editor:
https://user-images.githubusercontent.com/93127072/199694571-78f554db-c6c1-46b7-b227-2f0e3ddd05d0.mp4

hengsin:
Hi @PeterTakacs300 ,

- This is a rather hackish and incomplete implementation. As it is, it is not yet fit for integration to core.
- The change to DateRangePicker doesn't depends on DateRangeEditor and I think it is better to move that to IDEMPIERE-5467. I think the DateRangePicker changes by itself is ready for integration to core.
- 
Regards,
Low

norbertbede:
Hi @igorpojzl  please take a look into Peter's work. what is the problem with this implementation

hengsin:
- dateRangeEditor at WEditor is hackish
- Doesn't follow existing convention where editor is instantiated through editor factory using reference type
- Incomplete as this reference type while visible, wouldn't work for AD window and Info window

PeterTakacs300:
Hi @hengsin, 
- _"dateRangeEditor at WEditor is hackish"_ - removed all the hacks and changes from WEditor
  - the DateRangeEditor is now mapped in ProcessParameterPanel instead of being saved in WEditor
  - the ValueChangeEvent is now fired from setValue() methods of WDateEditor and WDateTimeEditor (similar to other editors) 
- _"Doesn't follow existing convention where editor is instantiated through editor factory using reference type"_
  - in the conversation in https://github.com/idempiere/idempiere/pull/1526 PR at the beginning of the implementation of the new component we talked about the Date Range Editor, and you mentioned, _"I think it is not a must that it has to be a WEditor component itself"_ - so we went this way, because if I understand correctly, we agreed on that if the DateRangeEditor would extend WEditor, it would be a too complicated implementation. Since DateRangeEditor is now not a WEditor, it can not be instantiated through editor factory - please correct me if I'm wrong, or tell us if you have an idea for this solution
- _"Incomplete as this reference type while visible, wouldn't work for AD window and Info window"_ - yes, probably fist I should've marked this PR as 'POC' - it would be good if you reviewed the concept for the process parameters, and after the concept is clear and approved, we would implement it to other places and commit to this PR

On top of this, I implemented a new feature to DateRangePicker, which relates mostly to the new DateRangeEditor:
- when Default Logic is set, the DateRangePicker tries to find the common ranges, which can be converted to "human readable" string
  - e.g. today is 2022-11-09, Default Logic = '2022-10-01', Default Logic 2 = '2022-10-31' - the DateRangePicker should recognise, and transform this range to display value "Previous Month" - and also set the DateRangePicker mode dropdown to "Last", offset to 1 and unit dropdown to "Month"
  - fallback mode is "Between"
  - may have some bugs because of the large number of possible combinations

https://user-images.githubusercontent.com/93127072/200864139-eab6a9a4-cca1-42a7-83ed-711d40795b8e.mp4

hengsin:
hi @PeterTakacs300 ,
- Since it is not a WEditor component, I don't think it has to be implemented as a reference/display type. There's no advantage doing that and doesn't feel natural.
- I don't see how the new date range picker feature has any dependency to date range editor. I think it can be implemented using the isRange approach too.
- My suggestion is to move the Date Range Picker changes to the Info Window IsRange ticket. The same can be implemented for process parameter using IsRange.

PeterTakacs300:
OK, so I am moving the Date Range Picker changes to the Info Window IsRange ticket - that include the new auto-detection feature too

hengsin (changes requested):
to address what have been agree in ticket.

PeterTakacs300:
Hi @hengsin, I've replaced the 'IsDisplayDateInWords' with the 'DateRangeOption' list, and added the agreed values to it.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -1142,6 +1166,25 @@ private void dynamicDisplay() {
...
+						m_separators.get(i).setVisible(false);
+						m_wEditors2.get(i).setVisible(false ,true);
+					}
+					dateRangeEditor.setFieldMandatoryStyle();
+					dateRangeEditor.setReadOnly(!(editor.isReadWrite() && m_wEditors2.get(i).isReadWrite()));
+				}
+			}
 		}
 		if (getParent() != null) {
 			getParent().invalidate();
```
hengsin, line 1190: should be implemented for info window too.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangePicker.java
```diff
@@ -1076,4 +1076,4 @@ public void valueChange(ValueChangeEvent evt) {
 			updateUI();
 		}
 	}
-}
+}
```
hengsin, line 1079: would be better to remove this type of unnecessary white space only changes.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+*                                                                     *
+* You should have received a copy of the GNU General Public License   *
+* along with this program; if not, write to the Free Software         *
+* Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,          *
+* MA 02110-1301, USA.                                                 *
+*                                                                     *
+* Contributors:                                                       *
+* - Peter Takacs, Cloudempiere                                        *
+**********************************************************************/
+package org.adempiere.webui.window;
```
hengsin, line 25: this probably should goes into the org.adempiere.webui.editor package instead.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+
+	/**
+	 *
+	 */
+	private static final long serialVersionUID = 3329360781729704243L;
+
+	private static final String IMAGES_CONTEXT_CALENDAR_PNG = "images/Calendar16.png";
+
+	private Textbox textbox;
+	ToolBarButton popupBtn;
```
hengsin, line 54: most of the time we should use protected instead of package visibility for instance variables.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+		popupBtn = new ToolBarButton();
+		popupBtn.addSclass("date-picker-calendar-button");
+		if (ThemeManager.isUseFontIconForImage())
+			popupBtn.setIconSclass("z-icon-Calendar");
+		else
+			popupBtn.setImage(ThemeManager.getThemeResource(IMAGES_CONTEXT_CALENDAR_PNG));
+		this.appendChild(popupBtn);
+
+		popup = new DateRangePicker(editor, editor2);
+		popupBtn.setTooltip(popup);
```
hengsin, line 89: this doesn't looks right ...

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	private boolean isMandatoryStyle() {
+		return editor.isMandatory() && editor.isReadWrite() && (isProcessParameter || editor.getGridField().isEditable(true)) && isNullOrEmpty(editor)
+				|| editor2.isMandatory() && editor2.isReadWrite() && (isProcessParameter || editor2.getGridField().isEditable(true)) && isNullOrEmpty(editor2);
+	}
+
+	private boolean isNullOrEmpty(WEditor editor) {
+		Object value = editor.getValue();
+		return value == null || value.toString().trim().length() == 0;
```
hengsin, line 128: should use Util.isEmpty(value.toString(), true) instead.

org.adempiere.ui.zk/WEB-INF/src/web/theme/default/css/fragment/window.css.dsp
```diff
@@ -153,3 +153,9 @@ input[type="checkbox"]:focus
 	font-weight: bold;
 	margin: 5px;
 }
+
+.date-picker-calendar-button {
+	position: absolute;
+    right: 0px;
+    top: 5px;
+}
```
hengsin, line 161: again ... please update idempiere.theme.default in lang-addon.xml when you  make changes to css ....

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+	public void setFieldMandatoryStyle() {
+		if (isMandatoryStyle())
+			LayoutUtils.addSclass("idempiere-mandatory", textbox);
+		else
+			LayoutUtils.removeSclass("idempiere-mandatory", textbox);
+	}
+
+	private boolean isMandatoryStyle() {
+		return editor.isMandatory() && editor.isReadWrite() && (isProcessParameter || editor.getGridField().isEditable(true)) && isNullOrEmpty(editor)
+				|| editor2.isMandatory() && editor2.isReadWrite() && (isProcessParameter || editor2.getGridField().isEditable(true)) && isNullOrEmpty(editor2);
```
hengsin, line 123: this is a strange one - why always mandatory if it is process parameter ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+	private Textbox textbox;
+	ToolBarButton popupBtn;
+	private DateRangePicker popup;
+	private boolean isProcessParameter;
+	private WEditor editor;
+	private WEditor editor2;
+
+	/**
+	 * Constructor
+	 */
```
hengsin, line 62: wrong javadoc, all @param is missing.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+			popup.open(popupBtn, "after_center");
+			LayoutUtils.autoDetachOnClose(popup);
+		});
+
+		Button btn = popup.getUpdateButton();
+		btn.addEventListener(Events.ON_CLICK, e -> {
+			textbox.setValue(popup.getDisplayValue());
+		});
+		editor.addValueChangeListener(this);
+		editor2.addValueChangeListener(this);
```
hengsin, line 101: more natural way is to fire valuechange event from DateRangePicker and register this class as valuechange listeners for daterangepicker

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+		popupBtn.addEventListener(Events.ON_CLICK, event -> {
+			popup.setPage(popupBtn.getPage());
+			popup.open(popupBtn, "after_center");
+			LayoutUtils.autoDetachOnClose(popup);
+		});
+
+		Button btn = popup.getUpdateButton();
+		btn.addEventListener(Events.ON_CLICK, e -> {
+			textbox.setValue(popup.getDisplayValue());
+		});
```
hengsin, line 99: hmm .. so you are listening for both editor value change and ok btn on_click and they seems to do just the same thing. Also, the seemingly duplicate event handling aside, if the changes to editor & editor2 is immediate, why do we need another ok button here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
```diff
@@ -0,0 +1,142 @@
...
+
+		popup = new DateRangePicker(editor, editor2);
+		popupBtn.setTooltip(popup);
+		popupBtn.addEventListener(Events.ON_CLICK, event -> {
+			popup.setPage(popupBtn.getPage());
+			popup.open(popupBtn, "after_center");
+			LayoutUtils.autoDetachOnClose(popup);
+		});
+
+		Button btn = popup.getUpdateButton();
```
hengsin, line 96: the more natural way of doing this is to register this class as listener or callback to DateRangePicker and DateRangePicker will invoke the register listener or callback when it is dismiss by user. The current approach is not good since the UI design of DateRangePicker can change and DateRangePicker can be dismissed by user through other means (for e.g, keyboard shortcut).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangeEditor.java
PeterTakacs300: I think it is unnecessary, left there by mistake

PeterTakacs300:
Hi @hengsin, except the info window implementation, I think I covered all the requested changes - not sure though, if it is valid if I call the fireValueChange(evt) method inside the valueChange() method of DateRangePicker, but I could not invoke any issues yet wit this implementation.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/ProcessParameterPanel.java
```diff
@@ -1178,6 +1205,25 @@ private void dynamicDisplay() {
...
+				DateRangeEditor dateRangeEditor = m_dateRangeEditors.get(i);
+				if(dateRangeEditor != null) {
+					dateRangeEditor.setVisible(editor.isVisible());
+					m_Rows.get(i).setVisible(editor.isVisible());
+					m_Rows.get(i).setAttribute(Group.GROUP_ROW_VISIBLE_KEY, editor.isVisible());
+					editor.setVisible(false, true);
+					if (mField.getVO().isRange) {
+						m_separators.get(i).setVisible(false);
+						m_wEditors2.get(i).setVisible(false ,true);
+					}
```
hengsin, line 1222: I think 1218-1222 is probably redundant since if date range editor is use, editor and editor2 would have already been set invisible before this.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/DateRangePicker.java
```diff
@@ -112,17 +112,28 @@ public class DateRangePicker extends Popup implements EventListener<Event>, Valu
...
+	/** value of date to **/
 	private Date dateTo;
-	private Date oldValueFrom;
-	private Date oldValueTo;
+	/** current/new display value **/
 	private String displayValue;
+	/** old display value **/
+	private String oldDisplayValue = "";
+	/** enable to fire value change listeners **/
 	private boolean enableValueChange = true;
```
hengsin, line 128: took me sometime to figure out that the javadoc comment for this variable is wrong. you are setting enableValueChange to false when you don't want to listen for ValueChangeEvent from editor and editor2.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/DateRangeEditor.java
```diff
@@ -0,0 +1,136 @@
...
+	public void setFieldMandatoryStyle() {
+		if (isMandatoryStyle())
+			LayoutUtils.addSclass("idempiere-mandatory", textbox);
+		else
+			LayoutUtils.removeSclass("idempiere-mandatory", textbox);
+	}
+
+	private boolean isMandatoryStyle() {
+		return editor.isMandatory() && editor.isReadWrite() && editor.getGridField().isEditable(true) && isNullOrEmpty(editor)
+				|| editor2.isMandatory() && editor2.isReadWrite() && editor2.getGridField().isEditable(true) && isNullOrEmpty(editor2);
```
hengsin, line 117: please use bracket to make this easier to read

hengsin:
[pr1552.patch.txt](https://github.com/idempiere/idempiere/files/10781848/pr1552.patch.txt)

@PeterTakacs300 , please merge the attached patch.

About the style of the "Apply" button - if your project needs a specific style, please add a new sclass for that instead of hard coding a setstyle call that make it different from all other buttons in the apps.

hengsin:
hi @CarlosRuiz-globalqss , this tested fine for me, WDYT ?

hengsin:
[pr1552.patch1.txt](https://github.com/idempiere/idempiere/files/10789350/pr1552.patch1.txt)

Changes:
- display date range text (for e.g "02/01/2023  - 02/28/2023") when display text doesn't include date range (i.e "Current Month", "Next 2 Months", etc).

hengsin (approved):
tested

CarlosRuiz-globalqss (approved):
tested successfully

CarlosRuiz-globalqss:
@PeterTakacs300 - I think you are using a different version of the model generator than our version in core, because every time you generate X/I files some order changes.  Can you please check and sync your model generator with latest in core?

PR #1557, opened 2022-11-11
IDEMPIERE-870 Fixed Assets using non-existant context variable

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-870

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAssetAddition.java
+++ b/org.adempiere.base/src/org/compiere/model/MAssetAddition.java
@@ -434,12 +434,9 @@ public class MAssetAddition extends X_A_Asset_Addition
 	 */
 	public void setIsApproved()
 	{
-		if(!isProcessed())
+		if(!isProcessed() && MRole.getDefault().isCanApproveOwnDoc())
 		{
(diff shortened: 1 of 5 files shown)


org.adempiere.base/src/org/compiere/model/MAssetAddition.java
```diff
@@ -433,12 +433,9 @@ public void setSourceAmt(BigDecimal amt)
 	 */
 	public void setIsApproved()
 	{
-		if(!isProcessed())
+		if(!isProcessed() && MRole.getDefault().isCanApproveOwnDoc())
 		{
-			String str = Env.getContext(getCtx(), "#IsCanApproveOwnDoc");
```
hengsin, line 438: Perhaps we should add this variable instead, could be useful ?

mschnbeck: I think so, too. I may be useful in other circumstances, too.

CarlosRuiz-globalqss: Done, implemented in last commit

hengsin (approved):
looks good.

PR #1558, opened 2022-11-11
IDEMPIERE-5475 Fixed Assets - NPE when creating an asset addition without invoice

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5475


hengsin (approved):
trivial fix, good to go.

PR #1561, opened 2022-11-14
IDEMPIERE-5136 Set Messages at tenant level / Fix NPE

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5136?focusedCommentId=48852

Fix Null Pointer Exception when saving a message without filling Value


hengsin (approved):
minor fix

PR #1564, opened 2022-11-15
IDEMPIERE-5013 set HikariCP initial pool size to 10 - as it was in c3p0

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5013?focusedCommentId=48859

c3p0 had initial pool size set to 10.

HikariCP suggest a fixed pool size approach, so it starts immediately the 90 connections.

Changing the default back to 10 so it mimics better our previous c3p0 behavior.


hengsin (approved):
trivial change, good to go

PR #1565, opened 2022-11-16
IDEMPIERE-5472 Unclosed resources

etantg:
Close the unclosed resources to avoid resource leaks


hengsin (changes requested):

hengsin (approved):
looks good, pass unit test.

PR #1572, opened 2022-11-18
IDEMPIERE-5482 - Field Style Scope HTML Element

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5482

# Pull Request Checklist


hengsin (changes requested):
this design is inappropriate , ad_style can be applied to place more than just info window cell.

PeterTakacs300:
Hi @hengsin, would this approach be acceptable, if we would implement it in other places where AD_Style can be applied?

hengsin:
hi @PeterTakacs300, it make more sense to just add it to info window. It is not appropriate to just assume span and table cell works for all places where ad_style is being used.

PeterTakacs300:
Hi @hengsin, we came up with a slightly different approach.

After research, I've noticed that when you switch to grid view in regular windows, the ad_style is applied already properly, since there is always a span in the td element.

The new approach: we are trying to unify the application of ad_style based on this window-grid view:
- the code always adds a wrapper span in info window and in reports, when we want to apply ad_style (from info column and from print format item)
- We've removed the AD_StyleScope field, since it is useless with this new approach

In this case, the ticket name and description should be also updated.

hengsin:
[pr1572.patch1.txt](https://github.com/idempiere/idempiere/files/11201390/pr1572.patch1.txt)
- shouldn't call mstyle.get if ad_fieldstyle_id is 0.

PeterTakacs300:
Thanks @hengsin, tested and applied

hengsin (approved):
tested

PR #1594, opened 2022-11-30
IDEMPIERE-5505 : Add Value and PrintName on GL_Category

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-5505

# Pull Request Checklist

## Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/migration/iD10/oracle/202211301123_IDEMPIERE-5505.sql
+++ b/migration/iD10/oracle/202211301123_IDEMPIERE-5505.sql
@@ -1,0 +1,88 @@
+-- IDEMPIERE-5505
+SELECT register_migration_script('202211301123_IDEMPIERE-5505.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Nov 30, 2022, 11:23:47 AM CET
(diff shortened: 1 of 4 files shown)


hengsin:
Some thing to fix/consider here - this break the existing convention that Value(Search Key) is mandatory.

nmicoud:
you're right.

I see 2 options : 
 - add an update to set Value = GL_Category_ID and then set the column as Mandatory
 - change the name from "Value" to something else (my first idea was "Code", but Carlos suggested Value as it was the convention ; so maybe CategoryCode or something like that)

wdyt?

CarlosRuiz-globalqss:
Yes, the Value column is usually mandatory and unique per tenant.

In principle I think you can set it to be equal as Name, that's the common setup in most sample tables.

nmicoud:
Ok, I'll adapt the script and add an unique index on it.
I'll add it on the Migration Notes (as the script may fail because of that)

nmicoud:
Added an update to populate Value before making the column mandatory.
And I don't think the migration note is needed as only the first 40 characters are used ; so it **should** not fail.
wdyt?

nmicoud:
So a migration note should be enough ? 
I don't think we can found GL_Category.Name with lenght > 40
wdyt?
Another possibility would be to take 30 first characters of name and add the GL_Category_ID

CarlosRuiz-globalqss:
I think is better to make the Value column same length as Name

A migration note sounds OK for me, it's still possible to get duplicates in the Name because there is no index to prevent that, but in real life that would be very unlikely.

nmicoud:
Value length increased to 60
Just seen that there is an index "gl_category_name", so we can't find duplicate names.
I also realize I forgot to add a "gl_category_value" index, I'll do this asap

nmicoud:
done !

nmicoud:
hi @CarlosRuiz-globalqss ,
Done,
Thanks

CarlosRuiz-globalqss (approved):
tested

PR #1610, opened 2022-12-21
IDEMPIERE-5238 - Record_ID editor

PeterTakacs300:
JIRA Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5238

# Pull Request Checklist


PeterTakacs300:
Showcasing the new Related To editor and dialog:
https://user-images.githubusercontent.com/93127072/208875808-4f45fbef-afd9-45ae-8335-ca124389b60a.mp4

Zoom button (legacy functionality):
https://user-images.githubusercontent.com/93127072/208876122-0fb48027-51a4-4647-9360-d8d3d56141a7.mp4

Itt seems to work fine in windows, except in Grid Mode. There is probably a limitation in core, where we need help or advice. 

The problem is: we can not render the Identifier string of the related records, because it would need the AD_Table_ID from each line, but we have access only to the AD_Table_ID from the line, on which we opened the grid mode. It is demonstrated here with an example, where we have even the same Record_ID-s with different tables:

https://user-images.githubusercontent.com/93127072/208877740-7ad184b0-4065-4704-8c9a-0687e308b220.mp4

The question is, how can we access the correct AD_Table_ID for each row? Maybe a change will be needed for this in the engine, which loads/renders the data in grid mode?

hengsin (changes requested):
please address the change promise in the ticket

PeterTakacs300:
Hi @hengsin, the editor and field type is renamed from 'Related To' to 'Record ID'

org.adempiere.base/src/org/compiere/util/DisplayType.java
```diff
@@ -174,6 +175,10 @@
 	public static final int ChosenMultipleSelectionTable = REFERENCE_DATATYPE_CHOSEN_MULTIPLE_SELECTION_TABLE;
 	
 	public static final int ChosenMultipleSelectionSearch = REFERENCE_DATATYPE_CHOSEN_MULTIPLE_SELECTION_SEARCH;
+	
+	public static final int RelatedTo = REFERENCE_DATATYPE_RELATED_TO;
```
hengsin, line 179: hi @PeterTakacs300 , the rename of RelatedTo to RecordId is incomplete.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/util/DisplayType.java
PeterTakacs300: Thank you, renaming completed.

CarlosRuiz-globalqss:
Hi @PeterTakacs300 - I'm doing a test adding AD_Table_ID+Record_ID on Test table

When clicking on the Zoom button with Record_ID empty it throws a NPE:
```
11:59:23.278===========> UiEngineImpl.error:  [185]
java.lang.NullPointerException
	at org.adempiere.webui.editor.WRecordIDEditor.actionZoom(WRecordIDEditor.java:168)
	at org.adempiere.webui.editor.WRecordIDEditor.onEvent(WRecordIDEditor.java:291)
	at org.zkoss.zk.ui.AbstractComponent.onEvent(AbstractComponent.java:3191)
	at org.zkoss.zk.ui.AbstractComponent.service(AbstractComponent.java:3161)
```

CarlosRuiz-globalqss:
Some additional strange behavior:

1 - I filled AD_Table_ID with AD_Element - and then selected one element, f.e. Calculation Type
Then I change the Table to AD_Reference, the field Record_ID is cleared, but when saving it still saves 1605 in Record_ID -> I think the field Record_ID must be dependent of AD_Table and cleared when AD_Table changes (sometimes it hits a record that exists and change the description accordingly, but that sounds dangerous)

2 - Following same test, then I click on Related Record and change the table to AD_Message (inside the Dialog) and it shows me an error "Lookup cannot be null"

3 - The Zoom function on label (and the button Related Record ID) are not checking for permission on the related table, if is not too difficult it would be good to make it zoomable just when the user has permission on tables - not a big deal if is too hard

4 - I enabled Test table to be used in All tenants - and ran Role Access Update
I have a second tenant with C_AcctSchema_ID[1000000] =client1 UN/35 US Dollar
In GardenWorld I create a new freight category "Test" and it has M_FreightCategory_ID=1000000
Then in GW test I select M_FreightCategory and the new record test - so it shows "Test"
Then I go to the table and change it to C_AcctSchema - and it shows me "client1 UN/35 US Dollar" <- this is a cross-tenant information - if I click on Zoom - it shows error "Cross tenant PO reading request detected from session 1000150 for table C_AcctSchema Record_ID=1000000"
When I save it and reopen - the field is empty and the record is shown unsaved, and trying to save it shows a "SaveErrorNoChange" error

5 - I executed the following SQL in my test DB:
```SQL
UPDATE ad_column SET ad_reference_id = 200202
WHERE  columnname='Record_ID' AND ad_reference_id = 28 AND ad_column_id < 1000000;

UPDATE ad_field SET xposition=xposition-1
WHERE ad_field_id < 10000000 AND ad_column_id IN (SELECT ad_column_id FROM ad_column WHERE  columnname='Record_ID' AND ad_reference_id = 200202);
```
Then I created a Sales Order in GW and completed it, then I navigate to Active Workflows using the toolbar button - for some reason the Record_ID is shown empty and the record is marked as modified to be saved, if I click the Undo Changes button, it shows me the Order correctly

PeterTakacs300:
Hi @CarlosRuiz-globalqss, thank you for the testing

The NPE, and point 1, 2, 4 should be fixed - I plan to check the rest of it until Monday. Till then perhaps you could take a look at these fixes so we are one step further.

PeterTakacs300:
Providing the remaining fixes including checking for permission on the related table on zoom event

CarlosRuiz-globalqss:
Hi @PeterTakacs300 - retested this:

My test scenario:

Added AD_Table_ID+Record_ID on Test table.  Both new fields are updateable.

This issue is still not solved:

1 - I filled AD_Table_ID with AD_Element - and then selected one element, f.e. Calculation Type
Then I change the **window field**  Table to AD_Reference, the field Record_ID is cleared, but when saving it still saves 1605 in Record_ID -> I think the field Record_ID must be dependent of AD_Table and cleared when AD_Table changes (sometimes it hits a record that exists and change the description accordingly, but that sounds dangerous)

Please note that I'm talking about changing the window field for the table, not the field withing the dialog.

---

This is not implemented:

3 - The Zoom function on label (and the button Related Record ID) are not checking for permission on the related table, if is not too difficult it would be good to make it zoomable just when the user has permission on tables - not a big deal if is too hard

---

The 4th changed behavior, now is showing cross-tenant information (security issue):

4 - I enabled Test table to be used in All tenants - and ran Role Access Update
I have a second tenant with C_AcctSchema_ID[1000000] =client1 UN/35 US Dollar
In GardenWorld I create a new freight category "Test" and it has M_FreightCategory_ID=1000000
Then in GW test I select M_FreightCategory and the new record test - so it shows "Test"
Then I go to the table and change it to C_AcctSchema - and it shows me "client1 UN/35 US Dollar" <- this is a cross-tenant information - if I click on Zoom - it shows error "Cross tenant PO reading request detected from session 1000150 for table C_AcctSchema Record_ID=1000000"
When I save it and reopen - the field is **NOT empty** and **is showing the information from the other tenant**

---

The 0, 2 and 5 are solved.

PeterTakacs300:
Hi @CarlosRuiz-globalqss, commit summary:

issue 1 - should be fixed

issue 3 - it is implemented in WRecordIDEditor:163-164 (in method actionZoom())

issue 4 - after fixing issue 1 the Record ID is always set to null when changing AD_Table_ID - looks like this solves issue 4, I couldn't reproduce it after the fix

CarlosRuiz-globalqss:
Hi @PeterTakacs300, thanks for your work on this.

New round of tests:

1 - is solved

3 - OK - solved showing the error

4 - is solved in UI via 1 as you mentioned , but still open for other means of filling the Record-ID with cross-tenant information, I did a test importing a CSV file with table C_AcctSchema and RecordID=1000002 and it shows me the information from another tenant, as this field becomes editable and showing information is better to ensure that any cross-tenant scenario is forbidden within this field4

New issues found:

6 - When I open the Test window in GardenWorld, it shows the Record ID empty and the save button enabled, if I click the Undo button then the description appears in Record ID field.  Same happens if I open the Test window in System.

7 - error when I select AD_Role and System Administrator - it saves NULL in the database (instead of zero) and it shows "156" in the description, although is null (156 is the table ID for AD_Role)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	@Override
+	public String getDisplay() {
+		if(((Integer) tableIDGridField.getValue() != null) && ((Integer) gridField.getValue() != null))
+			return getIdentifierColumns((Integer) tableIDGridField.getValue(), (Integer) gridField.getValue());
+		else
+			return "";
+	}
+	
+	@Override
+    public String getDisplayTextForGridView(Object value) {
```
hengsin, line 291: This wouldn't work for grid view since you can't get the row's table id value. A possible approach to fix this is to add GridRowCtx parameter and you can get the row's table id value with `String key = gridTab.getWindowNo + "|AD_Table_ID"; Object rowTableIdValue = gridRowCtx.get(key);`

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	/**
+	 * Get Identifier String from AD_Table_ID and Record_ID
+	 * @param tableID
+	 * @param recordID
+	 * @return String
+	 */
+	public static String getIdentifierColumns(int tableID, int recordID) {
+		String value = String.valueOf(tableID);
```
hengsin, line 331: default should be "" or null.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	public void actionRefresh(Object value)
+    {
+		setValue(value);
+    }
+	
+	@Override
+	public void actionZoom()
+	{
+		Integer tableID = (Integer) tableIDGridField.getValue();
+		Integer recordID = (Integer) gridTab.getValue("Record_ID");
```
hengsin, line 160: becarefull with such casting, you will get NPE if either AD_Table_ID or Record_ID value is null.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	private GridTabDataBinder dataBinder;
+	
+	private Div wrapperDiv;
+	private Textbox recordTextBox;
+	private ToolBarButton editButton;
+	private ToolBarButton zoomButton;
+	
+	protected String selectedTableName = "";
+	protected String selectedRecordIdentifyer = "";
+
```
hengsin, line 83: looks like both variable above is not use anywhere.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	private static final String IMAGES_CONTEXT_ZOOM_PNG = "images/Zoom16.png";
+	private static final String IMAGES_CONTEXT_EDIT_RECORD_PNG = "images/EditRecord16.png";
+
+	public WRecordIDEditor(GridField gridField, boolean tableEditor, IEditorConfiguration editorConfiguration) {
+		super(new Vlayout(), gridField, tableEditor, editorConfiguration);
+		init();
+	}
+
+	private void init() {
+		wrapperDiv = new Div();
```
hengsin, line 93: is this wrapperDiv really needed here ? why not use new Div() in the constructor above and apply the recordid-editor sclass to it ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		wrapperDiv.setSclass("recordid-editor");
+		wrapperDiv.setParent(this.getComponent());
+		wrapperDiv.addEventListener(Events.ON_RIGHT_CLICK, this);
+
+		if(gridTab != null) {
```
hengsin, line 98: I guess this editor wouldn't work at all if gridTab is null ? if that's the case, should just throw some sort of runtime exception when gridTab is null.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			tableIDGridField = gridTab.getField("AD_Table_ID");
```
hengsin, line 99: should throw runtime exception if tableIDGridField is null

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			tableIDEditor = WebEditorFactory.getEditor(tableIDGridField, true);
+			
+			if(tableIDGridField.getValue() != null)
+				tableIDEditor.setValue(tableIDGridField.getValue());
+						
+			dataBinder = new GridTabDataBinder(gridTab);
+			tableIDEditor.addValueChangeListener(dataBinder);
```
hengsin, line 106: why do we need to do this here ? this should the job of the field editor for AD_Table_ID field.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			tableIDGridField.addPropertyChangeListener(new PropertyChangeListener() {
+				@Override
+				public void propertyChange(PropertyChangeEvent evt) {
+					if(evt.getNewValue() != oldTableID) {
+						oldTableID = evt.getNewValue();
+						setValue(null);
+					}
+				}
+			});
+			oldTableID = tableIDGridField.getValue();
```
hengsin, line 118: again, shouldn't deal with AD_Table_ID change here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+	
+	public void actionRefresh(Object value)
+    {
+		setValue(value);
+    }
+	
+	@Override
+	public void actionZoom()
+	{
+		Integer tableID = (Integer) tableIDGridField.getValue();
```
hengsin, line 159: wouldn't gridTab.getValue("AD_Table_ID") works here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		Integer recordID = (Integer) gridTab.getValue("Record_ID");
+		if(tableID == null || tableID <= 0 || recordID == null || recordID <= 0)
+			return;
+		if (!MRole.getDefault().isTableAccess(tableID, false))
+			throw new AdempiereException(Msg.getMsg(Env.getCtx(), "AccessTableNoView"));
+		MTable savedTable = new MTable(Env.getCtx(), tableID, null);
```
hengsin, line 165: that's a confusing variable name here. you got nothing to save here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+
+		if (AD_Window_ID > 0) {
+			if(!Util.isEmpty(keyColumn)) {
+				MQuery zoomQuery = new MQuery();
+				zoomQuery.addRestriction(keyColumn, MQuery.EQUAL, (Integer)po.get_Value(keyColumn));
+				zoomQuery.setRecordCount(1);
+	
+				AEnv.zoom(AD_Window_ID, zoomQuery);
+			}
+		}
```
hengsin, line 192: don't really understand why not use use AEnv.zoom(AD_Table_ID, Record_ID) here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		if (value == null || value.toString().trim().length() == 0)
+		{
+			oldValue = null;
+			value = null;
+			recordTextBox.setValue("");
+		}
+		else {
+			if(oldTableID != tableIDGridField.getValue()) {
+				oldTableID = tableIDGridField.getValue();
+			}
```
hengsin, line 259: shouldn't gridTab.getValue("AD_Table_ID") work here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+
+	@Override
+	public Object getValue() {
+		return oldValue;
+	}
+
+	@Override
+	public String getDisplay() {
+		if(((Integer) tableIDGridField.getValue() != null) && ((Integer) gridField.getValue() != null))
+			return getIdentifierColumns((Integer) tableIDGridField.getValue(), (Integer) gridField.getValue());
```
hengsin, line 285: again, use gridTab.getValue("AD_Table_ID"). for convenience, you can add a private getAD_Table_ID method that did that and return int (and handling the problem of null and Integer casting inside).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
```diff
@@ -749,7 +750,7 @@ private void setTableAndKeyColumn() {
 			} else {
 				m_tableName = m_keyColumnName.substring(0, m_keyColumnName.length()-3);
 			}
-		} else if (getGridField() != null && getGridField().getGridTab() != null && getGridField().getAD_Column_ID() > 0) {
+		} else if (getGridField() != null && getGridField().getGridTab() != null && getGridField().getAD_Column_ID() > 0 && !Util.isEmpty(m_keyColumnName)) {
```
hengsin, line 753: this might break existing use case, need detail verification

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+		MLookup recordsLookup = getRecordsLookup();
+    	recordsEditor = recordsLookup != null ? new WSearchEditor("Record_ID", false, false, true, recordsLookup) : new WSearchEditor(editor.getGridField());
+    	parentTextBox = new Textbox();
+    	
+		setPage(page);
+		setClosable(true);
+		setTitle(Msg.getMsg(Env.getCtx(), "ChooseRelatedRecord"));
+		setPosition("center");
+		setSclass("recordid-dialog");
+		doHighlighted();
```
hengsin, line 114: doHighlighted() call should be the last line of the method.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+		labelsDiv = new Div();
+		fieldsDiv = new Div();
+		contentDiv = new Div();
+		confirmPanelDiv = new Div();
+    	okBtn = ButtonFactory.createNamedButton(ConfirmPanel.A_OK, true, true);;
+    	cancelBtn = ButtonFactory.createNamedButton(ConfirmPanel.A_CANCEL, true, true);;
+    	tableIDEditorCopy = new WTableDirEditor("AD_Table_ID", false, false, true, tableIDGridField.getLookup(), true);
+		tableIDEditorCopy.setValue(tableIDGridField.getValue());
+		MLookup recordsLookup = getRecordsLookup();
+    	recordsEditor = recordsLookup != null ? new WSearchEditor("Record_ID", false, false, true, recordsLookup) : new WSearchEditor(editor.getGridField());
```
hengsin, line 106: A WSearchEditor without lookup ? Don't get what's the possible use case here ...

hengsin:
hi @PeterTakacs300 , please document your code better, it is really hard to review now, thanks.

hengsin:
Is this editor means to edit two field (AD_Table_ID and Record_ID) and the field AD_Table_ID means to be make invisible ?

hengsin:
Carlos have been using the Test window to test this, I think we should add that to this PR to illustrate the intended usage of this editor.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+	private WTableDirEditor tableIDEditorCopy;
+	
+	private Button okBtn;
+	private Button cancelBtn;
+	private Textbox parentTextBox;
+	private WSearchEditor recordsEditor;
+	
+	private Integer value;
+	
+	public WRecordIDDialog(Page page, WRecordIDEditor editor, GridField tableIDGridField) {
```
hengsin, line 81: no javadoc

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+		appendChild(vLayout);
+	}
+	
+	@Override
+	public void onEvent(Event event) throws Exception {
+		if(event.getName().equalsIgnoreCase(Events.ON_CLICK)) {
+			if(event.getTarget().equals(okBtn)) {
+				value = (Integer) recordsEditor.getValue();
+				tableIDGridField.getGridTab().setValue("AD_Table_ID", tableIDEditorCopy.getValue());
+				editor.setValue(value);
```
hengsin, line 164: should validate here either both AD_Table_ID and Record_ID is fill with a valid value or both is null.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+	
+	private Page page;
+	private WRecordIDEditor editor;
+	private GridField tableIDGridField;
+	
+	private Div contentDiv;
+	private Div labelsDiv;
+	private Div fieldsDiv;
+	private Div confirmPanelDiv;
+	private WTableDirEditor tableIDEditorCopy;
```
hengsin, line 72: more concise and less confusing to just call this tableIDEditor

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+	
+	private Button okBtn;
+	private Button cancelBtn;
+	private Textbox parentTextBox;
+	private WSearchEditor recordsEditor;
+	
+	private Integer value;
```
hengsin, line 79: you are really managing 2 instead of 1 value here. more precise to have tableIDValue and recordIDValue instead of just value.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,196 @@
...
+			}
+			else if(event.getTarget().equals(cancelBtn)) {
+				onClose();
+			}
+		}
+		if(event.getName().equalsIgnoreCase(Events.ON_SELECT)) {
+			if (event.getTarget().equals(tableIDEditorCopy.getComponent())) {
+				recordsEditor.getComponent().detach();
+				MLookup recordsLookup = getRecordsLookup();
+		    	recordsEditor = recordsLookup != null ? new WSearchEditor("Record_ID", false, false, true, recordsLookup) : new WSearchEditor(editor.getGridField());
```
hengsin, line 175: what's the use case of WSearchEditor with no lookup ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			tableIDGridField = gridTab.getField("AD_Table_ID");
+			tableIDEditor = WebEditorFactory.getEditor(tableIDGridField, true);
+			
+			if(tableIDGridField.getValue() != null)
+				tableIDEditor.setValue(tableIDGridField.getValue());
+						
+			dataBinder = new GridTabDataBinder(gridTab);
+			tableIDEditor.addValueChangeListener(dataBinder);
+			
+			tableIDGridField.addPropertyChangeListener(tableIDEditor);
```
hengsin, line 108: shouldn't need tableIDEditor here, that should be used in WRecordIDDialog only.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
hengsin, line 259: should just update oldTableID from WRecordIDDialog, i.e call editor.setAD_Table_ID(..) in WRecordIDDialog

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+				Integer recordID = Integer.parseInt(String.valueOf(value));
+				Integer tableID = Integer.parseInt(String.valueOf(oldTableID));
+				if(recordID != null && tableID != null)
+					recordTextBox.setValue(getIdentifierColumns(tableID, recordID));
+			}
+		}
+
+//		Data Binding
+		if (fire) {
+			ValueChangeEvent changeEvent = new ValueChangeEvent(this, this.getColumnName(), oldValue, value);
```
hengsin, line 270: there should be a need to fire ValueChangeEvent for AD_Table_ID before this, either over here or at setAD_Table_ID (call from WRecordIDDialog)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		else
+			return "";
+	}
+	
+	@Override
+    public String getDisplayTextForGridView(Object value) {
+		String s = "";
+		if(value != null &&  value.equals(gridField.getValue())) {
+			this.setValue(value);
+			s = getDisplay();
```
hengsin, line 295: like I've posted above, getDisplay() wouldn't works for gridview.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		if(((Integer) tableIDGridField.getValue() != null) && ((Integer) gridField.getValue() != null))
+			return getIdentifierColumns((Integer) tableIDGridField.getValue(), (Integer) gridField.getValue());
+		else
+			return "";
+	}
+	
+	@Override
+    public String getDisplayTextForGridView(Object value) {
+		String s = "";
+		if(value != null &&  value.equals(gridField.getValue())) {
```
hengsin, line 293: != null is fine but not `value.equals(gridField.getValue())` check.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			this.setValue(value);
```
hengsin, line 294: this is completely wrong, you must not call setValue in getDisplayTextForGridView

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+			s = getDisplay();
+			if ("<0>".equals(s)) {
+			s = "";
+			}
+		}
+		else {
+			this.setValue(value);
+			s = "<"+value+">";
```
hengsin, line 302: again, must not call setvalue. if it is null, should just show empty string.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,367 @@
...
+		}
+	}
+
+	/**
+	 * Get Identifier String from AD_Table_ID and Record_ID
+	 * @param tableID
+	 * @param recordID
+	 * @return String
+	 */
+	public static String getIdentifierColumns(int tableID, int recordID) {
```
hengsin, line 330: if you can get the lookup (like in WRecordIDDialog), it is better to use the display text from lookup instead. Reason being that would make it consistent with other part of the application.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WSearchEditor.java
```diff
@@ -261,7 +262,7 @@ public void setValue(Object value)
 	{
 		Object curValue = this.value;
         this.value = value;
-		if (value != null && !"".equals(String.valueOf(value)))
+		if (value != null && !"".equals(String.valueOf(value)) && lookup != null)
```
hengsin, line 265: I think both changes here is not needed for WSearchEditor. this editor is not going to work without a lookup.

CarlosRuiz-globalqss:
> Is this editor means to edit two field (AD_Table_ID and Record_ID) and the field AD_Table_ID means to be make invisible ?

AD_Table_ID is still needed because the Record_ID just shows the identifier of the record, not to which table it refers.  Also is important to keep AD_Table_ID for search and ordering purposes.


> Carlos have been using the Test window to test this, I think we should add that to this PR to illustrate the intended usage of this editor.

Sounds good.
Also, I've always wondered why Test window has just System access configured, when half of the fields cannot be tested in System but in Tenants.  I think we better move that window to All access - do you agree?

The other thing that I'm thinking is that this editor is a full replacement of the button Record_ID, but with big improvements, so, wondering if we must change all the actual Record_ID to this new editor.
This would be very valuable for example in AD_Scheduler.Record_ID which at this moment needs to be filled with a number (prone to errors).

If we go ahead we have three scenarios:

1 - all the Record_ID buttons converted to this editor and made them read-only
that can be done with these updates:
UPDATE ad_column SET ad_reference_id = 200202
WHERE  columnname='Record_ID' AND ad_reference_id = 28 AND ad_column_id < 1000000;
UPDATE ad_field SET xposition=xposition+1, isreadonly='Y'
WHERE ad_field_id < 10000000 AND ad_column_id IN (SELECT ad_column_id FROM ad_column WHERE  columnname='Record_ID' AND ad_reference_id = 200202);

2 - the AD_Scheduler.Record_ID would be the only one editable

3 - We have a Record_ID in AD_PInstance which is not usable here because there is no AD_Table_ID, so options are we keep it as Integer, or somehow we add AD_Table_ID to AD_PInstance, but that cannot be filled backwards, we would need to change where AD_PInstance is created to fill AD_Table_ID, and we would need to check how this editor behaves in those records that didn't have AD_Table_ID filled.

CarlosRuiz-globalqss:
@PeterTakacs300 - can you please integrate the following patch:
[pr1610_1.patch.txt](https://github.com/idempiere/idempiere/files/10757504/pr1610_1.patch.txt)

It implement the changes agreed with @hengsin above:
- change Test window to All access level
- add Table+RecordID to Test window - this will help for our tests here
- change all record_id buttons in official dictionary to use this new editor and make these fields read-only
- change AD_Scheduler.Record_ID to use the new editor as editable - and mark this field as advanced
- fix a problem found when we replace AD_PInstance.Record_ID with this editor and table is empty (which will be done in future)

> * For the AD_Table_ID field, I think it should be readonly when recordid editor is use.

About this, I agree with making the AD_Table_ID read-only when using this editor, but I think we cannot leave that decision to the implementor, otherwise data corruption could happen:
- can we force table to be read-only using this editor?
- if not, then maybe we need to implement it in MField.beforeSave - to force the AD_Table_ID field to be read-only when the accompanying Record_ID is using this editor and is not read-only


> * AD_PInstance.Record_ID - that's simply a design error that we never fix. How would Record_ID work without AD_Table_ID ? It is not possible.

I think we better implement that in a different ticket (to avoid polluting this one):
[IDEMPIERE-5580](https://idempiere.atlassian.net/browse/IDEMPIERE-5580)

CarlosRuiz-globalqss:
BTW - I found another bug, when selecting table A_Asset_Change the editor throws a Null Pointer Exception

PeterTakacs300:
Thank you for the tests and review. The latest commits should contain the requested changes by @hengsin and the patch by @CarlosRuiz-globalqss 

The current state can be labelled as work in progress - next I will implement the 'forcing AD_Table_ID field to readonly' feature. Besides that, thanks to your review the code structure is becoming cleaner, so there is a need for a 'code refactor' and also 'bug fixes' commit

PeterTakacs300:
The code is now ready to review. 

In the first commit contains bug fixes for the bugs previously mentioned on this PR.

In the second commit I implemented the forcing of the AD_Table_ID field to be read-only. I could not find a way to force it from the WRecordIDEditor.

hengsin:
please don't test or review this, big patch incoming ...

hengsin:
hi @CarlosRuiz-globalqss , this PR has migration script in both ID10 and ID11, which folder shall it goes to ?

PeterTakacs300:
Hi @hengsin, thank you for the patch. 

Besides that, I've found a NullPointerException (it was there before the patch too). Steps to reproduce:
- open the Record ID Dialog
- select Table ASP_ClientException
- type something in the Record field or open the info window through the three dots
- an error dialog opens with message: Error: No query columns found and also a NullPointerException is thrown

What do you think is the best way to avoid this?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
```diff
@@ -0,0 +1,356 @@
...
+		
+		getComponent().setSclass("recordid-editor");
+		getComponent().addEventListener(Events.ON_RIGHT_CLICK, this);
+		
+		init();
+	}
+
+	private void init() {
+		if(gridTab == null) {
+			throw new IllegalArgumentException("gridTab is null");
```
PeterTakacs300, line 101: gridTab is null when we call this from FindWindow, so now the find window cannot be opened on windows with the RecordID editor because this exception is thrown. If you suggest to have this exception here, we should handle the gridTab == null situation somewhere. Maybe ignore this editor in FindWindow.java by adding a check to addSelectionColumn(GridField mField, Group group) method?

```
if(DisplayType.RecordID == mField.getDisplayType())
      	return false;
```

hengsin: does it works on the find window if we remove the gridtab exception ?

PeterTakacs300: yes, it works if we'd add a simple `return;` statement instead of the exception

hengsin: alright, if it is working fine, you can roll it back to the simple return.

hengsin: hmm .. just tested and it seems you can't use the record id editor at find window. what do you means by it is working ?

PeterTakacs300: Oh sorry I misunderstood you, taught we are talking about skipping the editor in find window... I can get it work by sending the GridTab from Find Window. Providing the fix...

hengsin:
> Hi @hengsin, thank you for the patch.
> 
> Besides that, I've found a NullPointerException (it was there before the patch too). Steps to reproduce:
> 
>     * open the Record ID Dialog
> 
>     * select Table ASP_ClientException
> 
>     * type something in the Record field or open the info window through the three dots
> 
>     * an error dialog opens with message: Error: No query columns found and also a NullPointerException is thrown
> 
> 
> What do you think is the best way to avoid this?

[infopanel_npe.patch.txt](https://github.com/idempiere/idempiere/files/10802067/infopanel_npe.patch.txt)
- FIX NPE.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WRecordIDEditor.java
PeterTakacs300: So after further analysis, it looks like Find Window doesn't have its own GridTab, so I can't access the AD_Table_ID field of the Find Window. If I'm correct, through the GridTab in FindWindow I can only access the AD_Table_ID of the Window itself (that would cause that I can change AD_Table_ID of the Window from FindWindow)

hengsin:
[findwindow.patch.txt](https://github.com/idempiere/idempiere/files/10809981/findwindow.patch.txt)
- Fix support for find window 
- Remove mandatory record_id check

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordIDDialog.java
```diff
@@ -0,0 +1,220 @@
...
+				onClose();
+			}
+		}
+		if(event.getName().equalsIgnoreCase(Events.ON_SELECT)) {
+			if (event.getTarget().equals(tableIDEditor.getComponent())) {
+				// the Record_ID should be cleared when a different AD_Table_ID is selected
+				if (recordsEditor != null) {
+					recordsEditorLabel.detach();
+					recordsEditor.getComponent().detach();
+					editor.setValue(null);
```
hengsin, line 206: this is not right, should only happens when user press the apply button.

PeterTakacs300: ok thanks, then probably best is to clear recordsEditor here, since that value is what editor.setValue(..); uses on OK clicked

hengsin (approved):
tested

PR #1632, opened 2023-01-10
IDEMPIERE-5538 - Add isValid to Drill Rule allow validate parameters with beforesave

PeterTakacs300:
JIRA Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5538

# Pull Request Checklist


org.adempiere.base/src/org/compiere/model/MProcessDrillRule.java
```diff
@@ -211,4 +221,40 @@ public PO markImmutable() {
...
+		return true;
+	}
+	
+	/**
+	 * Validate Drill Rule - update IsValid
+	 */
+	public void validate() {
+		String sql = " UPDATE " + MProcessDrillRule.Table_Name + " SET " + MProcessDrillRule.COLUMNNAME_IsValid + " = ? "
+				+ " WHERE " + MProcessDrillRule.Table_Name + "_ID = ?";
+		DB.executeUpdateEx(sql, new Object[] {allMandatoryParaSet() ? "Y" : "N", getAD_Process_DrillRule_ID()}, get_TrxName());
```
hengsin, line 258: It seems strange to use afterSave and executeUpdateEx here. Should be much simpler to just do it in beforeSave and set IsValid to true of false.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MProcessDrillRulePara.java
```diff
@@ -95,4 +97,36 @@ public PO markImmutable() {
...
+			MProcessPara processPara = new Query(getCtx(), MProcessPara.Table_Name, whereClause, get_TrxName()).setParameters(getAD_Process_Para_ID()).first();
+			if(processPara.isMandatory()) {
+				validateParent(new Object[] {"N", getAD_Process_DrillRule_ID()});
+			}
+		}
+		return super.afterDelete(success);
+	}
+	
+	private void validateParent(Object[] para) {
+			String sql = " UPDATE " + MProcessDrillRule.Table_Name + " SET " + MProcessDrillRule.COLUMNNAME_IsValid + " = ? "
```
CarlosRuiz-globalqss, line 128: here too - this is better done using the model class instead of direct SQL

org.adempiere.base/src/org/compiere/model/MProcessDrillRule.java
PeterTakacs300: ok, I've rewritten and moved it to beforeSave

org.adempiere.base/src/org/compiere/model/MProcessDrillRulePara.java
PeterTakacs300: hi @CarlosRuiz-globalqss , using the model class in afterDelete causes the following issue when deleting Drill Rule Para:

Screenshot 2023-01-16 at 13 02 28

Since we don't want to validate and update the parent if the deletion was not successful, the validation probably needs to be in AfterDelete. Is there a solution to avoid getting this warning message?

norbertbede: From ux perspective this is a phantom error message. If no other solution then probably  would be better validate it only by process not after save, after delete. wdyt ?

CarlosRuiz-globalqss: Thanks @norbertbede - I understand your concern about ux perspective.

From technical perspective, it seems bad to use direct SQL instead of model to overcome a ux problem, in such case is better to solve the ux problem.

You would need to check where is the message triggered and check for some condition to avoid it when is controlled (like in the case of payment term and here)

PeterTakacs300: Hi @CarlosRuiz-globalqss ,

With @norbertbede we decided to remove the validation from Drill Rule and Drill Rule Para beforeSave/afterSave/afterDelete, so the only way to validate the Drill Rule will be through the new process for now.

The issue regarding to the warning message could go probably for another ticket.

If you agree with this, we would update the ticket description too and it could be closed sooner.

CarlosRuiz-globalqss: > With @norbertbede we decided to remove the validation from Drill Rule and Drill Rule Para beforeSave/afterSave/afterDelete, so the only way to validate the Drill Rule will be through the new process for now.
> 
> The issue regarding to the warning message could go probably for another ticket.
> 
> If you agree with this, we would update the ticket description too and it could be closed sooner.

I did the following test:

- Create _Drill Rule_ with _Process_ = Accounting Fact Daily_RV_Fact_Acct_Day
- _Capture Process Parameters_ - define _Accounting Schema_ which is mandatory
- Run _Validate_ process
- The rule is set as Valid because the mandatory parameter is set
- Delete the _Accounting Schema_ record
- The rule is still valid, but it will fail if used - which is what you're trying to avoid with this ticket

So, which one is worst for user: a harmless phantom error message?  or allowing to use an invalid rule?

org.adempiere.base/src/org/compiere/model/MProcessDrillRule.java
```diff
@@ -154,7 +155,7 @@ public static MProcessDrillRule[] getByTable(Properties ctx, int AD_Table_ID, St
 	 */
 	public static MProcessDrillRule[] getByColumnName(Properties ctx, String columnName, String trxName) {
 
-		String whereClause = "";
+		String whereClause = " IsValid = 'Y' ";
```
CarlosRuiz-globalqss, line 158: I see the getByTable is not used anywhere, but in case is used in future, must not have the same where filter for valid rules?

PeterTakacs300: Yes, thank you, it is fixed

org.adempiere.base/src/org/compiere/model/MProcessDrillRulePara.java
PeterTakacs300: Hi @CarlosRuiz-globalqss, I've tested similar use case in Info Column - when I delete an Info Column it also calls the validation method of the parent, which sets the IsValid column, then saves the parent record. In this case however there is no phantom error.

I could not find how is it solved there, or what is the difference between the two implementations. While debugging, I've found that in GridTable:3991, method hasChanged(PO po) it checks the current value of "Updated" with the old value of "Updated". When I'm on the info column, they are the same, but when I'm on drill rule para, they are different. Seems like the Updated column of the info window is not changed when I set IsValid after deleting an info column? I could not get further than this.

PeterTakacs300:
Hi @CarlosRuiz-globalqss, I've added back the validation to the beforeSave/afterDelete methods as this type of validation is throwing that phantom message at Info Column deletion too.

In addition, we have added an improvement, which allows the user to use System Drill Rules in Tenants correctly. The issue there was that if you have a Drill Rule defined in System and have mandatory parameter e.g. Warehouse, you can not define the default value of that in System through the Capture Parameters process - this way in tenant you can not use the rule, it will throw fill mandatory error.
Our solution is to extend the validation of the system drill rules:

not valid if it has any mandatory parameters and Show Help is not set to "Show Help"
if it is set to "Show Help" then the rule is valid and the user in tenant can set the mandatory parameters when starting the report from the drill assistant

PeterTakacs300:
Adding changes requested here: https://idempiere.atlassian.net/browse/IDEMPIERE-5302?focusedCommentId=49082
Also updating the ticket description.

CarlosRuiz-globalqss (changes requested):
Hi @PeterTakacs300 - this looks like a bug fix for IDEMPIERE-5302, please move the scripts to iD10 instead as the feature was included in release-10

About your new use case it failed in my tests:
- created a System Drill Rule for Aging report - no parameters
- log into GardenWorld
- tried to create a new Drill Rule Parameter (both manually or using the Capture form) and failed with "Error: Cross tenant PO writing request detected from session 1000107 for table AD_Process_DrillRule Record_ID=1000000"

PeterTakacs300:
Hi @CarlosRuiz-globalqss 

Sorry I didn't describe is clearly. The new use case is only about the validation of the System Drill Rules. You can not modify a system drill rule from tenant. 

The use case is about that if in system you set Show Help on the drill rule, it will be considered as 'valid' even if there are no Drill Rule parameters (if there are mandatory parameters, the user is able to set them in the process dialog when running the report thanks to setting show help)

To avoid the cross tenant error, I've added display logic to the process toolbar button - it will be hidden for drill rules of different tenant (same as the 'validate' process).

CarlosRuiz-globalqss:
Thanks @PeterTakacs300 - I just tested and I can manually add GardenWorld parameters in a System drill rule - is that OK - what is the effect of that?

PeterTakacs300:
That also shouldn't happen, there was a missing read only logic - fixed

CarlosRuiz-globalqss:
Hi Peter, please move all the scripts to iD10

CarlosRuiz-globalqss:
Tested successfully - just pending the move of the scripts to folder iD10

PeterTakacs300:
ok done - sorry I wasn't sure if you consider the whole PR as a bug fix

CarlosRuiz-globalqss (approved):
Hi @hengsin - for me this is ready - basic tests passed.

Is it OK for you?

hengsin (approved):
looks good and tested by Carlos

PR #1640, opened 2023-01-16
IDEMPIERE-5548 Issues with Log Management

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5548

- removed the button Reload Log Props from the About Dialog as is now being reloaded when changing level


hengsin (approved):
looks good

PR #1649, opened 2023-01-26
IDEMPIERE-5302 Drill Assistant Feature

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5302


CarlosRuiz-globalqss:
forcing push because is security issue

PR #1656, opened 2023-01-28
IDEMPIERE-5528 Context parsing on process Parameter not considering tab no

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5528

- Solve NPE


hengsin (approved):
trivial fix, good to go

PR #1664, opened 2023-02-07
IDEMPIERE-5443 Date Range Filter Component / SystemIDs

CarlosRuiz-globalqss:
- minor change - reorganize SystemIDs


hengsin (approved):
trivial fix, good to go

PR #1673, opened 2023-02-13
IDEMPIERE-5574 Drop Ship BPartner must have opposite IsSOTrx logic on Info Window

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5574

## Checklist:

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultInfoFactory.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/factory/DefaultInfoFactory.java
@@ -34,6 +34,8 @@ import org.adempiere.webui.panel.InfoPanel;
 import org.adempiere.webui.panel.InfoPaymentPanel;
 import org.adempiere.webui.panel.InfoProductPanel;
 import org.compiere.model.GridField;
+import org.compiere.model.I_C_Order;
+import org.compiere.model.I_M_InOut;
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - could you please peer review this approach.

I know it sounds hackish, but it works fine with a very little cost.

I was analyzing how to implement a "proper" solution, and we would need to know the name of the field that is triggering the info window, I imagine putting a context variable for that purpose and changing the default/display logic of the IsCustomer and IsVendor fields to behave different according to the column DropShip_BPartner_ID.
The way how Info Window is called at this moment is not passing the information of the caller field, so the change would become too big and I'm afraid potentially disruptive because it could have collateral effects.

hengsin:
hi @CarlosRuiz-globalqss ,
- looks fine but will cause conflict with the active/inactive pr.
- after we incorporate the active/inactive pr, caller field is pass to info window.

CarlosRuiz-globalqss:
OK, thanks for the review.

Let's wait then until IDEMPIERE-5573 is pushed and I'll update this patch here.

This is the modified patch:
[5574_over_5573.patch.txt](https://github.com/idempiere/idempiere/files/10744278/5574_over_5573.patch.txt)

I added some more logic to avoid working on other tables that maybe are not intended to behave that way.

CarlosRuiz-globalqss:
Hi @hengsin - this is ready for review - merged this pull request with latest changes

PR #1678, opened 2023-02-16
IDEMPIERE-5576 - Disallow "period close and permanent close" when unposted documents exists

PeterTakacs300:
JIRA Ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5576

# Pull Request Checklist


org.adempiere.base/src/org/compiere/model/MPeriod.java
```diff
@@ -1000,4 +1001,53 @@ public MPeriod markImmutable()
...
 
+    public boolean hasUnpostedDocs() {
+    	return hasUnpostedDocs(0);
+    }
+    
+    public boolean hasUnpostedDocs(int periodControlID) {
+    	
+    	String sql = " SELECT 1 "
+    			+ " FROM RV_UnPosted up "
+    			+ " WHERE up.DocStatus IN('CO', 'CL') AND up.AD_Client_ID = ? AND up.DateAcct BETWEEN ? AND ? ";
```
CarlosRuiz-globalqss, line 1012: several comments about this:

1 - no javadoc for new methods

2 - better to regenerate serialversionid when adding public methods 

3 - I think reversed documents also do posting, not just CO/CL

4 - I haven't checked or tested, but wondering if you took into account multi-accounting schema, what happens if the document has posting in one schema but not in a second schema?

5 - Again, I haven't checked or tested, but I'm not sure that adding AD_Org_ID filter to period is how this works.  From what I see in code you need to look is for the calendar of that organization.  In other words, an organization can have a calendar, and still their periods be in Org *  (haven't tested)
Also, when the organization has no calendar, it falls back to the tenant calendar.

CarlosRuiz-globalqss (changes requested):

org.adempiere.base/src/org/compiere/model/MPeriod.java
CarlosRuiz-globalqss: Another thing about this SELECT, as you just need to get one record, I think @norbertbede mentioned that adding a "FETCH FIRST 1 ROWS ONLY" clause makes the DB optimizer works better.

And I think you can do the same calling DB.getSQLValueEx which is more optimal code.

norbertbede: point 3. we should remove this validation. if eg. user leave status Draft, then must be processed or change dateacct period to next open period. for further processing (that is what Steven suggested me in email.)

point 4. 
document has posted boolean, that assume all schema must be accounted properly. i would suggest we can close this, then i will contact community member from Taiwan to check in their use case, they are using as i know many schemas.

PeterTakacs300: I've committed the rest of the changes + a new behaviour: if you manually select more Period Controls in grid mode and try to close them, the process will close those with no unprocessed docs, and return a list of the skipped ones.

I could not implement this logic in the Open/Close All process on Perion tab, because for that I would probably need to change the hasUnprocessedDocs() method to return a list of C_PeriodControl_IDs

CarlosRuiz-globalqss: > point 3. we should remove this validation. if eg. user leave status Draft, then must be processed or change dateacct period to next open period. for further processing (that is what Steven suggested me in email.)

@norbertbede - I disagree about this, that's why that was [my first question on the ticket](https://idempiere.atlassian.net/browse/IDEMPIERE-5576?focusedCommentId=49140).


The situation is like this:

- the accountant wants to close January and there are 120 draft documents entered by 40 employees
- so we're forcing the accountant to contact 40 employees to move the dates of their documents, so he can do his work?

IMO the accountant must be able to close the period and every employee will face the "Period Closed" error when they try to complete the document and they'll fix the situation when needed.

hengsin: I guess the proactive way is to auto notify the employees about those draft and invalid status documents. anyway, I'm more towards Carlos opinion above.

PeterTakacs300: In the latest commit I've added back the DocStatus check

PeterTakacs300: Hi @CarlosRuiz-globalqss, is this PR in state "waiting for review" or did I miss any of the requested changes?

CarlosRuiz-globalqss:
@PeterTakacs300, please apply the following patch:
[pr1678PR1.patch.txt](https://github.com/idempiere/idempiere/files/10903023/pr1678PR1.patch.txt)

It resolves multiple issues:
- change SysConfig to Tenant level
- solve NPE when running from the Info Window Period Control Management
- solve oracle migration script too long line
- improve the logic to avoid calling the time consuming method hasUnposted when not necessary
- fix the logic for the organization calendars

PeterTakacs300:
Thank you for the fixes

CarlosRuiz-globalqss (approved):
tested

PR #1703, opened 2023-03-08
IDEMPIERE-5238 Record_ID editor

CarlosRuiz-globalqss:
- exclude AD_ChangeLog and AD_RecentItem from the validation to avoid affecting performance

Resolving @hengsin 's comment at
https://idempiere.atlassian.net/browse/IDEMPIERE-5238?focusedCommentId=49202

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -5520,11 +5520,15 @@ public abstract class PO
 	 * @return true if all the foreign keys are valid
 	 */
 	private void checkRecordIDCrossTenant() {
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
Hi @hengsin -

I did some tests and I think is better to exclude the validation from AD_ChangeLog and AD_RecentItem as those are high-usage tables and there is no way to modify that from the UI.  It would be still open to insert bad records using webservices, but the information would not be visible as the system throws a Cross Tenant error when visiting such record.

I also changed the order of checking for the columns, first Record_ID as it happens less frequently, must improve a bit performance.

hengsin (approved):
looks good

PR #1706, opened 2023-03-08
IDEMPIERE-5616 Reverse Shipments don't post because of Not Balanced in special case

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5616

## Checklist:

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Doc_InOut.java
+++ b/org.adempiere.base/src/org/compiere/acct/Doc_InOut.java
@@ -288,7 +288,7 @@ public class Doc_InOut extends Doc
 				{
 					//	Set AmtAcctDr from Original Shipment/Receipt
 					if (!dr.updateReverseLine (MInOut.Table_ID,
-							m_Reversal_ID, line.getReversalLine_ID(),Env.ONE, true))
+							m_Reversal_ID, line.getReversalLine_ID(),Env.ONE))
 					{
 						if (! product.isStocked())	{ //	ignore service
 							fact.remove(dr);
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/acct/Doc_InOut.java
```diff
@@ -288,7 +288,7 @@ public ArrayList<Fact> createFacts (MAcctSchema as)
 				{
 					//	Set AmtAcctDr from Original Shipment/Receipt
 					if (!dr.updateReverseLine (MInOut.Table_ID,
-							m_Reversal_ID, line.getReversalLine_ID(),Env.ONE))
+							m_Reversal_ID, line.getReversalLine_ID(),Env.ONE, true))
```
hengsin, line 291: perhaps it is safer and easier to understand to say we are looking for "DR" or "CR" line instead of "First" and "Last" line ?

CarlosRuiz-globalqss: Hi @hengsin

I thought about that but it won't work well for the normal case.

If I change the discovery to debit and credit then it can affect the normal case - which expects just one line per account - but I can't be 100% sure if the exact account line being reversed is credit or debit because it can be both

For example: it is expected that asset in shipment is always credit, but if somebody entered a negative shipment then it would be debit.

The last/first approach is just safe for both cases, as 99.9999% of the cases there will be just one line - this is more like solving a corner case that was probably misconfigured and it cannot be fixed in normal ways.  (I could reconfigure the accounts to be different, but that would imply reopening historic periods that are permanently closed and change old balances, etc).

Regards,

Carlos Ruiz

hengsin:
[pr1706.patch.txt](https://github.com/idempiere/idempiere/files/10967007/pr1706.patch.txt)
- the first/last approach doesn't pass unit test. it is not working when there's currency balancing line.
- I've uploaded a different approach, please verify whether that works for your test cases.

CarlosRuiz-globalqss:
Thanks @hengsin - tested your patch in our problematic case and it works fine - also passed all unit tests

hengsin (approved):
tested

PR #1715, opened 2023-03-10
IDEMPIERE-5608: Improve server logs by tenant and user data

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-5608

# Pull Request Checklist

### Tests
### Documentation


migration/iD10/oracle/202303101652_IDEMPIERE-5608.sql
```diff
@@ -0,0 +1,10 @@
+-- IDEMPIERE-5608
+SELECT register_migration_script('202303101652_IDEMPIERE-5608.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Mar 10, 2023, 4:52:29 PM CET
+INSERT INTO AD_SysConfig (AD_SysConfig_ID,AD_Client_ID,AD_Org_ID,Created,Updated,CreatedBy,UpdatedBy,IsActive,Name,Value,Description,EntityType,ConfigurationLevel,AD_SysConfig_UU) VALUES (200219,0,0,TO_TIMESTAMP('2023-03-10 16:52:28','YYYY-MM-DD HH24:MI:SS'),TO_TIMESTAMP('2023-03-10 16:52:28','YYYY-MM-DD HH24:MI:SS'),10,10,'Y','LOG_PREFIX','@#AD_Client_Name@ - @#AD_User_Name@ - @#AD_Session_ID@','Prefix for logs ; you can use any context variables such as @#AD_Client_Name@, @#AD_User_Name@, @#AD_Session_ID@','D','C','f7c613cf-6264-44b8-b283-2fb642a9a7b4')
```
hengsin, line 8: this is misleading as the prefix is only for logging to files.

migration/iD10/postgresql/202303101652_IDEMPIERE-5608.sql
```diff
@@ -0,0 +1,7 @@
+-- IDEMPIERE-5608
+SELECT register_migration_script('202303101652_IDEMPIERE-5608.sql') FROM dual;
+
+-- Mar 10, 2023, 4:52:29 PM CET
+INSERT INTO AD_SysConfig (AD_SysConfig_ID,AD_Client_ID,AD_Org_ID,Created,Updated,CreatedBy,UpdatedBy,IsActive,Name,Value,Description,EntityType,ConfigurationLevel,AD_SysConfig_UU) VALUES (200219,0,0,TO_TIMESTAMP('2023-03-10 16:52:28','YYYY-MM-DD HH24:MI:SS'),TO_TIMESTAMP('2023-03-10 16:52:28','YYYY-MM-DD HH24:MI:SS'),10,10,'Y','LOG_PREFIX','@#AD_Client_Name@ - @#AD_User_Name@ - @#AD_Session_ID@','Prefix for logs ; you can use any context variables such as @#AD_Client_Name@, @#AD_User_Name@, @#AD_Session_ID@','D','C','f7c613cf-6264-44b8-b283-2fb642a9a7b4')
+;
```
hengsin, line 6: Sceptical that this should be configurable at Client level since the intention is to be able to identified the source tenant of the message.

org.adempiere.base/src/org/compiere/util/CLogFile.java
```diff
@@ -334,6 +337,22 @@ public synchronized void publish (LogRecord record)
 		}
 	}	//	publish
 
+	private String getPrefix() {
+		try
+		{
+			if (DB.isConnected()) {
```
hengsin, line 343: need investigation whether this could cause infinite loop when db connection is not available for whatever reason.

hengsin: Note that we have to be very careful with the use of DB connection and MSysConfig inside Log handler since one of the potential exception that's to be logged is exception connection to DB.

CarlosRuiz-globalqss:
After reading @hengsin peer review, it sounds easier/safer to add this like a VM argument?

Something like
```-Dorg.idempiere.FileLogPrefix="@#AD_Client_Name@ - @#AD_User_Name@ - @#AD_Session_ID@"```

Would that make sense?

**Comment by hengsin** [comment]
+1 for Carlos suggestion.

**Comment by nmicoud** [comment]
ok for me too (ability to set client level was just a "in case of") ; but I don't know how to retrieve such parameter.
I'll try to work on that by the end of this week

**Comment by CarlosRuiz-globalqss** [comment]
@nmicoud - would be simply to replace the line:

String prefix = MSysConfig.getValue(LOG_PREFIX, Env.getAD_Client_ID(Env.getCtx()));
with
String prefix = System.getProperty("org.idempiere.FileLogPrefix");

**Comment by nmicoud** [comment]
SysConfig replaced by VM argument

File: org.adempiere.base/src/org/compiere/util/CLogFile.java [hunk 4]
```diff
@@ -334,6 +333,19 @@ public synchronized void publish (LogRecord record)
 		}
 	}	//	publish
 
+	private String getPrefix()
+	{
+		try
+		{
+			return Env.parseContext(Env.getCtx(), 0, System.getProperty("org.idempiere.FileLogPrefix"), false);
```
hengsin, line 340: System.getProperty("org.idempiere.FileLogPrefix") should be read once in a static variable, i.e
```
private final static s_fileLogPrefix = System.getProperty("org.idempiere.FileLogPrefix")
```

Also, if we want to have a default:
```
private static s_fileLogPrefix;
static {
     s_fileLogPrefix = System.getProperty("org.idempiere.FileLogPrefix");
     if (Util.isEmpty(s_fileLogPrefix, true) {
          s_fileLogPrefix = "@#AD_Client_Name@ - @#AD_User_Name@ - @#AD_Session_ID@";
     }
}
```

hengsin: And we shouldn't call Env.parseContext if s_fileLogPrefix is empty or null.

org.adempiere.base/src/org/compiere/util/CLogFile.java
```diff
@@ -334,6 +333,19 @@ public synchronized void publish (LogRecord record)
...
 
+	private String getPrefix()
+	{
+		try
+		{
+			return Env.parseContext(Env.getCtx(), 0, System.getProperty("org.idempiere.FileLogPrefix"), false);
+		}
+		catch (Exception ex)
+		{
+			reportError ("LogFilePrefix", ex, ErrorManager.FORMAT_FAILURE);
```
hengsin, line 344: perhaps we should consider to set s_fileLogPrefix to empty here to avoid repeating the parse exception for every logging call.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/util/CLogFile.java
CarlosRuiz-globalqss: @hengsin - making it static would mean it cannot be changed during server execution?

hengsin: yeah, if you want to have the flexibility to change it during server execution, better not to make it static.

nmicoud: maybe a dumb question: how to change the value as it is defined on a VM argument?

CarlosRuiz-globalqss: > maybe a dumb question: how to change the value as it is defined on a VM argument?

I do it frequently for logging SQL statements - even on production servers.

For example, to log SQL statements you can pass the VM argument:
```-Dorg.idempiere.db.debug=true```

Now, if you didn't define it when you started the server (the normal case in production), I open the OSGI console:
```telnet localhost 12612```

and within the OSGI console I execute:
```setprop org.idempiere.db.debug=true```

and voilà!  SQL statements start appearing in the log
after I found what I'm looking for I simply set it back to false.

Extra tip:
The amount of SQL statements is overwhelming in a production server (and even in local dev), so there is another parameter to filter what is logged based on a regular expression, for example:
```setprop org.idempiere.db.debug.filter='(?i)(?s:.)*\bAD_ChangeLog\b(?s:.)*'```
log statements that contain the ad_changelog table

nmicoud: Ah great! thanks for the tip
I only use the "org.idempiere.db.debug" locally.
Great to have be able to change that on live
Merci beaucoup :)

nmicoud:
Env.parseContext is not called if prefix is empty.
I also force the property to "" in case of exception.

CarlosRuiz-globalqss (changes requested):
Hi @nmicoud - must we assume you didn't test the pull request?   :-)     - I'm asking because the checklist is not filled

Also, can you please create a new feature page in the wiki?

I did a peer review of this and would suggest to apply the following patch:
[PR1715pr.patch.txt](https://github.com/idempiere/idempiere/files/11000819/PR1715pr.patch.txt)

I used this parameter to test it:
``` -Dorg.idempiere.FileLogPrefix="@#AD_Client_Name@ - @#AD_User_Name@ - @#AD_Session_ID@ - "```

The result of your patch looks like this:
```GardenWorld - SuperUser - 1000127 - 11:56:02.106===========> DB.saveError: SaveErrorNotUnique - ERROR: llave duplicada viola restricción de unicidad «c_bpartner_value»```

But I think that's not good for most of log analyzers out there, so my suggested patch change the format to this:
```12:24:23.417===========> GardenWorld - SuperUser - 1000129 - DB.saveError: SaveErrorNotUnique - ERROR: llave duplicada viola restricción de unicidad «c_bpartner_value»```

nmicoud:
Hi,
Patch tested & integrated ; checklist filled ; I'll work on the documentation page asap

CarlosRuiz-globalqss:
Hi @hengsin - I tested this successfully and is now documented, is it OK from your side?

hengsin (approved):
tested  by Carlos.

PR #1727, opened 2023-03-13

CarlosRuiz-globalqss:

# Pull Request Checklist

### Tests
### Documentation


org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
```diff
@@ -79,12 +80,15 @@ public ModelClassGenerator (int AD_Table_ID, String directory, String packageNam
 	{
 		this.packageName = packageName;
 
+		MTable table = MTable.get(AD_Table_ID);
+		boolean uuidTable = table.getKeyColumns().length != 1;
```
hengsin, line 84: don't quite understand this one, isn't uuidTable is table with 1 UUID keycolumn ?

org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -2425,9 +2425,22 @@ else if (value instanceof Timestamp) {
 					multiRowWHERE.append (tableName).append(".").append(columnName)
 						.append ("=").append (DB.TO_STRING(value.toString()));
 			}
+			if (columnName.equals(PO.getUUIDColumnName(tableName)))
```
hengsin, line 2428: so the design here is the table with UUID key column wouldn't mark that UUID column as IsKey=Y ?

CarlosRuiz-globalqss: Right - IsKey would be reserved just for ID keys - I changed the DisplayLogic of that field to be shown just when the reference=ID (and when is already set for some exception cases that are already present in core)

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
CarlosRuiz-globalqss: The model class generator works generating a lot of methods related to the ID key column.

Those methods are basically worthless in the case of multi-parent-key tables, so, my idea here is that for all tables managing multi-parent key the ModelClassGenerator consider this table as a UUID based table.

This makes two changes in the generated class:
1 - in the toString instead of calling get_ID it calls get_UUID
2 - instead of generating a method getKeyNamePair (ID+Name), it generates a method getValueNamePair (UUID+Name)

And that's all, I tried to make it different, for example not generating the ID constructors and instead generating just the UUID constructors, but that breaks a lot of M classes.

Also, I found an issue with the class MRefTable (and probably there are other cases) where the ID is AD_Reference, but is not defined as the key, but as the parent key.  That forced me to keep the X_ class as it was before (just added the UUID constructors).

CarlosRuiz-globalqss: BTW - do you think is OK to generate all classes after we move these changes on Model Generator ahead?

CarlosRuiz-globalqss: @hengsin - I changed the name of the variable to be more accurate about its purpose.

I generated model classes for a new test table, and also generated model classes for Test table to illustrate how would be the changes on all the model classes.

org.adempiere.base/src/org/compiere/model/GridTable.java
hengsin: ic, this part doesn't looks right, it shouldn't happens for every table with _ID and _UU column.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1419,6 +1484,16 @@ public boolean load (String trxName, String ... virtualColumns) {
 	 */
 	protected boolean load (String uuID, String trxName, String ... virtualColumns)
```
hengsin, line 1485: this is a mistake from earlier revision, needs to rename it to something else as it is getting dangerously close to `public boolean load (String trxName, String ... virtualColumns)`

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -153,15 +153,26 @@ public PO (Properties ctx)
...
 	 */
 	public PO (Properties ctx, int ID, String trxName)
 	{
 		this (ctx, ID, trxName, null, (String[]) null);
 	}   //  PO
 
+	/**
+	 *  Create and Load existing Persistent Object
+	 *  @param ctx context
+	 *  @param UUID The unique UUID of the object
```
hengsin, line 168: Needs to document what's the value for new record ("" or a public constant like UUID_NEW_RECORD)

org.adempiere.base/src/org/compiere/model/PO.java
hengsin: Perhaps rename it to loadByUUOrID. Needs to fix the javadoc too to reflect that if uuID is null, it will reload using current IDs values.

hengsin: We need t fix `public boolean load (String trxName, String ... virtualColumns)` too. This is commonly use to reload PO during or after trx and with the new design change, it must works with table that only has _UU key too.

hengsin: For convenience, probably should add a new method to mtable, for e.g IsUUIDTable() to easily identify table that don't have ID key and is using UU as key.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -1884,31 +1959,55 @@ private void setKeyInfo()
...
 				}
-				catch (Exception e)
+				else
+					m_IDs[i] = get_Value(m_KeyColumns[i]);
+				if (log.isLoggable(Level.FINEST)) log.finest("(FK) " + m_KeyColumns[i] + "=" + m_IDs[i]);
+			}
+		}
+
+		if (m_KeyColumns == null || m_KeyColumns.length == 0)
+		{
```
hengsin, line 1990: This is inconsistent with GridTable.getWhereClause. At GridTable.getWhereClause _UU take precedence over isColumnParent (i.e multi key) and it is the other way round here.

org.adempiere.base/src/org/compiere/model/MZoomCondition.java
```diff
@@ -277,6 +277,29 @@ public static int findZoomWindowByTableId(int AD_Table_ID, int recordID)
...
+	}
+
+	/**
+	 * find AD_Window_ID from matching zoom condition record
+	 * @param AD_Table_ID
+	 * @param recordID
+	 * @param recordUU
+	 * @return AD_Window_ID
+	 */
+	public static int findZoomWindowByTableIdOrUU(int AD_Table_ID, int recordID, String recordUU, int windowNo)
```
hengsin, line 302: I think this should be private instead of public. Open to misuse since many tables have both ID and UU value.

org.adempiere.base/src/org/compiere/model/PO.java
hengsin: With the new setKeyInfo code, there's one troubling difference here - MTable.getKeyColumns and PO.get_KeyColumns could be different. Thus breaking a very well known API contract.

hengsin:
[pr1727.patch.txt](https://github.com/idempiere/idempiere/files/11007480/pr1727.patch.txt)
- for your consideration.

CarlosRuiz-globalqss:
> [pr1727.patch.txt](https://github.com/idempiere/idempiere/files/11007480/pr1727.patch.txt)
> 
>     * for your consideration.

Thanks, integrated ... testing

org.adempiere.base/src/org/compiere/model/PO.java
CarlosRuiz-globalqss: > Needs to document what's the value for new record ("" or a public constant like UUID_NEW_RECORD)

Ready, added PO.UUID_NEW_RECORD and changed test class accordingly

org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -2110,14 +2110,28 @@ public static int getZoomWindowID(MQuery query)
...
 		return getZoomWindowID(AD_Table_ID, Record_ID, 0);
 	}
 
 	public static int getZoomWindowID(int AD_Table_ID, int Record_ID, int windowNo)
 	{
-		int AD_Window_ID = MZoomCondition.findZoomWindowByTableId(AD_Table_ID, Record_ID, windowNo);
+		return getZoomWindowIDOrUU(AD_Table_ID, Record_ID, null, windowNo);
+	}
+
+	public static int getZoomWindowIDOrUU(int AD_Table_ID, int Record_ID, String Record_UU, int windowNo)
```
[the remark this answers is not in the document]
CarlosRuiz-globalqss: > I think this should be private instead of public. Open to misuse since many tables have both ID and UU value.

Ready, changed to private

org.adempiere.base/src/org/compiere/model/PO.java
CarlosRuiz-globalqss: > With the new setKeyInfo code, there's one significant side effect here - MTable.getKeyColumns and PO.get_KeyColumns could be different. Thus breaking a very well known API contract.

Readu, changed back GridTable to keep behavior same as PO

org.adempiere.base/src/org/compiere/model/MZoomCondition.java
CarlosRuiz-globalqss: > I think this should be private instead of public. Open to misuse since many tables have both ID and UU value.

Not possible, need to be called from Env class

migration/iD11/oracle/202303171334_IDEMPIERE-5567.sql
```diff
@@ -0,0 +1,250 @@
...
+UPDATE AD_Column SET IsAllowCopy='N', FKConstraintName='CreatedBy_TestUU', FKConstraintType='N',Updated=TO_TIMESTAMP('2023-03-17 13:35:40','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=215781
+;
+
+-- Mar 17, 2023, 1:35:40 PM CET
+UPDATE AD_Column SET IsAllowCopy='N', FKConstraintName='UpdatedBy_TestUU', FKConstraintType='N',Updated=TO_TIMESTAMP('2023-03-17 13:35:40','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=215783
+;
+
+-- Mar 17, 2023, 1:35:40 PM CET
+CREATE TABLE TestUU (AD_Client_ID NUMBER(10) NOT NULL, AD_Org_ID NUMBER(10) NOT NULL, Created DATE NOT NULL, CreatedBy NUMBER(10) NOT NULL, Description VARCHAR2(255 CHAR) DEFAULT NULL , Help VARCHAR2(2000 CHAR) DEFAULT NULL , IsActive CHAR(1) DEFAULT 'Y' CHECK (IsActive IN ('Y','N')) NOT NULL, Name VARCHAR2(60 CHAR) NOT NULL, TestUU_UU VARCHAR2(36 CHAR) DEFAULT NULL , Updated DATE NOT NULL, UpdatedBy NUMBER(10) NOT NULL, Value VARCHAR2(40 CHAR) DEFAULT NULL , CONSTRAINT TestUU_UU_idx UNIQUE (TestUU_UU))
+;
```
hengsin, line 89: UUID table should be created with _UU column as the primary key constraint.

hengsin (changes requested):

migration/iD11/oracle/202303171334_IDEMPIERE-5567.sql
CarlosRuiz-globalqss: > UUID table should be created with _UU column as the primary key constraint.

Ready @hengsin - implemented with commit: https://github.com/idempiere/idempiere/pull/1727/commits/16d531042dce3d350c8b74ea1ca0e86f9d76bd41

hengsin (approved):
looks good

PR #1741, opened 2023-03-20
IDEMPIERE-5567 Support of UUID as Foreign Key (FHCA-4195)

CarlosRuiz-globalqss:
# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MZoomCondition.java
+++ b/org.adempiere.base/src/org/compiere/model/MZoomCondition.java
@@ -255,12 +255,13 @@ public class MZoomCondition extends X_AD_ZoomCondition implements ImmutablePOSup
 							StringBuilder sql = new StringBuilder("SELECT ").append(parentTab.getLinkColumnName())
 									.append(" FROM ").append(parentTab.getTableName())
 									.append(" WHERE ");
-							if (parentTab.getLinkColumnName().endsWith("_UU")) {
+							MTable parentTable = MTable.get(Env.getCtx(), parentTab.getTableName());
+							if (parentTable.isUUIDKeyTable()) {
 								sql.append(PO.getUUIDColumnName(parentTab.getTableName())).append("=").append(DB.TO_STRING(parentId.toString()));
 							} else {
 								sql.append(parentTab.getTableName()).append("_ID=").append(parentId);
 							}
-							parentId = DB.getSQLValue(null, sql.toString());
+							parentId = null;
 							if (parentTab.getLinkColumnName().endsWith("_UU")) {
 								parentId = DB.getSQLValueString(null, sql.toString());
 							} else {
@@ -338,7 +339,7 @@ public class MZoomCondition extends X_AD_ZoomCondition implements ImmutablePOSup
 			};
 
 			String whereClause;
-			if (recordUU != null)
+			if (table.isUUIDKeyTable())
 				whereClause = PO.getUUIDColumnName(table.getTableName())+"="+DB.TO_STRING(recordUU);
 			else
 				whereClause = table.getTableName() + "_ID="+recordID;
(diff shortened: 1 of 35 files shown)


hengsin:
It might be a good idea to take a look at the possibility to migrate _UU column to native UID column in PostgreSQL and Raw(16) in Oracle before adding foreign key support. Migration will be a lot more difficult once we started using it as foreign key as well.

hengsin:
For PostgreSQL, to convert existing column from varchar to uuid type:
1. alter table test alter column test_uu drop default;
2. alter table test alter column test_uu type uuid using test_uu::uuid;

hengsin:
http://crafted-software.blogspot.com/2013/03/uuid-values-from-jdbc-to-postgres.html (looks like getString wouldn't work).

hengsin:
For Oracle:
ALTER TABLE test ADD test_uu_raw RAW(16);
UPDATE test SET test_uu_raw = hextoraw(replace(test_uu, '-', ''));
ALTER TABLE test DROP COLUMN test_uu;
ALTER TABLE test RENAME COLUMN test_uu_raw TO test_uu;

CarlosRuiz-globalqss:
> It might be a good idea to take a look at the possibility to migrate _UU column to native UID column in PostgreSQL and Raw(16) in Oracle before adding foreign key support. Migration will be a lot more difficult once we started using it as foreign key as well.

Hi @hengsin - I'm afraid that will take too much time and is out of scope of my current deliverable.

I don't think a future migration will be harder because we already as a String.

By the way, because is not a SQL standard and lack of compatibility between different database, I think we must consider to keep managing the object as String on java side, and just change the way it is written/read on the database side.  But again, that's something that I wouldn't want to tackle at this moment.

hengsin:
Migration is harder when foreign key constrain is in the picture.

One of the prime reason to use UUID key table is to store huge amount of data that's not possible with integer key. For that, the space and performance difference between storing UUID as native UID type (16 bytes) vs string (36 bytes) is pretty significant.

CarlosRuiz-globalqss:
> Migration is harder when you add foreign key constrain to the picture.
> 
> One of the prime reason to use UUID key table is to store huge amount of data that's not possible with integer key. For that, the space and performance difference between storing UUID as native UID type (16 bytes) vs string (36 bytes) is pretty significant.

I mean, migrating from varchar to uuid (postgresql) sounds kind of "easy" as postgresql supports the alter within transaction.

Migrating that in oracle cab be a bit more sql involved as it implies dropping the constraints first, then migrating the data and then recreating back the constraint, but I don't see that as harder, just as extra sql steps.

I understand the advantages of native data types vs varchar (storage and index access mostly), but I also see there is a big problem about being non-standard SQL and big differences about how oracle and postgresql manage it.
Is doable, but my main problem at this moment is about the time that will be required to implement and test that.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MZoomCondition.java
```diff
@@ -244,16 +252,33 @@ public static int findZoomWindowByWindowId(int AD_Window_ID, MQuery query, int w
...
+								sql.append(parentTab.getTableName()).append("_ID=").append(parentId);
+							}
+							parentId = DB.getSQLValue(null, sql.toString());
+							if (parentTab.getLinkColumnName().endsWith("_UU")) {
+								parentId = DB.getSQLValueString(null, sql.toString());
+							} else {
+								int tmpId = DB.getSQLValue(null, sql.toString());
+								if (tmpId > 0)
+									parentId = Integer.valueOf(tmpId);
+							}
```
hengsin, line 270: duplicate call to DB.getSQLValue ?

CarlosRuiz-globalqss: Thanks @hengsin - fixed all the concerns

hengsin (approved):
looks good

PR #1772, opened 2023-04-09

CarlosRuiz-globalqss:
New generated model classes with new UUID constructor

Generated code - no checklist


CarlosRuiz-globalqss:
Hi @hengsin - this pull request is basically the generated model classes, and I added on GardenWorld record in Test table, and added on MTestTest unit test to check the UUID constructor working with normal tables.

hengsin:
hi @CarlosRuiz-globalqss , I've a doubt here - it seems UUID constructor is generated for ID table too, is there benefit of doing so ? Also, why UUID constructor is being added to all X_* class, the same doesn't happens for M* class (for e.g MProduct and MBPartner), making that added constructor kind of pointless.

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , I've a doubt here - it seems UUID constructor is generated for ID table too, is there benefit of doing so ? Also, why UUID constructor is being added to all X_* class, the same doesn't happens for M* class (for e.g MProduct and MBPartner), making that added constructor kind of pointless.

Hi @hengsin - I find very handy to be able to use UUID constructor in all tables.

I added it to MTest class as an example.

Is your suggestion that we add it to all M classes in core?
I can add that to all M classes if that's your suggestion.

But now I have a doubt, must we change also the IModelFactory to reflect the new constructor?
Maybe with a default method that raises an Exception ("not implemented yet") for customizations.

CarlosRuiz-globalqss:
> yes, if we add it to X*, we should add it to M* as well, otherwise it is not usable. If time and resource permitted, we shall add that to IModelFactory too, looks possible to add the implementation to AbstractModelFactory. If you don't want to break backward binary compatibility, you can make that a default method that raise exception.

Ready @hengsin - I added the new UUID constructor to almost all M model classes.

There was a problem with MPaymentBatch that I documented here:
https://wiki.idempiere.org/en/Migration_Notes#Version_iDempiere_11

The FactLine class has a special constructor, so I didn't add UUID constructor to this class.

hengsin (approved):
looks good

PR #1780, opened 2023-04-13
IDEMPIERE-5668 - Improvements to the copy role process

joseleite550:
# Pull Request Checklist

### Tests
- [X] I checked all collaterals that can be affected by my changes, and tested other potential affected scenarios
- [X] New and existing unit tests pass locally with my changes
- [] I have added unit tests that prove my fix is effective or that my feature works
### Documentation
- [X] I have made corresponding changes to the documentation as follows:

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/CopyRole.java
+++ b/org.adempiere.base.process/src/org/compiere/process/CopyRole.java
@@ -27,6 +27,7 @@ import org.compiere.model.MProcessPara;
 import org.compiere.model.MRole;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
+import org.compiere.util.Msg;
 
 
 /**
@@ -43,7 +44,7 @@ public class CopyRole extends SvrProcess
 {
 	private int m_AD_Role_ID_From = 0;
 	private int m_AD_Role_ID_To = 0;
-	private int m_AD_Client_ID = 0;	
+	private Integer m_AD_Client_ID = null;	
 	private int m_AD_Org_ID = 0;
 	
 	/**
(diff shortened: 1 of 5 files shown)


org.adempiere.base.process/src/org/compiere/process/CopyRole.java
```diff
@@ -43,7 +44,7 @@ public class CopyRole extends SvrProcess
 {
 	private int m_AD_Role_ID_From = 0;
 	private int m_AD_Role_ID_To = 0;
-	private int m_AD_Client_ID = 0;	
+	private Integer m_AD_Client_ID = null;
```
hengsin, line 47: Since the process has no AD_Client_ID parameter, this should be removed.

hengsin: The name.equals("AD_Client_ID") part in prepare() should be removed too.

org.adempiere.base.process/src/org/compiere/process/CopyRole.java
```diff
@@ -78,12 +79,15 @@ else if (name.equals("AD_Org_ID"))
 	 */
 	protected String doIt() throws Exception
 	{	
+		if(m_AD_Client_ID == null)
+			m_AD_Client_ID = Env.getAD_Client_ID(getCtx());
```
hengsin, line 83: should change to int AD_Client_ID = Env.getAD_Client_ID(getCtx());

org.adempiere.base.process/src/org/compiere/process/CopyRole.java
joseleite550: This validation was done because when the modified AD_Client parameter is not filled in in the process, it takes the logged-in tenant and does not put zero in the AD_Client field

joseleite550: This modification from int to integer was made to validate whether the AD_Client parameter was filled in or not, as it is not mandatory, in case the AD_Client is not informed in the parameter we get  the AD_Client of the context

joseleite550: Changing it to integer we can assign null to the variable, enabling the validation if it was filled or not, since when assigned 0 when initializing it, we would not be sure if it was filled, since the dictionary has ID 0

hengsin: No, I means since the process actually doesn't have an "AD_Client_ID" parameter, the copy role process class shouldn't have the m_AD_Client_ID instance variable at all. You just always get it from the context through Env.getAD_Client_ID call.

joseleite550: Yes, I understand the case, but the copyRole process in the dictionary has the AD_client parameter, which was one of the parameters that I add dynamic validation

hengsin: ok, my mistake. checking with Carlos in ticket why do we need to have the client/tenant process parameter ? looks a strange one to me.

joseleite550: Adjustments made

Tenant will be obtained through context

Validation created for only Roles that the user has access removed

Adjustments to the naming of the Role_ID_From and Role_ID_To parameters

Improve the description of the process to make it clear that this will be destructive to previous Role to dat

hengsin (approved):
tested

PR #1782, opened 2023-04-14
IDEMPIERE-5671: Fixing Line net amount has error due to rounding.

dpansheriya:
# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base.callout/src/org/compiere/model/CalloutInvoice.java
+++ b/org.adempiere.base.callout/src/org/compiere/model/CalloutInvoice.java
@@ -666,7 +666,7 @@ public class CalloutInvoice extends CalloutEngine
 		}
 
 		//	Line Net Amt
-		BigDecimal LineNetAmt = QtyInvoiced.multiply(PriceActual);
+		BigDecimal LineNetAmt = QtyEntered.multiply(PriceEntered);
 		if (LineNetAmt.scale() > StdPrecision)
 			LineNetAmt = LineNetAmt.setScale(StdPrecision, RoundingMode.HALF_UP);
 		if (log.isLoggable(Level.INFO)) log.info("amt = LineNetAmt=" + LineNetAmt);
--- a/org.adempiere.base.callout/src/org/compiere/model/CalloutInvoice.java
+++ b/org.adempiere.base.callout/src/org/compiere/model/CalloutInvoice.java
@@ -602,7 +602,7 @@ public class CalloutInvoice extends CalloutEngine
 			pp.setM_PriceList_Version_ID(M_PriceList_Version_ID);
 			//
 			PriceEntered = MUOMConversion.convertProductFrom (ctx, M_Product_ID, 
-				C_UOM_To_ID, pp.getPriceStd());
+				C_UOM_To_ID, pp.getPriceStd(), 12);
 			if (PriceEntered == null)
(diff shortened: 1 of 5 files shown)


CarlosRuiz-globalqss (changes requested):
Hi @dpansheriya - I tested this successfully.

However, I think it is wrong to round the price based on the UOM precision.  The UOM precision is for quantities and is being used for rounding prices in several places.

Also, the prices can accept 12 decimal points when entered manually in the same product UOM, but the unit price is being rounded when there is a UOM conversion, I think both cases must accept the 12 decimals for consistency.

Can you please integrate the following patch:
[pr1782PR.patch.txt](https://github.com/idempiere/idempiere/files/11239585/pr1782PR.patch.txt)

dpansheriya:
> Hi @dpansheriya - I tested this successfully.
> 
> However, I think it is wrong to round the price based on the UOM precision. The UOM precision is for quantities and is being used for rounding prices in several places.
> 
> Also, the prices can accept 12 decimal points when entered manually in the same product UOM, but the unit price is being rounded when there is a UOM conversion, I think both cases must accept the 12 decimals for consistency.
> 
> Can you please integrate the following patch: [pr1782PR.patch.txt](https://github.com/idempiere/idempiere/files/11239585/pr1782PR.patch.txt)

Carlos,  I think 12 decimal hard coding is not good. I think we should round priceActual to costing precision as that is what used for costing too. If 12 digits precision printed on reports, it may not be good. also some how it must be controlled based on countries

CarlosRuiz-globalqss:
> Carlos, I think 12 decimal hard coding is not good. I think we should round priceActual to costing precision as that is what used for costing too. If 12 digits precision printed on reports, it may not be good. also some how it must be controlled based on countries

@dpansheriya - the 12 decimal hard coding is already there when you enter a price manually, so, I'm not introducing something new, but making the price rounding consistent when manual vs UOM-converted.

If I remember correctly that was implemented to solve a big issue when doing multiple invoices for a single order.
When a sales order generates multiple invoices it creates rounding issues on every invoice generated when the price is rounded to very few decimals - with 12 decimals there can be still issues with very large quantities, but the error is minimized.

About your concern when printing, in invoices the printed price is the PriceEntered, and if for any reason you need to print the UOM-converted price, then I think is more accurate to print 12.8033 than 12.8

dpansheriya:
> > Carlos, I think 12 decimal hard coding is not good. I think we should round priceActual to costing precision as that is what used for costing too. If 12 digits precision printed on reports, it may not be good. also some how it must be controlled based on countries
> 
> @dpansheriya - the 12 decimal hard coding is already there when you enter a price manually, so, I'm not introducing something new, but making the price rounding consistent when manual vs UOM-converted.
> 
> If I remember correctly that was implemented to solve a big issue when doing multiple invoices for a single order. When a sales order generates multiple invoices it creates rounding issues on every invoice generated when the price is rounded to very few decimals - with 12 decimals there can be still issues with very large quantities, but the error is minimized.
> 
> About your concern when printing, in invoices the printed price is the PriceEntered, and if for any reason you need to print the UOM-converted price, then I think is more accurate to print 12.8033 than 12.8
@CarlosRuiz-globalqss 
Applied your patch.

CarlosRuiz-globalqss (approved):

PR #1798, opened 2023-04-24
IDEMPIERE-5567 Fix NPE running Period Control Management

CarlosRuiz-globalqss:

A NPE was introduced with latest changes when running Period Control Management - or process from an info window that have an alias on the main table

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
@@ -2533,10 +2533,10 @@ public abstract class InfoPanel extends Window implements EventListener<Event>, 
 	 */
 	public void createT_Selection_InfoWindow(int AD_PInstance_ID)
 	{
-		MTable table = MTable.get(Env.getCtx(), getTableName());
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -2789,8 +2789,8 @@ public void zoom()
 	        fireValueChange(event);
     	}
     	else
-    	{    		
-    		int AD_Table_ID = MTable.getTable_ID(p_tableName);
+    	{
+    		int AD_Table_ID = infoWindow.getAD_Table_ID();
```
hengsin, line 2793: hi @CarlosRuiz-globalqss , zoom() also call from InfoGeneralPanel and infoWindow can be null here.

CarlosRuiz-globalqss: Hi @hengsin - I think the user cannot zoom from an InfoGeneralPanel, but anyways I added a check for null just in case.

hengsin (approved):
looks good

PR #1808, opened 2023-04-27
IDEMPIERE-5697 - Improve Process Log for more advanced logging options for Processes

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5697

# Pull Request Checklist


PeterTakacs300:
For the first round I've implemented the option to create and save a log immediately from a process with the `saveLog` methods

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +692,80 @@ public ProcessInfoLog[] getLogs()
...
+	 * @param P_Msg
+	 * @param tableId
+	 * @param recordId
+	 */
+	public void saveLog (int P_ID, Timestamp P_Date, BigDecimal P_Number, String P_Msg ,int tableId,int recordId)
+	{
+		saveLog (new ProcessInfoLog (P_ID, P_Date, P_Number, P_Msg,tableId, recordId));
+	}
+	
+	/**************************************************************************
```
hengsin, line 725: Just curious, why different style of javadoc is being used in the same file ?

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +692,80 @@ public ProcessInfoLog[] getLogs()
...
+	public void saveLog (int P_ID, Timestamp P_Date, BigDecimal P_Number, String P_Msg)
+	{
+		saveLog (new ProcessInfoLog (P_ID, P_Date, P_Number, P_Msg));
+	}	//	saveLog
+
+	/**
+	 * 	Save Log to DB
+	 *	@param logEntry log entry
+	 */
+	public void saveLog (ProcessInfoLog logEntry)
```
hengsin, line 754: javadoc should highlight the difference with existing addLog method where this is write to db immediately.

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
PeterTakacs300: That is a copy-paste issue...

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
PeterTakacs300: ok, fixed

org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
```diff
@@ -121,8 +165,18 @@ public String toString ()
 
 
 	private final static String insertSql = "INSERT INTO AD_PInstance_Log "
-			+ "(AD_PInstance_ID, Log_ID, P_Date, P_ID, P_Number, P_Msg, AD_Table_ID, Record_ID, AD_PInstance_Log_UU)"
-			+ " VALUES (?,?,?,?,?,?,?,?,?)";
+			+ "(AD_PInstance_ID, Log_ID, P_Date, P_ID, P_Number, P_Msg, AD_Table_ID, Record_ID, AD_PInstance_Log_UU, PInstanceLogType)"
+			+ " VALUES (?,?,?,?,?,?,?,?,?,?) "
+			+ " ON CONFLICT (AD_PInstance_Log_UU) "
```
hengsin, line 170: this is postgresql specific. please do the simple stuff here instead, separate insert and update sql

org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
```diff
@@ -145,27 +199,52 @@ public void saveEx ()
...
-		Object[] params = new Object[9];
-		params[0] = m_AD_PInstance_ID;
-		params[1] = m_Log_ID;
-		if (m_P_Date != null)
-			params[2] = m_P_Date;
-		if (m_P_ID != 0)
-			params[3] = m_P_ID;
-		if (m_P_Number != null)
-			params[4] = m_P_Number;
+		ArrayList<Object> params = new ArrayList <Object>();	// insert parameters
```
hengsin, line 202: please split insert and update

org.adempiere.base/src/org/compiere/model/X_AD_PInstance_Log.java
```diff
@@ -103,7 +103,7 @@ protected POInfo initPO (Properties ctx)
     public String toString()
     {
       StringBuilder sb = new StringBuilder ("X_AD_PInstance_Log[")
-        .append(get_ID()).append("]");
+        .append(get_UUID()).append("]");
```
hengsin, line 106: should log both id and uuid instead.

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +693,138 @@ public ProcessInfoLog[] getLogs()
...
+				logEntry.getP_Number(), 
+				logEntry.getP_Msg(),
+				logEntry.getAD_Table_ID(), 
+				logEntry.getRecord_ID(),
+				!Util.isEmpty(logEntry.getPInstanceLogType()) ? logEntry.getPInstanceLogType() : X_AD_PInstance_Log.PINSTANCELOGTYPE_Progress);
+		return il.save() ? il.getAD_PInstance_Log_UU() : "";
+	}	//	saveLog
+	
+	/**
+	 * Save Log to DB immediately
```
hengsin, line 777: should be update log instead.

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +693,138 @@ public ProcessInfoLog[] getLogs()
...
+				getAD_PInstance_ID(), 
+				logEntry.getLog_ID(), 
+				logEntry.getP_Date(),
+				logEntry.getP_ID(), 
+				logEntry.getP_Number(), 
+				logEntry.getP_Msg(),
+				logEntry.getAD_Table_ID(), 
+				logEntry.getRecord_ID(),
+				!Util.isEmpty(logEntry.getPInstanceLogType()) ? logEntry.getPInstanceLogType() : X_AD_PInstance_Log.PINSTANCELOGTYPE_Status);
+		return il.save();
```
hengsin, line 825: split insert and update - save() for insert and update() for update.

org.adempiere.base/src/org/compiere/process/SvrProcess.java
```diff
@@ -618,6 +618,85 @@ private void flushBufferLog () {
...
+		if (log.isLoggable(Level.INFO)) log.info(pInstanceLogUU + " - " + id + " - " + date + " - " + number + " - " + msg + " - " + tableId + " - " + recordId);
+	}	//	saveLog
+	
+	/**
+	 *  Save Log Entry with exact Log_ID, update if exists
+	 *  @param pInstanceLogUU AD_PInstance_Log_UU
+	 * 	@param id record id or 0
+	 *	@param date date or null
+	 * 	@param number number or null
+	 * 	@param msg message or null
```
hengsin, line 645: wrong javadoc

org.adempiere.base/src/org/compiere/process/SvrProcess.java
```diff
@@ -618,6 +618,85 @@ private void flushBufferLog () {
...
+	public void updateLog (String pInstanceLogUU, int id, Timestamp date, BigDecimal number, String msg)
+	{
+		if (m_pi != null)
+			m_pi.updateLog(pInstanceLogUU, id, date, number, msg);
+		
+		if (log.isLoggable(Level.INFO)) log.info(pInstanceLogUU + " - " + id + " - " + date + " - " + number + " - " + msg);
+	}	//	saveLog
+	
+	/**
+	 *  Save Log Entry with table name
```
hengsin, line 656: must highlight this save log to db immediately

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +693,138 @@ public ProcessInfoLog[] getLogs()
...
+		MPInstanceLog il = new MPInstanceLog(getAD_PInstance_ID(), 
+				logEntry.getLog_ID(), 
+				logEntry.getP_Date(),
+				logEntry.getP_ID(), 
+				logEntry.getP_Number(), 
+				logEntry.getP_Msg(),
+				logEntry.getAD_Table_ID(), 
+				logEntry.getRecord_ID(),
+				!Util.isEmpty(logEntry.getPInstanceLogType()) ? logEntry.getPInstanceLogType() : X_AD_PInstance_Log.PINSTANCELOGTYPE_Progress);
+		return il.save() ? il.getAD_PInstance_Log_UU() : "";
```
hengsin, line 773: use il.saveex(); return il.getAD_Instance_Log_UU() instead.

org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
PeterTakacs300: I see, I haven't checked that, ok, fixed in the latest commit

hengsin (changes requested):

PeterTakacs300:
hi @hengsin, requested changes applied

migration/iD11/oracle/202305020844_IDEMPIERE-5697.sql
```diff
@@ -0,0 +1,99 @@
...
+-- May 2, 2023, 9:10:25 AM CEST
+ALTER TABLE AD_PInstance_Log ADD PInstanceLogType VARCHAR2(3 CHAR) DEFAULT 'R' 
+;
+
+-- May 15, 2023, 10:36:39 AM CEST
+UPDATE AD_Column SET IsKey='N',Updated=TO_TIMESTAMP('2023-05-15 10:36:39','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=8779
+;
+
+-- May 15, 2023, 10:37:36 AM CEST
+ALTER TABLE AD_PInstance_Log MODIFY AD_PInstance_Log_UU VARCHAR2(36 CHAR) DEFAULT NULL
```
hengsin, line 53: is this needed ? seems unnecessary

org.adempiere.base/src/org/compiere/model/MPInstanceLog.java
```diff
@@ -62,6 +64,43 @@ public MPInstanceLog (int AD_PInstance_ID, int Log_ID, Timestamp P_Date,
 	 */
 	public MPInstanceLog (int AD_PInstance_ID, int Log_ID, Timestamp P_Date,
 			int P_ID, BigDecimal P_Number, String P_Msg, int AD_Table_ID, int Record_ID)
+	{
+		this(AD_PInstance_ID, Log_ID, P_Date, P_ID, P_Number, P_Msg, AD_Table_ID, Record_ID, null);
```
hengsin, line 68: should default to "R" for better backward compatibility

org.adempiere.base/src/org/compiere/process/ProcessInfo.java
```diff
@@ -691,6 +693,139 @@ public ProcessInfoLog[] getLogs()
...
+				logEntry.getP_Date(),
+				logEntry.getP_ID(), 
+				logEntry.getP_Number(), 
+				logEntry.getP_Msg(),
+				logEntry.getAD_Table_ID(), 
+				logEntry.getRecord_ID(),
+				!Util.isEmpty(logEntry.getPInstanceLogType()) ? logEntry.getPInstanceLogType() : X_AD_PInstance_Log.PINSTANCELOGTYPE_Status);
+		return il.update();
+	}	//	saveLog
+
```
hengsin, line 828: The multiple savelog and updatelog method here is heavily bias towards the log type of "P", would prefer a more balance API.

I'm not sure what to make up of the new methods added here as it seems you are using the "P"rogress log type in a very specific way but the API added here doesn't reflect that and there isn't any examples provided.

hengsin: hi @PeterTakacs300 , I think it is easier and more intuitive for developer to have "logProgress", "updateProgress" and "logStatus" method instead. Also, unless there's real need of it, not to have multiple variants of "logProgress" and "logStatus" method to keep the API simple.

WDYT?

PeterTakacs300: Hi @hengsin, I've made a new commit to the Generate Random Data for SO/PO process here: https://github.com/cloudempiere/org.cloudempiere.generaterandomorders/pull/2
There is an example of the use of the saveLog and updateLog methods.

The initial purpose of this PR was to add the possibility to create process audit logs while the process is running. I thought it could be quite intuitive to use save and update log methods (saveLog should always create a new log, updateLog updates existing log), and the log type is assigned automatically, so the implementor of the process should not worry about it.

Do you think it would be more balanced, if we would give setting the log type in the hands of the process implementor (by adding Log Type argument to the methods)?

hengsin: That's exactly what I means above, the method is design with just the "P" log type in mind. I've already posted my preference above (i.e replace saveLog and updateLog with logProgress, updateProgress and logStatus, the method name should be self explanatory here). However, if you prefer to just add a log type argument, that should be fine too.

hengsin: hi @PeterTakacs300 , what you have just committed is different from your message above. Please clarify.

PeterTakacs300: hi @hengsin, maybe there is a mismatch in the log type names in what you mentioned. If I'm correct it should be like this:
- Status log - displays always the current status - so this is the log type that can be updated
- Progress log - we originally named it 'Tail', changes to Progress after Chuck's suggestion - these are single logs saved continuously through the execution of the process
- Result log - legacy

Only Status logs can be updated.

In SvrProcess I renamed the methods as you suggested, but in the ProcessInfo layer I stayed with the saveLog/updateLog naming and added log type argument (to reduce the number of methods here, since the process developers usually do not use the logging methods from this level anyways). Can it stay like this, or ProcessInfo method naming should match the ones in SvrProcess?

hengsin: I think it is simpler to comprehend to have similar named method between the 2 class.

For me, Status and Progress should be the other way round, I think it is more common to say "Show Progress of a Process" (i.e 30%, 50%, etc).

PeterTakacs300: hi @hengsin, methods are renamed and logic exchanged between Status and Progress

hengsin:
[pr1808.patch.txt](https://github.com/idempiere/idempiere/files/11620497/pr1808.patch.txt)
- Fix Oracle migration script.

hengsin (approved):
tested

PR #1843, opened 2023-05-17
IDEMPIERE-5726 2Pack: TableElementHandler should reset Table cache af…

hengsin:
…ter commit of transaction

https://idempiere.atlassian.net/browse/IDEMPIERE-5726

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/TableElementHandler.java
+++ b/org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/TableElementHandler.java
@@ -45,8 +45,10 @@ import org.compiere.model.X_AD_Package_Exp_Detail;
 import org.compiere.model.X_AD_Package_Imp_Detail;
 import org.compiere.model.X_AD_Table;
 import org.compiere.process.DatabaseViewValidate;
+import org.compiere.util.CacheMgt;
 import org.compiere.util.Env;
 import org.compiere.util.Trx;
+import org.compiere.util.TrxEventListener;
 import org.xml.sax.SAXException;
 import org.xml.sax.helpers.AttributesImpl;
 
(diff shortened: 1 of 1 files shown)


org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/TableElementHandler.java
```diff
@@ -123,6 +125,27 @@ public void endElement(PIPOContext ctx, Element element) throws SAXException {
...
+			trx.addTrxEventListener(new TrxEventListener() {
+				
+				@Override
+				public void afterRollback(Trx trx, boolean success) {
+				}
+				
+				@Override
+				public void afterCommit(Trx trx, boolean success) {
+					if (success) {
+						CacheMgt.get().reset(MTable.Table_Name, mTable.get_ID());
```
CarlosRuiz-globalqss, line 140: Hi @hengsin - I have not tested, but checking the logic.

If there are several tables in the same 2Pack (same transaction), will this add several event listeners, calling cache reset multiple times?

CarlosRuiz-globalqss: Thanks @hengsin, more questions about,

* why is this needed, when this must be captured by the PO.saveFinish - not criticizing, but I think maybe there is something wrong with the PO reset cache because many times we need to do it manually

* I remember also changing a column, via 2Pack or manually, bring problems until cache reset is executed, why the PO doesn't reset that cache?
  * looking your approach here vs PO, could be related to the trx?  PO triggers the cache reset before the transaction is completed, if the column or table is read again in the cache, the committed version will be read, and that would be the obsolete record if is done before committed
  * if that's correct, maybe the same mechanism (TrxEventListener) can be implemented in PO.saveFinish?

* writing JUnit tests I have had also problems with the cache reset of single records, don't know why, but sometimes the unit test doesn't work with that and is needed to do a full cache reset, which can be bad in tests because of performance and parallelism
  * I remember for example in GridTabTest.testInsert - trying to remove the single field from the cache didn't work, a full cache reset was necessary, and we are full of that, there are 31 occurrences of `CacheMgt.get().reset();` in org.idempiere.test cases

hengsin: Immediate reset might not work because if the same record is read into cache again before commit of trx, it will get back the before change record.

hengsin: Well, that might work. Need to think further on this as your patch just increase the performance by 10x.

CarlosRuiz-globalqss (approved):
If you want to push this is OK.

My suggestion is just to make it more generic, and not specific for 2Pack.
But I leave it to you better.

org.adempiere.pipo.handlers/src/org/adempiere/pipo2/handler/TableElementHandler.java
hengsin: Ticket created for attached patch: https://idempiere.atlassian.net/browse/IDEMPIERE-5732

hengsin: I will push this first as it is for 2pack only and will have zero performance impact on a production system.

PR #1863, opened 2023-05-29
IDEMPIERE-5738 - Info Window ID columns are sorted by ID instead of display value

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5738

# Pull Request Checklist


PeterTakacs300:
The concept is to improve ColumnInfo by DisplayColumn which will be used for sorting, if it's not null. It is parsed in InfoWindow - createLookupColumnInfo method. Probably the parsing logic of the sql could take some improvements, I tried to adjust it to the use cases I found, which is primarily 2 types of display column:
- [ColumnName]
- NVL([ColumnName], '-1')

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -2991,7 +2995,16 @@ public void sort(Comparator<Object> cmpr, boolean ascending) {
...
 		indexOrderColumn = col;
 		isColumnSortAscending = ascending;
-		sqlOrderColumn = p_layout[col].getColSQL().trim();
+		String displayColumn = p_layout[col].getDisplayColumn();
+		sqlOrderColumn = !Util.isEmpty(displayColumn) ? displayColumn : p_layout[col].getColSQL().trim();
+
+		String columnName = p_layout[col].getColumnName();
+		if(columnName.endsWith("_ID")) {
+			int index = columnName.lastIndexOf("_ID");
+			String tableName = columnName = columnName.substring(0, index);
```
hengsin, line 3004: This doesn't always work, for e.g, you have AD_Reference_Value_ID but the foreign table is AD_Reference, not AD_Reference_Value.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -1051,10 +1055,44 @@ protected ColumnInfo createLookupColumnInfo(TableInfo[] tableInfos,
...
+	/**
+	 * Check and parse the correct aliases of the display SQL
+	 * @param displayColumn
+	 * @return parsed displayColumn
+	 */
+	private String parseAliases(String displayColumn, String tableName) {
+		if(Util.isEmpty(displayColumn))
+			return "";
+		String alias = getAlias(tableName);
+		if(!displayColumn.contains(".") && !displayColumn.startsWith("NVL")) {
```
hengsin, line 1071: That could be the common case in the core dictionary but I don't think there's any hard rule set on the display column sql here. What would happens here if the parse fail ? If it will just error out, I don't think this is acceptable.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: I added getByKeyColumns(keyColumns) method to MTable to fix that

hengsin: You have ad_reference_id and ad_reference_value_id in ad_infocolumn, should get the table name from there instead. Also, with the presence of ad_reference_id, we should check against that instead of using the "_ID" pattern.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
hengsin: Instead of hard coded function name, I think it is better to look for the pattern of {function name}({column name}).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2056,6 +2116,30 @@ protected String buildDataSQL(int start, int end) {
...
+		return "";
+	StringBuilder builder = new StringBuilder();
+	for(String tableName : joinTables) {
+		boolean doJoin = true;
+		for(TableInfo tableInfo : tableInfos) {
+			if(tableName.equalsIgnoreCase(tableInfo.getTableName()))
+				doJoin = false;
+		}
+		if(doJoin && !sqlMain.contains(" JOIN " + tableName)) {
+			builder.append(" JOIN ").append(tableName).append(" ON (")
```
hengsin, line 2135: inner join doesn't always work, the column could be optional and contains NULL value.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: Do you mean getting the table name from AD_Ref_Table? I've tested, and when on Info Column I have Table Direct as AD_Reference_ID, then AD_Reference_Value_ID is 0, so I can't get the table name from AD_Ref_Table in this case.

hengsin: If ad_reference_value_id is 0, you get the table name from column name. That must work because otherwise there's no other place to get the table name from.

PeterTakacs300: Ok thanks, done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
PeterTakacs300: Replaced with the right regex

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
PeterTakacs300: Ok, so I went with LEFT JOIN instead

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -3007,6 +3022,21 @@ public void sort(Comparator<Object> cmpr, boolean ascending) {
...
+	 * @param refValID
+	 * @param columnName
+	 * @return MTable[] tables
+	 */
+	private MTable[] getTables(int refValID, String columnName) {
+		if(refValID > 0) {
+			return new MTable[] {MTable.get(Env.getCtx(), MRefTable.get(Env.getCtx(), refValID).getAD_Table_ID())};
+		}
+		else {
+			return MTable.getByKeyColumns(Env.getCtx(), new String[] {columnName}, null);
```
hengsin, line 3036: This should be unnecessary. When AD_Reference_Value_ID is 0, you just need to remove the _ID part to get table name.

PeterTakacs300: ok, removed the unnecessary code

org.adempiere.base/src/org/compiere/model/MTable.java
```diff
@@ -915,7 +915,7 @@ public int getWindowIDFromMenu() {
 				+ "INNER JOIN AD_Menu m ON (a.AD_Window_ID=m.AD_Window_ID AND m.IsActive='Y' AND m.Action='W') "
 				+ "WHERE a.IsActive='Y' AND b.IsActive='Y' AND b.AD_Table_ID=? ORDER BY b.TabLevel, a.AD_Window_ID", getAD_Table_ID());
 	}
-
+
```
hengsin, line 918: It seems there's no changes being make to this class. You can use Replace With > Head Revision to remove this class from PR.

PeterTakacs300: This did not work as expected. Removed the whitespace change, but the class remains in the PR.

PeterTakacs300: No I did not, I do not see any remaining whitespace changes in the PR, there are identical lines as "removed" and "added". These were created in the diff, when I merged my branch with master because of a conflict in MTable

hengsin: ic, if Head Revision is not working, "Replace With" > "Branch, Tag, Reference" > "Remote Tracking" > "upstream/master" should did the trick.

PeterTakacs300: At the end `git checkout master -- [filename]` command in terminal and it worked, did not try your solution yet.

PeterTakacs300:
Discovered bug:
- define a where clause on e.g. order info definition in system (e.g. `1=1`)
- the parses tries to add the dynamic join clauses after the defined where clause

Fixed in the latest commit

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2056,6 +2123,30 @@ protected String buildDataSQL(int start, int end) {
 		return dataSql;
 	}
```
hengsin, line 2124: hi @PeterTakacs300 , since dynaJoint is in InfoPanel and m_sqlMain is created in InfoPanel, it seems it is better to move most of the changes here to InfoPanel instead (the prepareTable method).

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -2993,7 +2998,15 @@ public void sort(Comparator<Object> cmpr, boolean ascending) {
 		int col = lsc.getColumnIndex();
 		indexOrderColumn = col;
 		isColumnSortAscending = ascending;
-		sqlOrderColumn = p_layout[col].getColSQL().trim();
+		String displayColumn = p_layout[col].getDisplayColumn();
+		sqlOrderColumn = !Util.isEmpty(displayColumn) ? displayColumn : p_layout[col].getColSQL().trim();
+
+		// join tables to sort by display value
+		if(!Util.isEmpty(displayColumn) && (DisplayType.isID(p_layout[col].getAD_Reference_ID()) || DisplayType.isChosenMultipleSelection(p_layout[col].getAD_Reference_ID()))) {
```
hengsin, line 3005: should use DisplayType.isLookup instead of DisplayType.isID

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
PeterTakacs300: Hi @hengsin, I think that would not work well. In your case we would make the dynamic JOIN clauses a part of m_sqlMain (so they would not be dynamic anymore). For better performance, we should add them only if the InfoPanel.sort() method is called. But when sorting, InfoPanel.prepareTable() is not called anymore, thats why the parsing is done in InfoWindow.buildDataSQL(). Or am I missing something?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: ok, done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
hengsin: ok, that make sense. btw, in what use case would joinTables have > 1 entry ?

PeterTakacs300: Oh, you got me with this one, this needs to be fixed. 

That's a remain of an early concept of this solution. We should join only one table at a time since we can not sort multiple columns simultaneously... So currently if I'm sorting more ID-columns one after another, the array collects the tables and creates unnecessary joins.

Fixed in the latest commit.

PeterTakacs300:
Hi @hengsin, this has been pending here for some time, is it considered OK, or does it require further review?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -1353,7 +1357,8 @@ protected String getUserOrderClause() {
 	 * @return
 	 */
 	protected String getUserOrderClause(int col) {
-		String colsql = p_layout[col].getColSQL().trim();
```
hengsin, line 1356: should always reset joinTable to empty here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -2993,7 +2998,15 @@ public void sort(Comparator<Object> cmpr, boolean ascending) {
...
 		isColumnSortAscending = ascending;
-		sqlOrderColumn = p_layout[col].getColSQL().trim();
+		String displayColumn = p_layout[col].getDisplayColumn();
+		sqlOrderColumn = !Util.isEmpty(displayColumn) ? displayColumn : p_layout[col].getColSQL().trim();
+
+		// join tables to sort by display value
+		if(!Util.isEmpty(displayColumn) && (DisplayType.isLookup(p_layout[col].getAD_Reference_ID()) || DisplayType.isChosenMultipleSelection(p_layout[col].getAD_Reference_ID()))) {
+			MTable table = getTable(p_layout[col].getAD_Reference_Value_ID(), p_layout[col].getColumnName());
+			if(table != null && !table.getTableName().equals(joinTable))
+				joinTable = table.getTableName();
```
hengsin, line 3008: perhaps better to rename this as joinTableForUserOrder

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: done

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: variable renamed

PeterTakacs300:
Identified error case while testing:
- open order info
- sort by business partner
- set a parameter (e.g. Grand Total >= 60)
- requery -> error: missing join clause for table c_bpartner

Solution: the initialisation of joinTableForUserOrder need to be called in instead in `InfoPanel.getUserOrderClause` of `InfoPanel.sort`

Committing the fix...

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
hengsin: not seeing that changes, I means we should reset `joinTableForUserOrder = null` here

PeterTakacs300: In the latest commit I moved it to InfoWindow:2145.

I've found issues with Trl tables and using SELECT DISTINCT with ORDER BY - committing fixes soon..

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2056,6 +2123,29 @@ protected String buildDataSQL(int start, int end) {
...
+		StringBuilder builder = new StringBuilder();
+		for(TableInfo tableInfo : tableInfos) {
+			if(joinTableForUserOrder.equalsIgnoreCase(tableInfo.getTableName())) {
+				return "";
+			}
+		}
+		if(!sqlMain.contains(" LEFT JOIN " + joinTableForUserOrder)) {
+			builder.append(" LEFT JOIN ").append(joinTableForUserOrder).append(" ON (")
+				.append(joinTableForUserOrder).append(".").append(joinTableForUserOrder).append("_ID = ")
+				.append(getTableName()).append(".").append(joinTableForUserOrder).append("_ID)");
```
hengsin, line 2143: this needs to be changed to cater for https://idempiere.atlassian.net/browse/IDEMPIERE-5567

hengsin (changes requested):

PeterTakacs300:
Hi @hengsin,

I'm marking this PR as draft, because I'm not sure if this is a correct approach to add sorting.

With the added joins and order by some new issues appeared:
- if the display column (which I'm loading from the lookup) contains reference to Trl table, then the Trl table needs to be joined too - I reproduced it in our system by adding M_Product_ID to an info window
- if the info window is marked as IsDistinct=Y, then the ORDER BY expressions must appear in the SELECT DISTINCT clause too

Recently I was trying to fix these issues, I'm committing the result of it, but then some more issues appeared:
- with distinct info windows, by adding the order by expression into the select clause, some columns (e.g. Name, Value) often cause `column reference "value" is ambiguous` errors
- problems with column definitions like Vendor_ID with Search and Reference (here the column name does not match the display column: C_BPartner.Name) - this causes that the given column is not sorted correctly

Do you think the approach I started here can be correct? If yes, then I can provide a more detailed description of the problems.

hengsin:
hi @PeterTakacs300 , yes, join can be problematic since the potential of side effect from the joint. The safer but less efficient approach is to use sub query instead of outer joint, i.e, instead of left outer join c_bpartner, you use something similar to the following instead:

select documentno, c_bpartner_id, (select name from c_bpartner where c_bpartner.c_bpartner_id=c_order.c_bpartner_id) from c_order order by (select name from c_bpartner where c_bpartner.c_bpartner_id=c_order.c_bpartner_id);

PeterTakacs300:
Hi @hengsin, thanks for the suggestion, looks like this solution works better, it passed my basic tests on GardenWorld's Info Windows, the PR is ready for review.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2029,33 +2068,88 @@ else if (columnName.endsWith("_ID"))
...
 	}
 
+    /**
+     * Append ORDER BY expressions into the select list
+     * @param sql
+     * @param orderBy
+     * @return String parsed sql
+     */
+    private String appendOrderByToSelectList(String sql, String orderBy) {
+    	if(Util.isEmpty(sql) || Util.isEmpty(orderBy))
```
hengsin, line 2117: this is harmless but I think it is impossible for sql to be empty here.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2029,33 +2068,88 @@ else if (columnName.endsWith("_ID"))
...
+	
+		for(int i = 0; i < sql.length(); i++) {
+			// identify and ignore sub-query
+			char c = sql.charAt(i);
+				if (c == ')')
+					parenthesisLevel--;
+				else if (c == '(')
+					parenthesisLevel++;
+	
+				if(sql.substring(i).startsWith("FROM") && parenthesisLevel == 0)
```
hengsin, line 2146: safer check should be <white space>from<white space>

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
```diff
@@ -2029,33 +2068,88 @@ else if (columnName.endsWith("_ID"))
...
+     */
+    private String appendOrderByToSelectList(String sql, String orderBy) {
+    	if(Util.isEmpty(sql) || Util.isEmpty(orderBy))
+    		return sql;
+		int idxFrom = getIdxFrom(sql);
+		if(idxFrom < 0)
+			return sql;
+	
+		String select = sql.substring(0, idxFrom);
+		select += ", " + orderBy.replace("ORDER BY", "").replace("ASC", "").replace("DESC", "");
```
hengsin, line 2124: Order by should be safe, but need to include white space for asc and desc (you can have column that contain this character)

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -1318,21 +1320,23 @@ protected void validateOrderIndex() {
 		if (indexOrderColumn > 0 && (indexOrderColumn + 1 > p_layout.length || !p_layout[indexOrderColumn].getColSQL().trim().equals(sqlOrderColumn))) {
 			// try to find out new index of ordered column, in case has other column is hide or display
 			for (int testIndex = 0; testIndex < p_layout.length; testIndex++) {
-				if (p_layout[testIndex].getColSQL().trim().equals(sqlOrderColumn)) {
+				if (p_layout[testIndex].getColSQL().trim().equals(sqlOrderColumn) || sqlOrderColumn.equals(p_layout[testIndex].getDisplayColumn())) {
```
hengsin, line 1323: can sqlOrderColumn be null here ?

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -1385,10 +1411,68 @@ else if (!(last == '*' || last == '-' || last == '+' || last == '/' || last == '
...
+		String[] keyCols = table.getKeyColumns();
+		String sqlSelect = orderColumnInfo.getSelectClause();
+		String whereClause = "";
 		
-		return String.format(" ORDER BY %s %s ", colsql, isColumnSortAscending? "" : "DESC");
+		for(String keyCol : keyCols) {
+			if(!Util.isEmpty(whereClause))
+				whereClause += " AND ";
+			whereClause += tableName + "." + keyCol + " = " + sqlSelect;
+		}
```
hengsin, line 1472: This for loop is actually pointless. There's only 1 sqlSelect value, this will only works if keyCols is of length 1.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -3008,6 +3108,24 @@ public void sort(Comparator<Object> cmpr, boolean ascending) {
...
+	 * @return MTable[] tables
+	 */
+	private MTable getTable(int refValID, String columnName) {
+		if(refValID > 0) {
+			return MTable.get(Env.getCtx(), MRefTable.get(Env.getCtx(), refValID).getAD_Table_ID());
+		}
+		else if (columnName.endsWith("_ID")) {
+			return MTable.get(Env.getCtx(), columnName.substring(0, columnName.length() - 3));
+		}
+		else {
```
hengsin, line 3124: with the uuid only table changs, I think you can have columnName ends with _UU here as well.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
PeterTakacs300: In the latest commit, I've modified ColumnInfo.getDisplayColumn() and these if statements to be NPE safe.

PeterTakacs300:
Hi @hengsin, changes applied

hengsin (approved):
tested

PR #1864, opened 2023-05-29
IDEMPIERE-5683 Exception: NO Data found for Record_ID

CarlosRuiz-globalqss:
Fixing - Because of the ordering of scripts - the latest changes on IDEMPIERE-5683 destroyed changes on IDEMPIERE-5567


CarlosRuiz-globalqss:
forcing push as is a fix for broken functionality already approved

PR #1867, opened 2023-05-30
IDEMPIERE-4836 Adding Access tab on Document Status (Activity) - fix issue with native sequence

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4836?focusedCommentId=49630


CarlosRuiz-globalqss:
forced push as tested in real-life implementation

PR #1895, opened 2023-06-10
IDEMPIERE-5567 Support of UUID as Key (FHCA-4195) - Archive

CarlosRuiz-globalqss:

# Pull Request Checklist

### Tests


org.adempiere.base/src/org/compiere/model/MArchive.java
```diff
@@ -402,12 +406,26 @@ public File saveAsZip() {
 	 * @param Record_ID
 	 * @param trxName
 	 * @return int[], [0] = report count and [1] = document count
+	 * @deprecated - use {@link #getReportAndDocumentCountByRecordId(int, int, String, String)} instead
 	 */
 	public static int[] getReportAndDocumentCountByRecordId(int AD_Table_ID, int Record_ID, String trxName) {
+		return getReportAndDocumentCountByRecordId(AD_Table_ID, Record_ID, null, trxName);
```
hengsin, line 412: this couple with the Record_UU changes below is not just deprecation but it is a breaking change.

CarlosRuiz-globalqss: Thanks, fixed with commit https://github.com/idempiere/idempiere/pull/1895/commits/1654f4595e835024cc3467d2148e6034dff5d4f5

org.adempiere.base/src/org/compiere/model/MArchive.java
```diff
@@ -342,7 +345,8 @@ public void setStorageProvider(MStorageProvider p) {
 	 * @return File - the temporary file
 	 */
 	public File saveAsZip() {
-		String name = MTable.get(Env.getCtx(), getAD_Table_ID()).getTableName() + "_" + getRecord_ID();
+		String name = MTable.get(Env.getCtx(), getAD_Table_ID()).getTableName() + "_"
+				+ (getRecord_ID() > 0 ? getRecord_ID() : getRecord_UU());
```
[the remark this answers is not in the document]
CarlosRuiz-globalqss: Hi @hengsin - not a big deal, I tested packing out the archive for the Test_UU table, this is used for a temporary folder with very short live, the temp folder had the name:
/tmp/TestUU_8858ecc2-cf1d-405f-987f-793536037e76_9617073054397170185

the folder is just 64 bytes, even with a 27 chars tablename it would be 85 chars, very little for the supported names in linux or windows.

org.adempiere.base/src/org/compiere/model/MArchive.java
```diff
@@ -402,12 +410,26 @@ public File saveAsZip() {
...
+	 * @param Record_UU
+	 * @param trxName
+	 * @return int[], [0] = report count and [1] = document count
+	 */
+	public static int[] getReportAndDocumentCountByRecordId(int AD_Table_ID, int Record_ID, String Record_UU, String trxName) {
 		int reportCount = 0;
 		int documentCount = 0;
 		StringBuilder sql = new StringBuilder("SELECT IsReport, COUNT(*) FROM AD_Archive ")
-				.append("WHERE (AD_Table_ID=? AND Record_ID=?) ");
+				.append("WHERE (AD_Table_ID=? AND (Record_ID=? OR Record_UU=?)) ");
```
hengsin, line 432: I don't have a big enough data set to analyse this but I'm concern that this would be inefficient not able to make use of the ad_archive_tab_rec index.

CarlosRuiz-globalqss: > I don't have a big enough data set to analyse this but I'm concern that this would be inefficient and not able to make use of the ad_archive_tab_rec index.

Thanks @hengsin 

Improved with commit https://github.com/idempiere/idempiere/pull/1895/commits/cd72987fc0be9cba22714b417cf65a7205a46b49

I inserted 75.000 records in AD_Archive distributed over several tables, records and BPs and tested the query in postgresql.

Effectively, because of the OR clause postgresql was not able to do a proper usage of the new index, as the Record_UU is now basically always filled I think is better to do the query just with that.

org.adempiere.base/src/org/compiere/model/MArchive.java
hengsin: hi @CarlosRuiz-globalqss , with the removal of OR, this method is not going to work.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MArchive.java
```diff
@@ -402,12 +410,26 @@ public File saveAsZip() {
...
 	 */
 	public static int[] getReportAndDocumentCountByRecordId(int AD_Table_ID, int Record_ID, String trxName) {
+		return getReportAndDocumentCountByRecordId(AD_Table_ID, Record_ID, null, trxName);
+	}
+
+	/**
+	 * Get number of document and report archive by table and record id
+	 * 
+	 * @param AD_Table_ID
+	 * @param Record_ID
```
hengsin, line 423: this parameter is basically ignore and not doing anything here ...

CarlosRuiz-globalqss: > this parameter is basically ignore and not doing anything here ...

It is still used for the case of Business Partner window.  When checking the archives in BP window, the query counts all the documents related to the current BP, independent of the table.

hengsin: ic, miss that. however, this become a typical case of you wouldn't know what the method parameter is for until you read through the code in it.

CarlosRuiz-globalqss:
Hi @hengsin 

With commit https://github.com/idempiere/idempiere/pull/1895/commits/02ca9d4fbfcfdc5560fd7a160300f9a64c58f11a I improved better the method, so it uses ID when the UUID is not received, making the deprecated function still usable.

With commit https://github.com/idempiere/idempiere/pull/1895/commits/7e567e2a5a61bc951c4e29d311f1fcdbf23a627d I changed to ensure the UUID is always filled, even when called from foreign sources or points of the code where still is not set.

hengsin (approved):
tested

PR #1904, opened 2023-06-22
IDEMPIERE-5364 event name is truncated at bottom

uthadehikaru:
minor bug fix style for event calendar


hengsin:
replace by #1905

PR #1920, opened 2023-06-30
IDEMPIERE-5750 Error when switching to tab when choosing a language other than English

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5750

# Pull Request Checklist

### Tests


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/AdempiereIdGenerator.java
```diff
@@ -106,6 +107,8 @@ private static String getWidgetName(String widgetClass) {
 	}
 
 	public static String escapeId(String prefix) {
+		if (! "true".equals(System.getProperty("UseAdempiereIdGenerator.escapeId")))
+			return UUID.randomUUID().toString();
 		Pattern pattern = Pattern.compile("[^a-zA-Z_0-9]");
 		Matcher matcher = pattern.matcher(prefix);
```
hengsin, line 113: I think it is better to skip those setid call if it is not running selenium, i.e `if ("true".equals(System.getProperty("ZkUnitTest") setId(AdempiereIdGenerator.escapeId(...))`. The proposed patch work but it is not good for performance.

CarlosRuiz-globalqss: Hi @hengsin - I tested first simply returning null when not selenium, and it worked fine for the case reported on this ticket, and there are more setId that seems like can be skipped safely.

But I saw a lot of code also using the AdempiereIdGenerator.escapeId for setting setWidgetAttribute("instanceName"...

And I'm not sure how that setWidgetAttribute can affect it if we skip it completely too.

Can those be skipped safely too?

Also, I think it would be good if we collect in a single class all the System.getProperty spread in the system, but not sure how must be the name to start such class and where it must  be located.

hengsin: Those setWidgetAttribute call is for selenium test too. This is even safer to skip than the setId call.

CarlosRuiz-globalqss: Ready @hengsin - changed as per your suggestion

PR #1948, opened 2023-07-20
IDEMPIERE-4735 apply patch from Webservice CreateUpdate. PO passed to…

uthadehikaru:
… Call to WSValidator is not valid

this merge request is to apply patch from jira


hengsin (approved):
trivial fix, good to go

PR #1956, opened 2023-07-28
IDEMPIERE-5811 - Fix NPE on event for BroadcastMessages / fix kill session feature

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5811?focusedCommentId=49884


CarlosRuiz-globalqss:
forcing push because of an important feature broken

PR #1957, opened 2023-07-29
IDEMPIERE-5816 Upgrade Hazelcast version (FHCA-4436)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5816

- fix NPE when no hazelcast.xml file is in path or in bundle
- remove the library hazelcast-aws.jar - not needed anymore

I conducted tests with two amazon instances and it worked fine without the library as those classes are now included in hazelcast.jar


hengsin (approved):
looks good

PR #2006, opened 2023-09-13

CarlosRuiz-globalqss:
- fix for broken PackIn on AD_Menu.Parent_ID


CarlosRuiz-globalqss:
forced merge as is a bug fix for packin

PR #2013, opened 2023-09-18

CarlosRuiz-globalqss:

- Fix for MAttachment.get
- discovered while testing rest endpoing api/v1/models/testuu/{uuid}/attachments/zip

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAttachment.java
+++ b/org.adempiere.base/src/org/compiere/model/MAttachment.java
@@ -67,7 +67,6 @@ public class MAttachment extends X_AD_Attachment
 	 * @param AD_Table_ID
 	 * @param Record_ID
 	 * @return attachment or null
-	 * @deprecated Use {@link #get(Properties, int, int, String, String)}
 	 */
 	public static MAttachment get (Properties ctx, int AD_Table_ID, int Record_ID)
 	{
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/model/MAttachment.java
```diff
@@ -79,13 +81,42 @@ public static MAttachment get (Properties ctx, int AD_Table_ID, int Record_ID)
...
+		List<Object> params = new ArrayList<Object>();
+		params.add(AD_Table_ID);
+		if (Record_ID > 0) {
+			whereClause.append(" AND Record_ID=?");
+			params.add(Record_ID);
+		}
+		if (!Util.isEmpty(Record_UU)) {
+			whereClause.append(" AND Record_UU=?");
+			params.add(Record_UU);
+		}
```
hengsin, line 112: While it is harmless, it seems rather strange that the where clause would include both Record_ID=? and Record_UU=?.

org.adempiere.base/src/org/compiere/model/MAttachment.java
```diff
@@ -58,18 +59,19 @@ public class MAttachment extends X_AD_Attachment
...
-	private static final long serialVersionUID = 1447398065894212273L;
+	private static final long serialVersionUID = 5615231734722570658L;
 
 	/**
 	 * 
 	 * @param ctx
 	 * @param AD_Table_ID
 	 * @param Record_ID
 	 * @return attachment or null
+	 * @deprecated Use {@link #get(Properties, int, int, String, String)}
```
hengsin, line 70: I found it is kind of counter intuitive that this is mark as deprecated. When I know the target table has _ID key, it seems rather strange that I should supplied the UUID value as well to retrieve the attachment record.

hengsin (approved):
looks good

PR #2043, opened 2023-10-03
IDEMPIERE-5857 Cross tenant on Reset Password - remove eclipse warning

CarlosRuiz-globalqss:
minor remove an eclipse warning from previous commit


hengsin:
I've actually included that as part of https://github.com/idempiere/idempiere/pull/2038

PR #2056, opened 2023-10-15
Update CONTRIBUTING.md

d-ruiz:


CarlosRuiz-globalqss (changes requested):
@d-ruiz - good to add the link to the workflow, but I think is still valuable to have the link to the [Contributing to Core](https://wiki.idempiere.org/en/Contributing_to_Core) page, as that's not covered in the workflow page.

Also the last line about documenting the new feature is not covered there.

d-ruiz:
Changed

CarlosRuiz-globalqss (approved):

PR #2068, opened 2023-10-19

CarlosRuiz-globalqss:
- The process Create Window from Table is broken because it doesn't find the tab UUID (out of trx)

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/CreateWindowFromTable.java
+++ b/org.adempiere.base.process/src/org/compiere/process/CreateWindowFromTable.java
@@ -35,6 +35,7 @@ import org.compiere.model.MProcessPara;
 import org.compiere.model.MTab;
 import org.compiere.model.MTable;
 import org.compiere.model.MWindow;
+import org.compiere.model.PO;
 import org.compiere.model.SystemIDs;
 import org.compiere.util.DB;
 import org.compiere.util.Msg;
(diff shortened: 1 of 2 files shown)


hengsin:
hi @CarlosRuiz-globalqss , the changes doesn't seems to be related to the description above.

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , the changes doesn't seems to be related to the description above.

Hi @hengsin 

The problem is in the methods PO.checkRecordUUCrossTenant and PO.checkRecordIDCrossTenant

In this case the class is creating a tab within a transaction, and is creating a process instance out of transaction.

I was using the saveCrossTenantSafeEx to skip the validation, but, just tried a better approach setting the transaction of the process instance and it worked fine.  Committed that change.

org.adempiere.base.process/src/org/compiere/process/CreateWindowFromTable.java
```diff
@@ -205,7 +205,7 @@ else if (table.columnExistsInDB("Name"))
 			ProcessInfo processInfo = new ProcessInfo("", SystemIDs.PROCESS_AD_TAB_CREATEFIELDS, MTab.Table_ID, tab.getAD_Tab_ID(), tab.getAD_Tab_UU());
 
 			MPInstance instance = new MPInstance(getCtx(), SystemIDs.PROCESS_AD_TAB_CREATEFIELDS, MTab.Table_ID, tab.getAD_Tab_ID(), tab.getAD_Tab_UU());
-			instance.saveEx();
+			instance.saveEx(get_TrxName());
```
hengsin, line 208: This is inconsistent with current iDempiere practise and there's an obvious issue - if there's any error with the process, the process audit is loss since trx will be rollback.

hengsin (changes requested):

org.adempiere.base.process/src/org/compiere/process/CreateWindowFromTable.java
hengsin: I think the right fix is to copy the `if (I_AD_PInstance.Table_Name.equals(p_info.getTableName()))` check from checkRecordIDCrossTenant to checkRecordUUCrossTenant

CarlosRuiz-globalqss: Thanks  @hengsin - you're totally right - I missed that part

hengsin (approved):
trivial fix, good to go

PR #2070, opened 2023-10-20
IDEMPIERE-5889 - Cannot Open Lookup Info Window if no Window is Defined for the Table

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5889

# Pull Request Checklist

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
@@ -43,8 +43,10 @@ import org.compiere.model.I_C_ElementValue;
 import org.compiere.model.MColumn;
 import org.compiere.model.MLookupFactory;
 import org.compiere.model.MReference;
+import org.compiere.model.MTab;
 import org.compiere.model.MTable;
 import org.compiere.model.PO;
+import org.compiere.model.Query;
 import org.compiere.util.DB;
 import org.compiere.util.DisplayType;
 import org.compiere.util.Env;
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
```diff
@@ -490,29 +497,43 @@ private boolean initInfoTable ()
...
+					+ " INNER JOIN AD_Field f ON (tab.AD_Tab_ID=f.AD_Tab_ID AND f.AD_Column_ID=c.AD_Column_ID) ");
+		}	
+		sqlc.append( "WHERE t.AD_Table_ID=? ");
+		if(hasWindowAndTab) {
+			sqlc.append(" AND tab.IsSortTab='N' "
+					+ " AND tab.Ad_Tab_ID=(SELECT MIN(mt.AD_Tab_ID) FROM AD_tab mt WHERE mt.AD_Window_ID=? AND mt.AD_Table_ID=t.AD_Table_ID AND mt.IsActive='Y')"
+					+ " AND (c.IsKey='Y' OR "
+					+ "		(f.IsEncrypted='N' AND f.ObscureType IS NULL))");
+		} else {
+			sqlc.append(" AND c.IsKey='Y'");
```
hengsin, line 520: This cause the info window to display only column with IsKey=Y. For RV_C_Invoice, that's the C_Invoice_ID column. This will cause the execution of query fail since order by is hardcoded to "Order By 2" in this class (find the prepareTable call). Removing this condition resolve the error but the resulting info window doesn't looks very useful either. 

From my testing, without AD_Window definition to base on, the resulting info window looks not very useful and I'm wondering what's the real world use case here ?

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
norbertbede: you have right, probably not to sexy, but we would like avoid NPE - improve auto-generate info window. another side, avoid to click to icon to make invalid info window cause NPE.

**our use case:** we have a reference on top of materialised view subset of users. eg. role=businessfunction sales  then we show in sales order only salesreps - in this case we would like show search for this subset. BUT, we have 10 basic subsets (warehouse, accountant, purchasing etc.) - this would require 10 ref. lists with specialised info window.

hengsin: Anyway, the current patch with `IsKey=Y` is not working since it would return only 1 column from the underlying view. We would have to either remove the condition or change it to return more columns (for e.g, `(IsKey='Y' OR IsIdentifier='Y' OR IsParent='Y' OR IsSelectionColumn='Y' OR Upper(ColumnName) IN ('NAME','VALUE','DESCRIPTION','DOCUMENTNO') OR Upper(ColumnName) Like '%_NAME' OR Upper(ColumnName) Like '%_Value')`

PeterTakacs300: hi @hengsin, tested your example, in e.g. the case of RV_C_Invoice it returns the following columns in GW:

Screenshot 2023-10-27 at 08 19 23

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
```diff
@@ -490,29 +497,49 @@ private boolean initInfoTable ()
...
+					+ "		(f.IsEncrypted='N' AND f.ObscureType IS NULL))");
+		} else {
+			sqlc.append(" AND (c.IsKey='Y' "
+					+ "			OR IsIdentifier='Y' "
+					+ "			OR IsParent='Y' "
+					+ "			OR IsSelectionColumn='Y' "
+					+ "			OR Upper(ColumnName) IN ('NAME','VALUE','DESCRIPTION','DOCUMENTNO') "
+					+ "			OR Upper(ColumnName) Like '%_NAME' "
+					+ "			OR Upper(ColumnName) Like '%_Value') ");
+		}
```
hengsin, line 527: hi @PeterTakacs300 , please use c.<ColumnName> instead of just <ColumnName> (c.IsIdentifier, c.IsParent, etc)

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoGeneralPanel.java
PeterTakacs300: change committed

hengsin (approved):
tested

PR #2073, opened 2023-10-22
IDEMPIERE-4639 simplify calls using saveCrossTenantSafeEx when possible

CarlosRuiz-globalqss:
This is just helping for readability


hengsin:
hi @CarlosRuiz-globalqss , please consider the following patch to address nested `saveCrossTenantSafeEx()` call (due to model validator or event handler).

[pr2073.patch.txt](https://github.com/idempiere/idempiere/files/13063184/pr2073.patch.txt)

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , please consider the following patch to address nested `saveCrossTenantSafeEx()` call (due to model validator or event handler).

Thanks @hengsin - integrated your patch and added it in the other method

CarlosRuiz-globalqss:
390/390 unit tests passed

hengsin (approved):
looks good

PR #2130, opened 2023-12-07
IDEMPIERE-5952 - Adjust to return IDColumn recordID

VitorHenri:
# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
@@ -1717,19 +1717,21 @@ public abstract class InfoPanel extends Window implements EventListener<Event>, 
 	}   //  getIntSelectedRowKey
 
 	/**
-     *  Get the keys of selected row/s based on layout defined in prepareTable
-     *  @return IDs if selection present
-     *  author ashley
+     *  Get the keys of selected rows
+     *  @return selected IDs or UUIDs
      */
-	protected ArrayList<Object> getSelectedRowKeys() {
-		ArrayList<Object> selectedDataList = new ArrayList<Object>();
-
+    @SuppressWarnings("unchecked")
+	protected <T extends Serializable> List<T> getSelectedRowKeys()
+    {
+    	List<T> selectedDataList = new ArrayList<>();
 		for (Map.Entry<Object, List<Object>> rowInfo : getSelectedRowInfo().entrySet()) {
 			if(rowInfo.getValue().get(0) instanceof IDColumn idColumn)
-				selectedDataList.add(idColumn.getRecord_ID());
+				selectedDataList.add((T)idColumn.getRecord_ID());
+			else if(rowInfo.getValue().get(0) instanceof UUIDColumn uuidColumn)
+				selectedDataList.add((T)uuidColumn.getRecord_UU());
 		}
 		return selectedDataList;
-	}  //  getSelectedRowKeys
+    }   //  getSelectedRowKeys
 
     /**
 	 *	Get selected Keys as Collection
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/InfoPanel.java
```diff
@@ -1718,28 +1718,18 @@ protected Integer getIntSelectedRowKey(int tableId)
...
-        }
-
-        return selectedDataList;
-    }   //  getSelectedRowKeys
+		for (Map.Entry<Object, List<Object>> rowInfo : getSelectedRowInfo().entrySet()) {
+			if(rowInfo.getValue().get(0) instanceof IDColumn idColumn)
+				selectedDataList.add(idColumn.getRecord_ID());
+		}
+		return selectedDataList;
+	}  //  getSelectedRowKeys
```
hengsin, line 1732: isn't the change needed here is just to remove the if (p_multipleSelection) condition ?

VitorHenri: Hi @hengsin, If we remove this condition if(p_multipleSelection) the feature where we can select multiple items and then create multiple records in the window will stop working, as we do in the Sales Order lines by selecting multiple products.

hengsin: No, I means it can be changed to as below instead:
'''
protected ArrayList<Object> getSelectedRowKeys()
{
        ArrayList<Object> selectedDataList = new ArrayList<Object>();
        Collection<Object> lsKeyValueOfSelectedRow = getSelectedRowInfo().keySet();
        for (Object key : lsKeyValueOfSelectedRow){        		
        	selectedDataList.add(key);
        }
       return selectedDataList;
}
'''

VitorHenri: Hi @hengsin, It would not change the problem, the situation is that when Single Selection we get the IDColumn(The keycolumn of Table set in MInfoWindow)

When getSelectedRowKey() we ends up here:

But if we have multiple selection and a keycolumn defined on MInfoColumn we are going to get its value,but i think we should always get the key of table defined on MInfoWindow.
The key below can be differente from IDColumn when a keyColumn on MInfoColumn is defined it will be returned

hengsin: I understand where you are coming from here but that's not a backward compatible change.

VitorHenri: Hello @hengsin, do you think of any other way to deal with this behavior or do you have anything to suggest?

hengsin: hi @VitorHenri , it seems that method is only use by InfoWindow and is probably safe to ignore the backward compatibility issue. However, the changes above still need the following refinements:
- Follow return type of getSelectedRowKey, so it should return `List<T extends Serializable>` instead of `ArrayList<Object>`.
- Follow the implementation of getSelectedRowKey that cater for both IDColumn and UUIDColumn
- Remove the `@deprecated` javadoc tag.
- Remove author ashley from javadoc
- In javadoc, replace "Get the keys of selected row/s based on layout defined in prepareTable" with "Get the keys of selected rows" 
- In javadoc, replace "@return IDs if selection present" with "@return selected IDs or UUIDs"

hengsin (changes requested):
please review the suggested refinements above.

VitorHenri:
Hi @hengsin - I will apply the adjustments

hengsin (approved):
looks good

PR #2170, opened 2023-12-30
IDEMPIERE-5567 fix PO.get_Translation for Zero ID tables

CarlosRuiz-globalqss:


CarlosRuiz-globalqss:
forced push as stabilization of release-11 bugs

PR #2171, opened 2023-12-30
IDEMPIERE-5943 Implement table partitioning support - fix menu

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5943

- minor fix to menu - make it centrally maintained and correct the process name

- this created an issue when generating the wiki Manual as the menu name is different than the process name - which is unusual in core


hengsin (approved):
ok

PR #2172, opened 2023-12-30
IDEMPIERE-5567 Fix NPE when AD_RecentItem.Record_UU is null

CarlosRuiz-globalqss:
- this can happen for example when a recent item record doesn't have or doesn't manage UUID


CarlosRuiz-globalqss:
forcing push as emergency fix for release-11

PR #2174, opened 2024-01-02
IDEMPIERE-5973 - Jasper Report does not allow to Change Print Format or Language

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5973

Based on patch by @alanlesc1 

# Pull Request Checklist


PeterTakacs300:
After creating multiple tickets around jasper reports with several fixes/improvements, there are issues repeatedly causing problems across all tickets, so I will now try to summarise here what would we want to achieve.

Before reading this summary, please read [the ticket description](https://idempiere.atlassian.net/browse/IDEMPIERE-5973).

Now, basically we want to improve the usage of the ZKReportViewer (has print format and language dropdown) vs the ZKJRViewer (no print format and language dropdown). 

Until now, if I see it correctly after debugging, every time we were dealing with Jasper reports, the ZKJRViewer was opened. With the commit based on Alain's patch seems like fixing a specific use case, when first a standard report is started (invoice print in our test case), and then it starts the jasper report. In [AbstractProcessCtl.java](https://github.com/idempiere/idempiere/blob/master/org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java#L189) if the jasper process is run directly, isJasper is set to true, but in our case through the document print process, isJasper is false and the jasper process is started only in [ReportCtl.startDocumentPrint()](https://github.com/idempiere/idempiere/blob/master/org.adempiere.ui/src/org/compiere/print/ReportCtl.java#L479) method.

Our idea with Norbert is, that if a print format can be found for the jasper report, then it should be opened in the standard ZKReportViewer. The problem is, that in the [if(isJasper) block](https://github.com/idempiere/idempiere/blob/master/org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java#L242) we don't have a report engine, which finds the print format on the other case.

@hengsin I'd like to ask your opinion about this change, would it be ok, to decide which viewer to open based on whether a print format is found or not. I thought it would be better to discuss it before implementing.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -171,9 +171,6 @@ public PrintData getPrintData (Properties ctx, MPrintFormat format, MQuery query
 		if (format == null)
 			throw new IllegalStateException ("No print format");
 
-		 if (format.getJasperProcess_ID() > 0)
-			 return null;
-
```
hengsin, line 176: this is a dangerous change, you don't want to call getPrintDataInfo and loadPrintData for Jasper Report print format.

PeterTakacs300: the problem is that having no print data is considered as an illegal state for the report engine: https://github.com/idempiere/idempiere/blob/master/org.adempiere.base/src/org/compiere/print/ReportEngine.java#L391

hengsin: that's implementation detail, logically, it doesn't make any sense to call loadPrintData for jasper report.

PeterTakacs300: Before fixing this issue, what do you think about [this suggestion](https://idempiere.atlassian.net/browse/IDEMPIERE-5973?focusedCommentId=50440)?

hengsin: If you can make it work, that sounds fine to me.

PeterTakacs300:
hi @hengsin, in the latest commits I tried to fix the issue with the getPrintData you reported, added a bunch of null checks and improved the layout() method of the ReportEngine.

About the suggested change, I started to implement a method that will be able to find a Print Format for jasper report if exists, so the standard report viewer can be opened. The implementation is a work in progress, for now I try to find the print format from _PInstance_, _ProcessInfo_ or with _JasperProcess_ID and AD_Table_ID_ (this is an issue for views for now). Could you take a look at it if it could be a good approach before I finish the implementation?

org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java
```diff
@@ -290,6 +313,34 @@ public void run ()
...
+		else if(pi.getTransientObject() != null && pi.getTransientObject() instanceof MPrintFormat) {
+			return ((MPrintFormat) pi.getTransientObject()).getAD_PrintFormat_ID();
+		}
+		// find for table and process
+		if(pi.getAD_Process_ID() > 0 && pinstance.getAD_Table_ID() > 0) {
+			String sqlWhere = "JasperProcess_ID = ? AND AD_Table_ID = ?";
+			return new Query(Env.getCtx(), MPrintFormat.Table_Name, sqlWhere, null)
+					.setParameters(pi.getAD_Process_ID(), pinstance.getAD_Table_ID())
+					.firstId();
+		}
```
hengsin, line 340: If jasper report is running from a print format, that print format must have been saved into either AD_PInstance or ProcessInfo, don't think there's need to check via AD_Table_ID as well.

PeterTakacs300: I see thanks
I've removed the unnecessary code & marked the PR as ready for review

org.adempiere.ui/src/org/compiere/apps/AbstractProcessCtl.java
```diff
@@ -21,12 +21,15 @@
...
+import org.compiere.model.Query;
+import org.compiere.print.MPrintFormat;
 import org.compiere.print.ReportCtl;
 import org.compiere.process.ProcessInfo;
 import org.compiere.process.ProcessInfoUtil;
 import org.compiere.util.CLogger;
+import org.compiere.util.DB;
 import org.compiere.util.Env;
 import org.compiere.util.Ini;
 import org.compiere.util.Msg;
```
hengsin, line 35: Please fix the following warnings:
- The import org.compiere.model.Query is never used
- The import org.compiere.util.DB is never used

hengsin (changes requested):
A few issues with the report viewer:
- NPE for Datatable, XLS and XLSX
- Summary checkbox is visible for jasper report. Should be hidden since it is not applicable.
- Row count show zero which is confusing. If that can't be fixed, should hide it for jasper report.
- Don't have a jasper report print format to check whether the language drop down is working. Is that tested ?

PeterTakacs300:
Hi @hengsin, I've addressed your requests except the translations look like doesn't work yet.

I've also removed the "findPrintFormat" part from AbstractProcessCtl, to me it looks like a wrong approach after I've looked at this ticket after quite a long time.

First there is a bigger issue with this implementation: jasper reports only show in the standard ZkReportViewer if there is Print Data loaded. I recalled that earlier you mentioned, that with jasper reports loading the Print Data does not make sense, now I just temporarily enabled it for testing purposes with a TODO message.

Could not find an easy solution, without Print Data the standard report renderers (HTMLReportRenderer, PDFReportRenderer etc.) are showing just empty reports. 

I think I should not use these standard renderers, instead modify ZkReportViewer to be able to support rendering media as it is in [ZkJRViewer.onRenderReport()](https://github.com/idempiere/idempiere/blob/master/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/ZkJRViewer.java#L811).

I don't really have a deeper understanding how ZkJRViewer works yet, do you see any other, better, simpler, more elegant solutions for this?

hengsin:
hi @PeterTakacs300 ,

I think the right approach is when we open a jasper report in ZkReportViewer, the viewer needs to know that it is rendering a jasper report and uses the code from ZkJRViewer to do the export and rendering (will need quite a bit of refactoring of ZkJRViewer to make that possible).

For me, your current approach doesn't looks right and it is very fragile.

Regards,
Low

CarlosRuiz-globalqss:
I deployed this pull request into https://test-feature.idempiere.org:3243/webui/ in case you want to test it there.

norbertbede:
hi @PeterTakacs300 we agreed with @hengsin that he will continue this ticket. it's too challenging for us. Norbert

PR #2183, opened 2024-01-11
IDEMPIERE-5992 - Move the getIdentifier methods from WRecordEditor to MLookup class

PeterTakacs300:
JIRA ticket: https://idempiere.atlassian.net/browse/IDEMPIERE-5992

# Pull Request Checklist

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MLookup.java
+++ b/org.adempiere.base/src/org/compiere/model/MLookup.java
@@ -1004,7 +1004,7 @@ public final class MLookup extends Lookup implements Serializable
 	 * @return null if tableID <= 0 or the table doesn't have any key column, else {@link MLookup}
 	 */
 	public static MLookup getRecordsLookup(int tableID, int windowNo, int tabNo) {
-		return getRecordsLookup(tableID, windowNo, tabNo, null);
+		return getRecordsLookup(tableID, windowNo, tabNo, false);
 	}
 	
 	/**
@@ -1012,26 +1012,27 @@ public final class MLookup extends Lookup implements Serializable
 	 * @param tableID
 	 * @param windowNo
 	 * @param tabNo
-	 * @param keyColumn - override the default key column for the lookup
+	 * @param useUUIDKey - default false
 	 * @return null if tableID <= 0 or the table doesn't have any key column, else {@link MLookup}
 	 */
-	public static MLookup getRecordsLookup(int tableID, int windowNo, int tabNo, String keyColumn) {
+	public static MLookup getRecordsLookup(int tableID, int windowNo, int tabNo, boolean useUUIDKey) {
 		if(tableID <= 0)	
 			return null;
 		MTable mTable = MTable.get(Env.getCtx(), tableID, null);
 		
 		// load key column
-		if(Util.isEmpty(keyColumn)) {
+		String keyColumn = "";
+		if(!useUUIDKey) {
 			String[] keyColumns = mTable.getKeyColumns();
 			// the table has a single key column
 			if(keyColumns != null && keyColumns.length == 1) 
 				keyColumn = keyColumns[0];
-			// fallback to UU
-			else
-				keyColumn = PO.getUUIDColumnName(mTable.getTableName());
+		}
+		if(Util.isEmpty(keyColumn)) {
+			keyColumn = PO.getUUIDColumnName(mTable.getTableName());
 		}
 		
-		if (Util.isEmpty(keyColumn))
+		if(Util.isEmpty(keyColumn))
 			return null;
 
 		MColumn mColumn = MColumn.get(Env.getCtx(), mTable.getTableName(), keyColumn);
@@ -1059,7 +1060,7 @@ public final class MLookup extends Lookup implements Serializable
 	 * @return String
 	 */
 	public static String getIdentifier(int tableID, Serializable recordID, int windowNo, int tabNo) {
-		return getIdentifier(tableID, recordID, windowNo, tabNo, null);
+		return getIdentifier(tableID, recordID, windowNo, tabNo, false);
 	}
 	
 	/**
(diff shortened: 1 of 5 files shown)


PeterTakacs300:
I've added a keyColumn parameter to preserve the functionality of the abstract getKeyColumn method in WRecordEditor, also, the logic around the tableBased parameter in the getRecordsLookup method should work as before, but I'm not sure if it is used correctly for future use cases.

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,148 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
 		return vnpCache;
 	}
 	
+
+
```
hengsin, line 1000: please do not add multiple line of spaces, it served no purpose.

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,148 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
...
+	 * @return null if tableID <= 0 or the table doesn't have any key column, else {@link MLookup}
+	 */
+	public static MLookup getRecordsLookup(int tableID, boolean tableBased, int windowNo, int tabNo) {
+		return getRecordsLookup(tableID, tableBased, windowNo, tabNo, null);
+	}
+	
+	/**
+	 * Get Lookup
+	 * @param tableID
+	 * @param tableBased - if the lookup checks for the UUID Key Column based on the table (true), or on the columnName (false)
```
hengsin, line 1039: This fail to mention this is ignore if keyColumn argument is not null.
Also, it got nothing to do with whether the method is "Table Based", from what I've goes through below, it is more like whether it is using a table's integer id key column (if available) or UUID key column. In other word, both scenario is "table based".

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,148 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
...
+			return null;
+		MTable mTable = MTable.get(Env.getCtx(), tableID, null);
+		// load key column
+		if(Util.isEmpty(keyColumn)) {
+			if (tableBased) {
+				String[] keyColumns = mTable.getKeyColumns();
+				keyColumn = keyColumns != null && keyColumns.length > 0 ? keyColumns[0] : null;
+			} else {				
+				keyColumn = PO.getUUIDColumnName(mTable.getTableName());
+			}
```
hengsin, line 1056: I don't quite get it why do you need an argument (tableBased) for this. The method should be using integer id key column if the table has a single integer id key column and if that's not the case, fallback to uuid key column.

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,148 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
...
+		return new MLookup(lookupInfo, tabNo);
+	}
+
+	/**
+	 * Get Identifier String from AD_Table_ID and Record_ID
+	 * @param tableID
+	 * @param recordID
+	 * @return String
+	 */
+	public static String getIdentifier(int tableID, Object recordID) {
```
hengsin, line 1073: better to use Serializable for record id

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,148 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
...
+		return lookup != null ? lookup.getDisplay(recordID) : "";
+	}
+
+	/**
+	 * Get Parent Identifier String from AD_Table_ID and Record_ID
+	 * @param tableID
+	 * @param recordID
+	 * @return String
+	 */
+	public static String getParentIdentifier(int tableID, Object recordID) {
```
hengsin, line 1110: How is this is different from getIdentier ? looks the same to me.

org.adempiere.base/src/org/compiere/model/MLookup.java
PeterTakacs300: I've removed the tableBased argument completely, but I think the keyColumn is needed, since in the RecordEditor you want to tell if you want to use the _UU column as a key - in case of RecordUUEditor you want to set the UU column always. Also this reuses the abstract getKeyColumn method of RecordEditor

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -995,6 +996,86 @@ private synchronized static CCache<String, ValueNamePair> getDirectValueNamePair
...
+	public static MLookup getRecordsLookup(int tableID, int windowNo, int tabNo) {
+		return getRecordsLookup(tableID, windowNo, tabNo, null);
+	}
+	
+	/**
+	 * Get Lookup
+	 * @param tableID
+	 * @param windowNo
+	 * @param tabNo
+	 * @param keyColumn - override the default key column for the lookup
```
hengsin, line 1015: In practise, this is either using the table's single key column or using uuid column. In otherword, you just need a boolean flag, say "useUUIDKey" to control that. With this change, I think you don't need the getKeyColumn(table) method at WRecordIDEditor either.

PeterTakacs300: Ok, I've replaced the keyColumn parameter with a boolean flag, but I think instead of the `getKeyColumn(table) `method we still should use at least a `boolean isUseUUIDKey()` abstract method, because most of the calls of the getIdentifier methods are in the abstract WRecordEditor class. All changes committed.

hengsin (approved):
looks good

PR #2184, opened 2024-01-11
IDEMPIERE-5994 Connection leak when using DB.prepareStatement(badSql, null)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5994

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/db/PreparedStatementProxy.java
+++ b/org.adempiere.base/src/org/compiere/db/PreparedStatementProxy.java
@@ -81,7 +81,6 @@ public class PreparedStatementProxy extends StatementProxy {
 	 * Initialize the prepared statement wrapper object
 	 */
 	protected void init() {
-		boolean localConn = false;
 		try {
 			Connection conn = null;
 			Trx trx = p_vo.getTrxName() == null ? null : Trx.get(p_vo
@@ -89,7 +88,6 @@ public class PreparedStatementProxy extends StatementProxy {
 			if (trx != null) {
 				conn = trx.getConnection();
 			} else {
-				localConn = true;
 				m_conn = AutoCommitConnectionBroker.getConnection();
 				conn = m_conn;
 			}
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/db/PreparedStatementProxy.java
```diff
@@ -95,6 +98,12 @@ protected void init() {
 			p_stmt = conn.prepareStatement(p_vo.getSql(), p_vo
 					.getResultSetType(), p_vo.getResultSetConcurrency());
 		} catch (Exception e) {
+			if (localConn && m_conn != null) {
+				try {
+					m_conn.close();
+					m_conn = null;
+				} catch (SQLException e1) {}
+			}
```
hengsin, line 106: doesn't need the localConn stuff, just need to close this.close() here.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/db/PreparedStatementProxy.java
CarlosRuiz-globalqss: Hi @hengsin 
this.close() is not available at this point
the "localconn" is to avoid closing a connection that is managed elsewhere, I just need to close here the connections created locally in the same method.

hengsin: ok, you need to change StatementProxy.close to protected. Calling the close method of StatementProxy should be the correct approach here, not the method above.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/db/PreparedStatementProxy.java
CarlosRuiz-globalqss: Thanks @hengsin - tested your suggestion and it worked fine.

hengsin (approved):
looks good

PR #2185, opened 2024-01-11
IDEMPIERE-5989 Attachments are not accessible after upgrade from Idempiere 10 to Idempiere 11

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5989

IDEMPIERE-5567
Fix issue when the window doesn't have the UUID field

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTab.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTab.java
@@ -2121,11 +2121,16 @@ public class GridTab implements DataStatusListener, Evaluatee, Serializable
 	 *	Get Attachment_ID for current record.
 	 *	@return AD_AttachmentID or 0 if no attachment.
 	 */
+	@SuppressWarnings("deprecation")
 	public int getAD_AttachmentID()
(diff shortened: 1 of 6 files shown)


org.adempiere.base/src/org/compiere/model/GridTab.java
```diff
@@ -2121,11 +2121,16 @@ public boolean hasAttachment()
 	 *	Get Attachment_ID for current record.
 	 *	@return AD_AttachmentID or 0 if no attachment.
 	 */
+	@SuppressWarnings("deprecation")
```
hengsin, line 2124: if the method is still needed at so many places, I think we just shouldn't mark it as deprecated.

CarlosRuiz-globalqss: Thanks @hengsin 
Committed a slightly different approach - more consistent with other points with same Record_ID+UU discovery

hengsin (approved):
looks good

PR #2189, opened 2024-01-14
IDEMPIERE-5991:improve file download by append file list inside process result panel

hieplq:
# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/adempiere/util/IProcessUI.java
+++ b/org.adempiere.base/src/org/adempiere/util/IProcessUI.java
@@ -117,6 +117,7 @@ public interface IProcessUI {
 
 	/**
 	 * add to list of file available for download after process end 
+	 * think about to use {@link SvrProcess#addDownloadFile(File)} 
 	 * @param file
 	 */
 	public void download(File file);
--- a/org.adempiere.base/src/org/adempiere/util/IProcessUI.java
+++ b/org.adempiere.base/src/org/adempiere/util/IProcessUI.java
@@ -116,8 +116,8 @@ public interface IProcessUI {
 	}
 
 	/**
-	 * add to list of file available for download after process end 
-	 * think about to use {@link SvrProcess#addDownloadFile(File)} 
+	 * Add to list of file available for download after process end<br/> 
+	 * While this API is limited to processes with user interfaces, consider using {@link SvrProcess#addDownloadFile(File)} for broader support, as it stores file download information in the database 
 	 * @param file
 	 */
 	public void download(File file);
--- a/org.adempiere.base/src/org/compiere/process/SvrProcess.java
+++ b/org.adempiere.base/src/org/compiere/process/SvrProcess.java
@@ -16,6 +16,7 @@
  *****************************************************************************/
 package org.compiere.process;
 
+import java.io.File;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Type;
@@ -42,6 +43,7 @@ import org.adempiere.base.event.IEventTopics;
 import org.adempiere.util.IProcessUI;
 import org.compiere.model.MPInstance;
 import org.compiere.model.PO;
+import org.compiere.model.X_AD_PInstance_Log;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.Env;
@@ -84,7 +86,14 @@ public abstract class SvrProcess implements ProcessCall
 	 */
 	public void addBufferLog(int id, Timestamp date, BigDecimal number, String msg, int tableId ,int recordId) {
 		ProcessInfoLog entryLog = new ProcessInfoLog(id, date, number, msg, tableId, recordId);
-		
+		addBufferLog(entryLog);
+	}
+	
+	/**
+	 * Add log to buffer, only process total success, flush buffer
+	 * @param entryLog
+	 */
+	public void addBufferLog(ProcessInfoLog entryLog) {
 		if (listEntryLog == null)
 			listEntryLog = new ArrayList<ProcessInfoLog>();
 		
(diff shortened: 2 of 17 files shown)


hengsin:
hi @CarlosRuiz-globalqss  & @hieplq , I think not prompting user to download files should be just an alternative UI option. IMHO, depending on the user case involve (and the target user group), not prompting to download is not necessary the better behaviour.

org.adempiere.base/src/org/adempiere/util/IProcessUI.java
```diff
@@ -117,6 +117,7 @@ default public void askForInput(String message, MLookup lookup, int displayType,
 
 	/**
 	 * add to list of file available for download after process end 
+	 * think about to use {@link SvrProcess#addDownloadFile(File)}
```
hengsin, line 120: hi @hieplq, it would be good to perform a readability check on the comments added to code (for e.g through bard or chatgpt) to make sure it is easily understandable by others. also, if you do a mouse over check in Eclipse for the javadoc popup (that's what your API user will usually see), the javadoc above is probably not in the easily readable category.

hieplq:
> hi @CarlosRuiz-globalqss & @hieplq , I think not prompting user to download files should be just an alternative UI option. IMHO, depending on the user case involve (and the target user group), not prompting to download is not necessary the better UI option.

I adjust by "Introduce a new configuration named PROCESS_SHOW_SEPARATE_DOWNLOAD_DIALOG to control how downloaded files are displayed. By default, this configuration is set to true, meaning a separate download dialog will appear for each downloaded file"

org.adempiere.base/src/org/adempiere/util/IProcessUI.java
hieplq: thanks for suggest it's helpful
i ask him by this one
this's good english sentence for clarity and readability "my bad english"

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss & @hieplq , I think not prompting user to download files should be just an alternative UI option. IMHO, depending on the user case involve (and the target user group), not prompting to download is not necessary the better UI option.

Hi @hengsin 

Sorry, I'm a bit lost here (I haven't tested).

I don't think we need a new SysConfig, for me this sounds like a usability improvement, opening just one dialog for the user instead of two.

But I'm not following the part of "not prompting user to download files" - I think that's not the intention of the ticket, if that can happen, it sounds wrong.

Also two things:
- I think the migration script creating a test for this fix is not needed, to provide a test case like that is better to add a 2Pack in the ticket
- I don't think we need a SysConfig, but if needed, then I think it must be Tenant level and required in the migration script

hieplq:
> hi @CarlosRuiz-globalqss , there's 2 code path here
> 
>     * running a process from button
> 
>     * running a process from application menu
> 
> 
> Base on my understanding, the changes here will not open the file download dialog for both use case. @hieplq , correct me if I'm wrong here.

Instead of showing a separate download dialog, this improve appends the download link directly to the results dialog
I have images that illustrates the updated layout at https://idempiere.atlassian.net/browse/IDEMPIERE-5991?focusedCommentId=50502

hieplq:
> * I think the migration script creating a test for this fix is not needed, to provide a test case like that is better to add a 2Pack in the ticket

This test demonstrates the 'allow user to download file after process execution' feature, serving as a reference for future development and Idempiere demos.
Consider relocating this test to the Garden World client for better organization and accessibility.
To avoid importing test-related content on production servers, maybe creating a dedicated test plugin that centralizes similar test stuff, similar to the current Test window

hengsin:
> > hi @CarlosRuiz-globalqss , there's 2 code path here
> > ```
> > * running a process from button
> > 
> > * running a process from application menu
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Base on my understanding, the changes here will not open the file download dialog for both use case. @hieplq , correct me if I'm wrong here.
> 
> Instead of showing a separate download dialog, this improve appends the download link directly to the results dialog I have images that illustrates the updated layout at https://idempiere.atlassian.net/browse/IDEMPIERE-5991?focusedCommentId=50502

- For the first 2 image, while I agree is probably better to merge the 2 popup dialog into one, I think the layout can be better. In current design, I don't think it is obvious that those link is the files produce by the process for user to download.
- For the third image, this is where it is not always the desire or better UI. For many use case, you want to prompt the user to download the files produce by the process.

norbertbede:
would be great to add the ux as **gift/scereencast** before - current version behaviour and - new proposal design. these changes are important but ui, ux must be clear  to avoid troubles with users. thanks

CarlosRuiz-globalqss:
> This test demonstrates the 'allow user to download file after process execution' feature, serving as a reference for future development and Idempiere demos. Consider relocating this test to the Garden World client for better organization and accessibility. To avoid importing test-related content on production servers, maybe creating a dedicated test plugin that centralizes similar test stuff, similar to the current Test window

I still think is not a good idea to have an example in iDempiere code for something that specific.

Note how Heng Sin has this project https://github.com/hengsin/idempiere-examples/ where he collects his examples to show how some code must work.

This process will be on menu for every implementation.

CarlosRuiz-globalqss:
I'm wondering if it is absolutely necessary to save the files in the Log?

Because this was intended to be a "just UI" cosmetic change, but changing how the logs are saved becomes a big change that can potentially not be backward compatible.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/AEnv.java
```diff
@@ -993,4 +997,36 @@ public static void detachInputElement(Component parent) {
...
+		return AEnv.getDownloadLinkFromLog(downloadFile);
+	}
+	
+	/**
+	 * Create a link for download file<br/>
+	 * In case of error show error message on link text
+	 * @param downloadFile
+	 * @return
+	 */
+	public static Component getDownloadLinkFromLog(File downloadFile) {
```
hengsin, line 1017: this method name is misleading, should be just "getDownloadLinkForFile(File)"

org.adempiere.base/src/org/compiere/process/SvrProcess.java
```diff
@@ -901,4 +910,33 @@ private List<Field> getFieldsWithParameters(Class<?> clazz) {
...
+	 * For backward compatibility call {@link IProcessUI#download(File)}<br/>
+	 * Stored file path in the {@link ProcessInfoLog#m_P_Msg} with the {@link ProcessInfoLog#getPInstanceLogType} set to {@link X_AD_PInstance_Log#PINSTANCELOGTYPE_FilePath} 
+	 * @param file
+	 * @param onlySuccess normal file download keep on buffer log so appear only success case. In case want to show file ever un-success process then parse false  
+	 */
+	public void addDownloadFile(File file, boolean onlySuccess) {
+		if (processUI != null)
+			processUI.download(file);
+		
+		// store file path to audit log to get back for case infowindow
```
hengsin, line 931: this is actually less useful than it seems due to:
- file is usually temporary so the link will become invalid after a while
- in a distribute environment, the link is invalid when you look at it from a different node.

hieplq:
> would be great to add the ux as **gift/scereencast** before - current version behaviour and - new proposal design. these changes are important but ui, ux must be clear to avoid troubles with users. thanks

There are pictures of both the before and after state on my ticket

org.adempiere.base/src/org/compiere/process/SvrProcess.java
hieplq: @CarlosRuiz-globalqss I store only file paths. I use log types to distinguish between them so isn't big change

@hengsin I believe the concerns about data validity in a distributed environment are mitigated by this data use only for process instance

store file path also to support resolve IDEMPIERE-5997

hengsin: @hieplq , stored file path is not a reliable solution for IDEMPIERE-5997. In a multiple server node environment, you might land on a different server node on your next login and the stored path wouldn't work.

hieplq: I believe storing the file path in the log is acceptable for now. Since the job still runs on the old server, it can retrieve the files and complete its tasks (sending emails and creating attachments) even after finishing execution.

To address your concern about log, I've decided to close this pull request and propose a new solution that doesn't rely on logs. I've moved the downloadFiles from AbstractProcessDialog to ProcessInfo instead

PR #2190, opened 2024-01-14
IDEMPIERE-5567 - Parallel unit tests fail frequently with deadlock

CarlosRuiz-globalqss:

when executing unit tests in parallel, MTestUUTest fails frequently with error detected deadlock this can be caused by testReadingUpdatingTestUU trying to update translations and at the same time the testDeletingTestUU is deleting the same record and deleting the same translations

# Pull Request Checklist

### Tests

### DIFF
--- a/org.idempiere.test/src/org/idempiere/test/model/MTestUUTest.java
+++ b/org.idempiere.test/src/org/idempiere/test/model/MTestUUTest.java
@@ -38,12 +38,15 @@ import org.compiere.util.Env;
 import org.compiere.util.Util;
 import org.idempiere.test.AbstractTestCase;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 
 /**
  *
  * @author Carlos Ruiz - globalqss - bxservice
  *
  */
+@Execution(ExecutionMode.SAME_THREAD)
 public class MTestUUTest extends AbstractTestCase {
 
 	private static final String TestRecordInGardenWorld = "8858ecc2-cf1d-405f-987f-793536037e76";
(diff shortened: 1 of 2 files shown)


org.idempiere.test/src/org/idempiere/test/model/MTestUUTest.java
```diff
@@ -82,10 +82,20 @@ public void testInsertingTestUU() {
...
+		// insert a new record
+		MTestUU testuu = new MTestUU(ctx, PO.UUID_NEW_RECORD, trxName);
+		testuu.setName("Test UU record created on JUnit test");
+		testuu.saveEx();
+		testuu.load(trxName);
+	    assertEquals("Test UU record created on JUnit test", testuu.getName());
+	    String uukey = testuu.getTestUU_UU();
+
+		MTestUU testuuToDelete = new MTestUU(ctx, uukey, trxName);
+		testuuToDelete.deleteEx(true);
```
hengsin, line 96: before calling delete, should assert `testuuToDelete.get_UUID().equals(uuKey)` is true.

org.idempiere.test/src/org/idempiere/test/model/MTestUUTest.java
```diff
@@ -82,10 +82,20 @@ public void testInsertingTestUU() {
 	public void testDeletingTestUU() {
 		Properties ctx = Env.getCtx();
 		String trxName = getTrxName();
-		MTestUU testuu = new MTestUU(ctx, TestRecordInGardenWorld, trxName);
-		testuu.deleteEx(true);
-		MTestUU testuu2 = new MTestUU(ctx, TestRecordInGardenWorld, trxName);
-	    assertFalse(testuu2.get_UUID().equals(TestRecordInGardenWorld));
+
+		// deleting the TestRecordInGardenWorld is creating a deadlock with testReadingUpdatingTestUU in parallel execution
```
hengsin, line 86: This is not a suggestion but would to point out that the alternative solution for this is to add `@Execution(ExecutionMode.SAME_THREAD)` annotation to the MTestUUTest class.

hengsin:
Beside the issue noted here, there's a strange setName(null) call in MTestUU.java (line 60) that produce unnecessary console/server log.

 if ("".equals(Test_UU)) {
	setName(null);
 }

org.idempiere.test/src/org/idempiere/test/model/MTestUUTest.java
CarlosRuiz-globalqss: Thanks @hengsin - your suggestion sounds better.

I did a test and it worked fine, but this is sporadic, so let's push it and see if jenkins complains again.

hengsin (approved):
looks good

PR #2194, opened 2024-01-16
IDEMPIERE-5996 Record Access is not being updated when a new record is added

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-5996

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MRole.java
+++ b/org.adempiere.base/src/org/compiere/model/MRole.java
@@ -64,10 +64,10 @@ import org.idempiere.cache.POCopyCache;
  */
 public final class MRole extends X_AD_Role implements ImmutablePOSupport
 {
-	/**
-	 * generated serial id
+    /**
+	 * 
 	 */
-	private static final long serialVersionUID = -8937680640915708588L;
+	private static final long serialVersionUID = 7266911648463503849L;
 
 	/**
 	 * 	Get role for current session/context
@@ -95,7 +95,7 @@ public final class MRole extends X_AD_Role implements ImmutablePOSupport
 		int AD_Role_ID = Env.getContextAsInt(ctx, Env.AD_ROLE_ID);
 		int AD_User_ID = Env.getContextAsInt(ctx, Env.AD_USER_ID);
 
-		MRole defaultRole = getDefaultRole(); 
+		MRole defaultRole = getDefaultRole(ctx, AD_Role_ID, AD_User_ID); 
 		if (reload || defaultRole == null)
 		{
 			defaultRole = get (ctx, AD_Role_ID, AD_User_ID, reload);
@@ -124,7 +124,24 @@ public final class MRole extends X_AD_Role implements ImmutablePOSupport
 	 * @return MRole
 	 */
 	private static MRole getDefaultRole() {
-		return (MRole) Env.getCtx().get(ROLE_KEY);
+		return getDefaultRole(Env.getCtx(), Env.getAD_Role_ID(Env.getCtx()), Env.getAD_User_ID(Env.getCtx()));
+	}
+
+	/**
+	 * Get role for current session/context
+	 * @param ctx
+	 * @param AD_Role_ID
+	 * @param AD_User_ID
+	 * @return MRole
+	 */
+	private static MRole getDefaultRole(Properties ctx, int AD_Role_ID, int AD_User_ID) {
+		MRole role = (MRole) ctx.get(ROLE_KEY);
+		String key = AD_Role_ID + "_" + AD_User_ID;
+		if (! s_roles.containsKey(key)) {
+			ctx.remove(ROLE_KEY);
+			role = null;
+		}
+		return role;
 	}
 
 	/**
@@ -467,11 +484,7 @@ public final class MRole extends X_AD_Role implements ImmutablePOSupport
 		//
 		else if (is_ValueChanged("UserLevel"))
 			updateAccessRecords();
-		
-		//	Default Role changed
-		if (getDefaultRole() != null 
-			&& getDefaultRole().get_ID() == get_ID())
-			setDefaultRole(this);
+
 		return success;
 	}	//	afterSave
 	
--- a/org.adempiere.base/src/org/compiere/model/MRole.java
+++ b/org.adempiere.base/src/org/compiere/model/MRole.java
@@ -37,7 +37,6 @@ import java.util.logging.Level;
 
 import org.adempiere.exceptions.AdempiereException;
 import org.adempiere.process.UUIDGenerator;
-import org.compiere.util.CCache;
 import org.compiere.util.CLogger;
 import org.compiere.util.DB;
 import org.compiere.util.DisplayType;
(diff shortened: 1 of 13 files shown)


org.adempiere.base/src/org/compiere/util/CacheMgt.java
```diff
@@ -271,7 +271,7 @@ public int resetLocalCache()
 		CacheInterface[] instances = getInstancesAsArray();
 		for (CacheInterface stored : instances)
 		{
-			if (stored != null && stored.size() > 0)
+			if (stored != null && stored instanceof CCache)
```
hengsin, line 274: this change is puzzling, why do you wants to reset an empty cache ?

org.adempiere.base/src/org/compiere/model/MRole.java
```diff
@@ -3428,4 +3441,31 @@ public MRole markImmutable() {
...
+		}
+
+		@Override
+		public int reset(int recordId) {
+			if (recordId == Env.getAD_Role_ID(Env.getCtx()))
+				getDefault(Env.getCtx(), true); // reload the default role
+			return super.reset(recordId);
+		}
+
+	}
```
hengsin, line 3469: I don't think we should use cache listener to reload MRole.getDefault since that's not using cache at all. Should just call mrole.getdefault from the model class.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
CarlosRuiz-globalqss: > this change is puzzling, why do you wants to reset an empty cache ?

I noticed that the reset method of the CCacheListener was being called when updating a record, but not when doing a full Cache Reset

The reason is that the resetLocalCache(String,int) is comparing "stored instanceof CCache", while the resetLocalCache() is comparing "stored.size() > 0" - so, the CCacheListener is not called at all if the cache is empty, IMO we don't know which logic can be in those reset methods and it must be consistently calling on both cases.

I'm wondering now too if the getElementCount() has also a problem there, because is checking if "stored.size() > 0" but below is using a different method sizeNoExpire() to get the number.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
```diff
@@ -373,7 +373,7 @@ public int getElementCount()
...
```
hengsin, line 376: we already has instanceof check below and there's nothing wrong with the size() > 0 check. sizeNoExpire() will never return > 0 if size() is 0.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
hengsin: If we wants the checking to be more consistent, we can add the size() > 0 check to resetLocalCache(String,int) , not removing the check at resetLocalCache(). The reason we have the instanceof check at resetLocalCache(String,int) is because we have to cast it to CCache inside that if code block.

org.adempiere.base/src/org/compiere/model/MRole.java
hengsin: The right approach should be through using the IMessageService instead (Core.getMessageService()). The web client, upon login should subscribe to the "Reload Role" message topic, keep a reference to the desktop and when message arrive for that topic, schedule a event (Executions.schedule) to reload the role. In MRole's afterSave and beforeSave method, you will publish the "Reload Role" message to the "Reload Role" topic.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
CarlosRuiz-globalqss: > If we wants the checking to be more consistent, we can add the size() > 0 check to resetLocalCache(String,int) , not removing the check at resetLocalCache(). The reason we have the instanceof check at resetLocalCache(String,int) is because we have to cast it to CCache inside that if code block.

The approach that I'm trying to implement here (using cache listener in other sessions) doesn't work with the "stored.size() > 0" validation because then is not triggered.
That's why I noticed the inconsistency, it was being triggered by a specific record reset, but not by the general process "Cache Reset".

org.adempiere.base/src/org/compiere/model/MRole.java
hengsin: Another option is to address this is to add a loaded role cache and in the getDefaultRole() method, after getting role from Env.getCtx(), check that the role exists in the loaded role cache and return null if it is not. This way, role will be reloaded after reset of cache.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
CarlosRuiz-globalqss: > If we wants the checking to be more consistent, we can add the size() > 0 check to resetLocalCache(String,int) , not removing the check at resetLocalCache().

implemented your suggestion in https://github.com/idempiere/idempiere/pull/2194/commits/e0e9a1e84e17d225534a69c2e283c72558db3ba4

org.adempiere.base/src/org/compiere/model/MRole.java
CarlosRuiz-globalqss: Thanks @hengsin 

I started implementing your second suggestion (check if the role in context is in the cache), and then I found that it is a very easy task to move the default role to be a cache instead of keeping it in context.

When moved the default role to be a CCache, everything worked without need of special tricks.

org.adempiere.base/src/org/compiere/model/MRole.java
```diff
@@ -115,16 +115,15 @@ else if (defaultRole.getAD_Role_ID() != AD_Role_ID
 	 * @param defaultRole
 	 */
 	private static void setDefaultRole(MRole defaultRole) {
-		Env.getCtx().remove(ROLE_KEY);
-		Env.getCtx().put(ROLE_KEY, defaultRole);
+		s_defaultrole.put(ROLE_KEY, defaultRole);
```
hengsin, line 118: hi @CarlosRuiz-globalqss , this is not correct, that make it the default role for everyone (i.e global), not just the current user session. Like I've suggested earlier, you can check the existence of role in cache to determine whether you need to reload it but you can't directly get/put from it.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
```diff
@@ -307,7 +307,7 @@ protected int resetLocalCache (String tableName, int Record_ID)
 		CacheInterface[] instances = getInstancesAsArray();
 		for (CacheInterface stored : instances)
 		{
-			if (stored != null && stored instanceof CCache)
+			if (stored != null && stored.size() > 0)
```
hengsin, line 310: To avoid a potential classcast exception below, it is safer to keep the instanceof check as well.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MRole.java
CarlosRuiz-globalqss: Thanks @hengsin - implemented your suggestion, tested and it works perfect.

org.adempiere.base/src/org/compiere/model/MRole.java
CarlosRuiz-globalqss: Thanks @hengsin - implemented your suggestion, tested and it works perfect.

org.adempiere.base/src/org/compiere/util/CacheMgt.java
CarlosRuiz-globalqss: done

hengsin (approved):
looks good

PR #2202, opened 2024-01-22
IDEMPIERE-4819 Private Lock - Allow Unlock other users Private Records

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-4819

# Pull Request Checklist

### Tests


hengsin:
hi @CarlosRuiz-globalqss , while the code looks fine, it is a use case that should be validated by @norbertbede. My only concern is whether this can be perceived as a security back door.

CarlosRuiz-globalqss:
Hi @hengsin - IMO not a backdoor, is a missing functionality.

Users can mistakenly lock records and the tenant admin doesn't have any ability to look for that, is just visible using SQL (just the implementor or IT)

The Where clause in the Tab is to prevent users from seeing records they don't have permission to see or fix.
Also the beforeSave validation is to prevent being used as a backdoor (for example disabling a Private Record from REST web services).

CarlosRuiz-globalqss:
Hi @hengsin 

I documented this feature at:
https://wiki.idempiere.org/en/NF11_Private_Access_Window

If you find this OK I think there is no need to wait for @norbertbede approval.

This is not intended to solve Norbert's issue but to solve a missing feature on iDempiere.

Is it OK to move it ahead?

norbertbede:
hi, i'm quite busy, but i can take a look today. if you are sure, no need to wait for me. n

hengsin:
hi @CarlosRuiz-globalqss ,

I think we need to add 2 things here:
- notify user that their "Privately Locked" record have been unlocked by other
- need to let user know that their "Privately Locked" can be unlocked by other with "Personal Access" privilege. Perhaps a message or some documentation update.

CarlosRuiz-globalqss:
Hi @hengsin,

>     * notify user that their "Privately Locked" record have been unlocked by other

I don't agree that's needed.

This is usually a tenant admin person or an IT person solving a problem created by a non-admin user.  Just that at this moment is done via SQL and now they'll be able to do it directly in the app.

Usually actions from an administrator don't need to be notified to the non-admin users.

For example the use case described by @norbertbede doesn't apply:
"when user created privatelock (role has enabled private lock feature) of a record, only the same user able to unlock it. If user leave company or on holiday or in our case users makes lock in e-commerce portal -  nobody able to unlock it"

>     * need to let user know that their "Privately Locked" record can be unlocked by other with "Personal Access" privilege. Perhaps a message or some documentation update.

I'm not sure about this need.

The fields Personal Lock and Personal Access are documented, maybe the phrasing of the Personal Access help could be changed to:
"Users of this role have access to all records locked as personal, **and they can unlock records if they have access to the Private Access window**"


Anyways, if this is not good for core is very easy to make it a plugin instead.

hengsin (approved):
looks good

PR #2219, opened 2024-01-30
IDEMPIERE-6016:test case to simulate transaction timeouts and ensure no open connections remain afterwards

hieplq:
To observe the connection leak, please revert commit 6acc6f31 before running this test case.

# Pull Request Checklist

### Tests
### Documentation


org.idempiere.test/src/org/idempiere/test/base/DBTest.java
```diff
@@ -373,4 +377,47 @@ else if (DB.isOracle())
...
+			Thread.sleep(40000);//Wait for the transaction monitor to complete its task
+			
+			int afterActiveConnection = getNumConnectPerStatus(DB.getDatabase().getStatus(), REG_ACTIVE_CONNECT);
+			assertEquals(beforeActiveConnection, afterActiveConnection);
+		} catch (InterruptedException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
+	}
+
```
hengsin, line 422: this needs to be executed in isolation from other test. to ensure that, add `@Isolated` and `@Execution(ExecutionMode.SAME_THREAD)` annotation to the DBTest class.

hengsin:
[pr2219.patch.txt](https://github.com/idempiere/idempiere/files/14110798/pr2219.patch.txt)
- replace abort with rollback
- use shorter duration for timeout test 2
- add Isolatead and same thread execution class annotation

hengsin (changes requested):
hi @hieplq , please review attached patch, thanks.

hieplq:
>     * replace abort with rollback

Modifying this code, will the timeout now fire the AfterRollbackEvent instead of its original behavior?

hengsin:
> > ```
> > * replace abort with rollback
> > ```
> 
> Modifying this code, will the timeout now fire the AfterRollbackEvent instead of its original behavior?

No, to fire AfterRollbackEvent, you need to call Trx.rollback instead of Connection.rollback.

hieplq:
we need to call setAutoCommit (true) also?

hengsin:
> we need to call setAutoCommit (true) also?

You means to duplicate what's in Trx.close ? Well, perhaps it is safer to do that as well. I guess we can try to just call the rollback and close method of Trx instead.

hieplq:
> You means to duplicate what's in Trx.close ? 

yeb

> Well, perhaps it is safer to do that as well. I guess we can try to just call the rollback and close method of Trx instead.

This triggers the event, so timeouts case might require overloaded rollback/close methods to no raise event

hengsin:
hi @hieplq, I don't think any triggering of trx event (rollback, close) is a problem here.  Also, this is suppose to happens rarely and don't think we should put too much design effect into this.

hengsin:
hi @hieplq , I try with simplifying that to use existing trx.rollback and trx.close method and it seems work just fine.

[pr2219.patch1.txt](https://github.com/idempiere/idempiere/files/14174190/pr2219.patch1.txt)

hieplq:
it's ok for me

hengsin (approved):
tested

PR #2244, opened 2024-02-13
IDEMPIERE-5796: for several tables in a row

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-5796

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
+++ b/org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
@@ -913,6 +913,6 @@ public class ModelClassGenerator
 	 */
 	public static void generateSource(String sourceFolder, String packageName, String entityType, String tableName, String columnEntityType)
 	{
-		ModelInterfaceGenerator.generateSource(ModelInterfaceGenerator.GEN_SOURCE_INTERFACE, sourceFolder, packageName, entityType, tableName, columnEntityType);
+		ModelInterfaceGenerator.generateSource(ModelInterfaceGenerator.GEN_SOURCE_CLASS, sourceFolder, packageName, entityType, tableName, columnEntityType);
 	}
 }
--- a/org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
+++ b/org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
@@ -927,9 +927,7 @@ public class ModelClassGenerator
 		if (tableName == null || tableName.trim().length() == 0)
 			throw new IllegalArgumentException("Must specify table name");
 
-		StringBuilder tableLike = new StringBuilder().append(tableName.trim());
-		if (!tableLike.toString().startsWith("'") || !tableLike.toString().endsWith("'"))
-			tableLike = new StringBuilder("'").append(tableLike).append("'");
+		StringBuilder tableLike = new StringBuilder().append(tableName.trim().toUpperCase().replaceAll("'", ""));
 
 		StringBuilder entityTypeFilter = new StringBuilder();
 		if (entityType != null && entityType.trim().length() > 0)
@@ -982,9 +980,19 @@ public class ModelClassGenerator
 			.append("WHERE IsActive = 'Y' AND TableName NOT LIKE '%_Trl' ");
 		// Autodetect if we need to use IN or LIKE clause - teo_sarca [ 3020640 ]
 		if (tableLike.indexOf(",") == -1)
-			sql.append(" AND TableName LIKE ").append(tableLike);
-		else
-			sql.append(" AND TableName IN (").append(tableLike).append(")"); // only specific tables
+			sql.append(" AND UPPER(TableName) LIKE ").append(DB.TO_STRING(tableLike.toString()));
+		else { // only specific tables
+			StringBuilder finalTableLike = new StringBuilder("");
+			for (String table : tableLike.toString().split(",")) {
+				if (finalTableLike.length() > 0)
+					finalTableLike.append(", ");
+
+				finalTableLike.append(DB.TO_STRING(table.replaceAll("'", "").trim()));
+			}
+
+			sql.append(" AND UPPER(TableName) IN (").append(finalTableLike).append(")");
+		}
+
 		sql.append(" AND ").append(entityTypeFilter.toString());
 		if (filterViews != null) {
 			sql.append(filterViews);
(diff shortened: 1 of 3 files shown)


CarlosRuiz-globalqss:
@nmicoud - this is unnecessary

Same as the swing model generator the "Table Like" parameter accepts several notations, as the Help of the parameters states:

> You can use % or a comma separated list of table names enclosed within single quotes (case sensitive)

nmicoud:
Hi @CarlosRuiz-globalqss ,
In my tests, it is not working .

eg: 
image

doesn't generate anything

nmicoud:
with 'XXA_NdfLine','XXA_NdF'
only the last table (XXA_NdF) is generated :-/

CarlosRuiz-globalqss:
> with 'XXA_NdfLine','XXA_NdF' only the last table (XXA_NdF) is generated :-/

is case sensitive - are you sure the table name is XXA_NdfLine  (with lowercase f?)

CarlosRuiz-globalqss:
it is the same code that is written from the command line - I would just leave it as is - for example I have scripts that run on release time that use this notation, I think it has a lot of years like that

nmicoud:
ok, so maybe adding some code in GenerateModelParameterListener to add those quotes?
haven't tested, but when TableName parameter loses the focus, we can see if it contains comma(s) and check if single quotes are present ; and if not add them?
Or easier, add them on-the-fly on the GenerateModel.java (if TableName parameter contains a comma).
If still not "interested", I'll do to it locally - no bother

CarlosRuiz-globalqss:
I don't see the value added, that's a developer tool, not an end-user, and that change probably will require also changes on other parts, or make it incompatible with the command line tool.

But anyways - you can try to convince @hengsin or @dpansheriya - but IMO there is no value added there   :-(

nmicoud:
Ok, let's wait for their opinion (when I was talking about changes on GenerateModel.java, I was talking about the process - no change require elsewhere).

CarlosRuiz-globalqss:
BTW - at some point I tested if making that parameter ChosenMulti would be good, but it didn't work because you can add LIKE stuff there that are not tablenames

hengsin:
hi @nmicoud ,

The changes should be apply to ModelInterfaceGenerator.generateSource and ModelClassGenerator.generateSource instead.

I think making the following changes should make it more user friendly:
- comparision of table name is case insensitive (that should be straightforward, just upper or lower everything).
- make the use of single quote ("'") optional, i.e `'abc','def'`,`abc,def`,`abc%` and `'abc%'` should work.

nmicoud:
hi @hengsin ,
I've implemented what you suggested.
Does it sounds ok?
Thanks,

Nicolas

hengsin:
> hi @hengsin , I've implemented what you suggested. Does it sounds ok? Thanks,
> 
> Nicolas

Looks good to me.
@CarlosRuiz-globalqss , WDYT ?

org.adempiere.base/src/org/adempiere/util/ModelInterfaceGenerator.java
```diff
@@ -792,9 +792,7 @@ public static void generateSource(String sourceFolder, String packageName, Strin
 		if (tableName == null || tableName.trim().length() == 0)
 			throw new IllegalArgumentException("Must specify table name");
 
-		StringBuilder tableLike = new StringBuilder().append(tableName.trim());
-		if (!tableLike.toString().startsWith("'") || !tableLike.toString().endsWith("'"))
-			tableLike = new StringBuilder("'").append(tableLike).append("'");
+		StringBuilder tableLike = new StringBuilder().append(tableName.trim().toUpperCase());
```
CarlosRuiz-globalqss, line 795: @nmicoud 

Can you please check?

There are differences between the methods generateSource of ModelClassGenerator vs ModelInterfaceGenerator

In theory these two methods must be the same with exception of the line within the while ... new Model

Ideally, because this is duplicated code, it would be better to have it in just one place.

For example, if we can call ModelClassGenerator.generateSource from ModelInterfaceGenerator.

This line here is very relevant, if you don't remove the quotes, then it will fail on line 848 when receiving a quoted tablename

As we are dealing with tablenames, I think is better also to remove spaces too, with the previous approach spaces in the tableLike were allowed, but with this change, spaces will be problematic.

nmicoud: Hi @CarlosRuiz-globalqss ,

Something like that?
Not sure about the name of the new generateSource method with type parameter (which is not yet in javadoc).

And I don't understand what you suggest with spaces ; can you give an example where it can fail?

Thanks,

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
```diff
@@ -914,129 +913,6 @@ public String toString()
...
-		catch (SQLException e)
-		{
-			throw new DBException(e, sql.toString());
-		}
-		finally
-		{
-			DB.close(rs, pstmt);
-			rs = null; pstmt = null;
-		}
+		ModelInterfaceGenerator.generateSource(ModelInterfaceGenerator.GEN_SOURCE_INTERFACE, sourceFolder, packageName, entityType, tableName, columnEntityType);
```
CarlosRuiz-globalqss, line 916: Here probably is GEN_SOURCE_CLASS ?

CarlosRuiz-globalqss (changes requested):

org.adempiere.base/src/org/adempiere/util/ModelInterfaceGenerator.java
CarlosRuiz-globalqss: > And I don't understand what you suggest with spaces ; can you give an example where it can fail?

I haven't tested, but from the code I think before the patch the following will work:
`'AD_Table', 'AD_Column'`   <-  (see the space after the comma)
but after the patch I think it will fail because you are removing the quotes and adding them back, but not removing the spaces

nmicoud: Just made a test and both tables were considered

org.adempiere.base/src/org/adempiere/util/ModelClassGenerator.java
nmicoud: Oops, fixed

CarlosRuiz-globalqss (approved):
tested

PR #2290, opened 2024-04-03
IDEMPIERE-6064 Error when creating a table with a long name through CreateTable process

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6064

# Pull Request Checklist

### Tests


CarlosRuiz-globalqss:
> The only thing I see is the max length of the parameter _TableName_ of the CreateTable process which is still 27 ; it should be updated too.

Ready, size increased in that parameter, I did a deeper review and found more cases that I increased size too.


> BTW, what about adding a maximum size for objects, defined by implementors (to avoid having objects with too long names) ; its default value would be 63 and a SysConfig would allow to reduce it.

Having that can cause problem if the implementor has a shorter standard and try to install a plugin with a longer name.

Regards,

Carlos Ruiz

hengsin (approved):
looks good

PR #2293, opened 2024-04-04
IDEMPIERE-6091 Field shown as mandatory when filled with space of tab

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6091

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
@@ -890,7 +890,7 @@ public abstract class WEditor implements EventListener<Event>, PropertyChangeLis
 	 */
 	public boolean isNullOrEmpty() {
 		Object value = getValue();
-		return value == null || value.toString().trim().length() == 0;
+		return value == null || value.toString().length() == 0;
 	}
 	
 	/**


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WEditor.java
```diff
@@ -890,7 +890,7 @@ public boolean isMandatoryStyle() {
 	 */
 	public boolean isNullOrEmpty() {
 		Object value = getValue();
-		return value == null || value.toString().trim().length() == 0;
+		return value == null || value.toString().length() == 0;
```
hengsin, line 893: This is indeed correct but on the other hand, they are many mandatory fields that it is not alright for user to fill it up with space (imagine a user name with all space or tab), how do we deal with that in iDempiere ?

CarlosRuiz-globalqss: > Semantic wise, this is indeed the correct fix but on the other hand, they are many mandatory fields that it is not alright for user to fill it up with space (imagine a user name with all space or tab), how do we deal with that in iDempiere ?

That's something that users do sporadically to overcome those mandatory restrictions.  In principle is a valid input and the core accepts it - and I think it must accept it by default.

But, probably in a different ticket we could make an improvement to implement regular expression validation on AD_Column.VFormat - at this moment VFormat is very limited, but maybe we can add a new way to do regexp validation of values saved there, it can be for example if the VFormat starts with prefix ~ then it means is a regexp and we validate it on PO.save.  That would make the VFormat really powerful.

PR #2320, opened 2024-04-21
IDEMPIERE-6113 : Date Range boxes on Info Window are too small

uthadehikaru:
https://idempiere.atlassian.net/browse/IDEMPIERE-6113

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
@@ -1985,8 +1985,8 @@ public class InfoWindow extends InfoPanel implements ValueChangeListener, EventL
 		        editor2.fillHorizontal();
 		        if (DisplayType.isDate(mField.getDisplayType())) {
 		        	// give space for the Date Range button
-			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor.getComponent(), "40%");
-			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor2.getComponent(), "40%");
+			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor.getComponent(), "44%");
+			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor2.getComponent(), "44%");
 		        } else {
 			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor.getComponent(), "50%");
 			        ZKUpdateUtil.setWidth((HtmlBasedComponent) editor2.getComponent(), "50%");
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/info/InfoWindow.java
@@ -1797,15 +1797,19 @@ public class InfoWindow extends InfoPanel implements ValueChangeListener, EventL
 		Columns columns = new Columns();
 		parameterGrid.appendChild(columns);
 		noOfParameterColumn = getNoOfParameterColumns();
+		String labelWidth = ( 96 / ( 3 * ( getNoOfParameterColumns() / 2 ) ) ) + "%";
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss (changes requested):
I understand the patch is to make the size of the labels shorter, but that's not solving the problem of the date range being too small.

It kind of creates a different issue too, as for example the label "EMail Address" becomes two lines instead of one.

CarlosRuiz-globalqss:
@uthadehikaru - I tried the following patch (on top of yours):
[pr2320PR.patch.txt](https://github.com/idempiere/idempiere/files/15051910/pr2320PR.patch.txt)

For me it's working fine, can you please confirm?

uthadehikaru:
> @uthadehikaru - I tried the following patch (on top of yours): [pr2320PR.patch.txt](https://github.com/idempiere/idempiere/files/15051910/pr2320PR.patch.txt)
> 
> For me it's working fine, can you please confirm?

hi Carlos. the layout is not look nice for me from UI view. but yes it's solve the problem. i add minor fix to set the vertical align top on row so it make the view better.

[PR2320.patch.txt](https://github.com/idempiere/idempiere/files/15056000/PR2320.patch.txt)

hengsin:
If we are to make code changes, it make more sense to move all/any as the last query parameter. Can't imagine that's more difficult than what your guys has posted above.
Also, as it is, it is possible to improve the situation with some css changes (should move the dash style call to css too):
Screenshot from 2024-04-22 12-05-56
Screenshot from 2024-04-22 12-07-19

hengsin:
Patch to move All/Any to bottom of parameter grid:
[MoveAllAny.patch.txt](https://github.com/idempiere/idempiere/files/15057570/MoveAllAny.patch.txt)

CarlosRuiz-globalqss:
I combined all your ideas and came up with this commit:
https://github.com/idempiere/idempiere/pull/2320/commits/7cdacb44e17a8c0ba332687372a79c31daf64402

It position the All/Any in a fixed position always as last parameter in the same last row if there is space.

I also changed the extra column at the right to use just 4% of the screen, and changed the width for labels and fields to be calculated depending on the number of columns, fields are double size as labels.

I removed the dash in the middle as it seems unnecessary - I think the PlaceHolders are enough to explain is a range.

I also added back buttons for numberbox and datebox.

uthadehikaru:
i checked the latest change. the date range is not show full even after i hide the help panel. the previous layout even not looks good but user can see the date range in full view. there is still space on left and right side that can used to maximize the field. 
image

norbertbede:
that is worst, you lost a row.

hengsin:
> that is worst, you lost a row.

hi @norbertbede , what would you prefer here ? keep all/any as it is (perhaps we can reduce the space allocated to it and shorten the text label) and apply the css changes I posted above ?

CarlosRuiz-globalqss:
> that is worst, you lost a row.

you lose a row when the last row is fully occupied, but the problem here is about trying to get more width - so gaining a column is good for the purpose.

In mobile you have very little width, but you can scroll down to many rows.

hengsin:
hi @CarlosRuiz-globalqss , what about letting the grid to size by content instead:
[2320.patch.txt](https://github.com/idempiere/idempiere/files/15070774/2320.patch.txt)

uthadehikaru:
i tried maximize the possible width for editor. looks good on me
image
[patch.diff.txt](https://github.com/idempiere/idempiere/files/15071671/patch.diff.txt)

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , what about letting the grid to size by content instead: [2320.patch.txt](https://github.com/idempiere/idempiere/files/15070774/2320.patch.txt)

Hi @hengsin, I tried your patch and it looks almost good except:
* in my Business Partner Info it creates an horizontal scrollbar always - my test info has date range and TotalOpenBalance as numeric range
* almost all info windows looks OK except Asset Info, there is a multi-selection Business Partner there, and it creates a mini textbox where content is barely visible

BTW, moving the All/Any to the last row has a very good effect in mobile, before that change the All/Any consumes half of the mobile screen, after that it looks much better.

hengsin (approved):
tested

CarlosRuiz-globalqss (approved):
tested too

PR #2325, opened 2024-04-23
IDEMPIERE-6117:POI: WorkbookFactory.create cannot be used to open XLS…

hieplq:
[IDEMPIERE-6117](https://idempiere.atlassian.net/browse/IDEMPIERE-6117)

# Pull Request Checklist

### Tests
### Documentation

[IDEMPIERE-6117]: https://idempiere.atlassian.net/browse/IDEMPIERE-6117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/org.adempiere.base/src/org/adempiere/base/BaseActivator.java
+++ b/org.adempiere.base/src/org/adempiere/base/BaseActivator.java
@@ -28,11 +28,20 @@ import java.util.Map;
 import java.util.Properties;
 
 import org.adempiere.base.equinox.StackTraceCommand;
+import org.apache.poi.extractor.ExtractorFactory;
+import org.apache.poi.ooxml.extractor.POIXMLExtractorFactory;
+import org.apache.poi.sl.usermodel.SlideShowFactory;
+import org.apache.poi.ss.usermodel.WorkbookFactory;
+import org.apache.poi.xslf.usermodel.XSLFSlideShowFactory;
+import org.apache.poi.xssf.usermodel.XSSFWorkbookFactory;
 import org.compiere.Adempiere;
 import org.eclipse.osgi.framework.console.CommandProvider;
+import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.BundleEvent;
 import org.osgi.framework.Filter;
+import org.osgi.util.tracker.BundleTracker;
 import org.osgi.util.tracker.ServiceTracker;
 
 /**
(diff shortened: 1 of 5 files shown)


org.idempiere.test/pom.xml
```diff
@@ -133,6 +133,54 @@
...
+						<configuration>
+							<!--${url.file.srv}/database/11/Adempiere_pg.jar-->
+							<url>https://raw.githubusercontent.com/hieplq/binary.file/IDEMPIERE-6117/test/poi-ooxml-5.2.3.jar</url>
+							<unpack>false</unpack>
+							<outputDirectory>resources</outputDirectory>
+							<retries>5</retries>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
```
hengsin, line 183: this is unnecessary, just add the test xlsx to test project. it is just 1 file and I don't expect we need to make frequent update of it either.

org.adempiere.base/src/org/adempiere/base/BaseActivator.java
```diff
@@ -61,6 +143,9 @@ public void start(BundleContext context) throws Exception {
 		context.registerService(CommandProvider.class.getName(), new StackTraceCommand(), null);
 		
 		blacklistService = new ComponentBlackListService(context);
+		
+		int trackStates = Bundle.RESOLVED | Bundle.STOPPING | Bundle.ACTIVE | Bundle.STARTING;
+		new PoiBundleTracker(context, trackStates).open();
```
hengsin, line 148: The example at https://github.com/apache/poi/blob/trunk/osgi/src/main/java/org/apache/poi/osgi/Activator.java just add the provider in activator, why do we need to use bundle tracker here ?

org.idempiere.test/pom.xml
hieplq: In addition to the XLSX file, I also have two JAR files for testing case updates and multi-version installation.

org.adempiere.base/src/org/adempiere/base/BaseActivator.java
hieplq: example for case merger POI and POI-OOXML into a single bundle

In Idempiere, I need the Bundle Tracker to ensure that both the wrapper bundle already resolved

org.idempiere.test/pom.xml
hengsin: but that's not what iDempiere core is using, we shouldn't add test here that's for your project work.

org.adempiere.base/src/org/adempiere/base/BaseActivator.java
hengsin: but that's not how it is configure in iDempiere core.

org.idempiere.test/pom.xml
hengsin: Also, for multi version, AFAIU, if you don't limit the import version in MANIFEST.MF, the resolver should always resolve to the latest version.

org.adempiere.base/src/org/adempiere/base/BaseActivator.java
hengsin: Unless there's compelling reason for doing so, I would prefer the simple approach in the link example above and don't have to depends on hard coded bundle name check as well.

hieplq: I'd like a straightforward approach, similar to sample, but need to merged JARs for POI and POI-OOXML for deployment on the idempiere/binary.file

org.idempiere.test/pom.xml
hengsin: The test above doesn't test anything for multi version of POI. You install a different version and unless your bundle specifically import only that version, the OSGi resolver will just resolve to the latest version install.

org.adempiere.base/src/org/adempiere/base/BaseActivator.java
hengsin: If a plugin need a different version of POI, that should be managed in that plugin/bundle, it is not the core job to do that.

CarlosRuiz-globalqss:
replaced by https://github.com/idempiere/idempiere/pull/2357

PR #2329, opened 2024-04-24
IDEMPIERE-6120: Bugged SQL in DunningRunCreate

lHeidbreder:
https://idempiere.atlassian.net/browse/IDEMPIERE-6120

# Pull Request Checklist

- [-] In hard-to-understand areas, I have commented my code.
### Tests
### Documentation

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/DunningRunCreate.java
+++ b/org.adempiere.base.process/src/org/compiere/process/DunningRunCreate.java
@@ -181,12 +181,13 @@ public class DunningRunCreate extends SvrProcess
 			if (previousLevels!=null && previousLevels.length>0) {
 				StringBuilder sqlAppend = new StringBuilder();
 				for (MDunningLevel element : previousLevels) {
-					sqlAppend.append(" AND i.C_Invoice_ID IN (SELECT C_Invoice_ID FROM C_DunningRunLine WHERE ");
-					sqlAppend.append("C_DunningRunEntry_ID IN (SELECT C_DunningRunEntry_ID FROM C_DunningRunEntry WHERE ");
-					sqlAppend.append("C_DunningRun_ID IN (SELECT C_DunningRun_ID FROM C_DunningRunEntry WHERE ");
-					sqlAppend.append("C_DunningLevel_ID="); 
+					sqlAppend.append(" AND i.C_Invoice_ID IN (SELECT C_Invoice_ID FROM C_DunningRunLine cd"
+							+ " JOIN C_DunningRunEntry cd2 ON cd.C_DunningRunEntry_ID = cd2.C_DunningRunEntry_ID"
+							+ " JOIN C_DunningRun cd3 ON cd2.C_DunningRun_ID = cd3.C_DunningRun_ID"
+							+ " WHERE cd.Processed <> 'N'"
+							+ " AND cd2.C_DunningLevel_ID = ");
 					sqlAppend.append(element.get_ID ());
-					sqlAppend.append(")) AND Processed<>'N')");
+					sqlAppend.append(")");
 				}
 				sql.append(sqlAppend.toString());
 			}
--- a/org.adempiere.base/src/org/compiere/model/MDunningLevel.java
+++ b/org.adempiere.base/src/org/compiere/model/MDunningLevel.java
@@ -95,7 +95,7 @@ public class MDunningLevel extends X_C_DunningLevel
 		if (!getParent().isCreateLevelsSequentially ())
 			return null;
 		ArrayList<MDunningLevel> list = new ArrayList<MDunningLevel>();
-		String sql = "SELECT * FROM C_DunningLevel WHERE C_Dunning_ID=? AND DaysAfterDue+DaysBetweenDunning<?";
+		String sql = "SELECT * FROM C_DunningLevel WHERE C_Dunning_ID=? AND DaysAfterDue+DaysBetweenDunning<? AND IsActive='Y'";
 		PreparedStatement pstmt = null;
 		ResultSet rs = null;
 		try


org.adempiere.base.process/src/org/compiere/process/DunningRunCreate.java
```diff
@@ -181,12 +181,13 @@ else if (p_C_BP_Group_ID != 0)
...
-					sqlAppend.append("C_DunningRun_ID IN (SELECT C_DunningRun_ID FROM C_DunningRunEntry WHERE ");
-					sqlAppend.append("C_DunningLevel_ID="); 
+					sqlAppend.append(" AND i.C_Invoice_ID IN (SELECT C_Invoice_ID FROM C_DunningRunLine cd"
+							+ " JOIN C_DunningRunEntry cd2 ON cd.C_DunningRunEntry_ID = cd2.C_DunningRunEntry_ID"
+							+ " JOIN C_DunningRun cd3 ON cd2.C_DunningRun_ID = cd3.C_DunningRun_ID"
+							+ " WHERE cd.Processed <> 'N'"
+							+ " AND cd2.C_DunningLevel_ID = ");
 					sqlAppend.append(element.get_ID ());
-					sqlAppend.append(")) AND Processed<>'N')");
+					sqlAppend.append(")");
```
CarlosRuiz-globalqss, line 190: why removing here the condition `C_DunningRunLine.Processed <> 'N'` ?

On the other hand, checking the compiere code, there is another fix on the line 199, it must be changed to:
`+ "WHERE drl.Processed='Y' AND dr2.C_DunningRun_ID=? AND drl.C_Invoice_ID=?"; // ##1 ##2`

CarlosRuiz-globalqss (changes requested):

org.adempiere.base.process/src/org/compiere/process/DunningRunCreate.java
lHeidbreder: You will find `Processed<>'N'` two lines earlier in `WHERE cd.Processed <> 'N'`, so I didn't remove it. `cd` refers to `DunningRunLine`, simply because it's the query's lynchpin and it was first.  
If you'd like, I'd gladly give them more expressive names. That does sound like a smart idea.

The other change I'll also gladly add while I'm at it.

lHeidbreder: Also while renaming I just noticed that joining `C_DunningRun` seems to be unnecessary currently - however maybe it could use safeguards like checks for `C_DunningRun.AD_Client_ID`. What are your thoughts on that?

CarlosRuiz-globalqss: I think your conversion is wrong, reviewing the equivalent SQL must look like this:
```sql
AND i.C_Invoice_ID IN (SELECT drl.C_Invoice_ID 
  FROM C_DunningRunLine drl
	JOIN C_DunningRunEntry dre ON (dre.C_DunningRunEntry_ID=drl.C_DunningRunEntry_ID)
	JOIN C_DunningRunEntry dre2 ON (dre.C_DunningRun_ID=dre2.C_DunningRun_ID)
  WHERE dre2.C_DunningLevel_ID = ${c_dunninglevel_id} AND drl.Processed <> 'N'
)
```

There is no C_DunningRun on the initial query, but there is a second C_DunningRunEntry

lHeidbreder: I agree that I'm not converting the original completely, as I'm trying to fix a bug in the original that stems from using `C_DunningRunEntry` twice.

As described here: https://idempiere.atlassian.net/browse/IDEMPIERE-6120  
there is a bug in the old query when you create a dunning run using all levels of a sequential dunning strategy. So unless I'm mistaken in calling this behavior a bug, I don't think my new query is actually wrong.

Obviously I don't know of all use cases but I don't see any where this would be correct

CarlosRuiz-globalqss: I think the double C_DunningRunEntry has a reason, but I don't have data to test it, and Dunning has different modes of setting this up.

So, unless there is a failing case that can be tested, and that doesn't affect the working cases, I would suggest that we keep the query as is (or at least converting it in a way that is compatible).

Would be worthy to set up unit tests with the dunning cases, sequential and non-sequential.

lHeidbreder: I may be able to extract some test data but I can't write unit tests, as the test fragment does not compile for me.

I am certain though that the current query does not work as intended.

CarlosRuiz-globalqss: if you can provide a 2pack with the configuration - or explain in the ticket steps to configure in GardenWorld

mschnbeck: As the part in question is used only when doing sequential dunning, it's sure that it won't affect other dunning methods. Sequential dunning with more than two level was wrong with the current implementation because as soon as there was one invoice with level three or four all invoices which where dunned once then would immediately dunned with level three or four. The decision which level has to be used is not done according to the invoice in question but looking at all invoices ever dunned.

CarlosRuiz-globalqss: please describe the failing test case so we can follow the need of the fix

d-ruiz:
Hi @mschnbeck and @lHeidbreder,

**Quick status check:**
- [ ] Still working on this and plan to complete soon? 
- [ ] Ready to close for now?

mschnbeck:
Hi @d-ruiz , of course this bug should be corrected. But there is nothing we could do other than providing the given pull request. Carlos asked for a description how to show the failure and that's what I did last.

PR #2340, opened 2024-04-26
IDEMPIERE-6123 Query in search window causing slowness and load spikes in the database (FHCA-5356)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/issues/IDEMPIERE-6123

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -647,7 +647,7 @@ public class GridTable extends AbstractTableModel
 			m_buffer = new ArrayList<Object[]>(m_rowCount+10);
 		}
 		m_sort = new ArrayList<MSort>(m_rowCount+10);
-		//actual row count or -1 for timeout
+		// actual row count or timeout
 		if (m_rowCount > 0 || m_rowCountTimeout) 
 		{
 			m_loader.setContext(ServerContext.getCurrentInstance());
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -95,16 +95,15 @@ public class GridTable extends AbstractTableModel
 	implements Serializable
 {
 	/**
-	 * generated serial id
+	 * 
 	 */
-	private static final long serialVersionUID = -5564364545827057092L;
+	private static final long serialVersionUID = -2602189278069194311L;
 
 	protected static final String SORTED_DSE_EVENT = "Sorted";
 	
-	public static final int DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS = 10;
-	
-	public static final int DEFAULT_COUNT_TIMEOUT_IN_SECONDS = 1;
-	
+	public static final int DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS = 30;
+	public static final int DEFAULT_GRIDTABLE_COUNT_TIMEOUT_IN_SECONDS = 1;
+
 	public static final String LOAD_TIMEOUT_ERROR_MESSAGE = "GridTabLoadTimeoutError";
 
 	public static final String DATA_REFRESH_MESSAGE = "Refreshed";
@@ -3019,7 +3018,8 @@ public class GridTable extends AbstractTableModel
 			{
 				pstmt = DB.prepareStatement(m_SQL_Count, get_TrxName());
 				setParameter (pstmt, true);
-				int timeout = DEFAULT_COUNT_TIMEOUT_IN_SECONDS;
+		        int timeout = MSysConfig.getIntValue(MSysConfig.GRIDTABLE_INITIAL_COUNT_TIMEOUT_IN_SECONDS, 
+		        		DEFAULT_GRIDTABLE_COUNT_TIMEOUT_IN_SECONDS, Env.getAD_Client_ID(Env.getCtx()));
 				if (timeout > 0)
 					pstmt.setQueryTimeout(timeout);
 				rs = pstmt.executeQuery();
@@ -3074,7 +3074,7 @@ public class GridTable extends AbstractTableModel
 				{
 					m_pstmt.setMaxRows(this.maxRows);					
 				}
-				//ensure not all row is fectch into memory for virtual table
+				//ensure not all rows are fetch into memory for virtual table
 				if (m_virtual)
 					m_pstmt.setFetchSize(100);
 				setParameter (m_pstmt, false);
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -101,7 +101,9 @@ public class GridTable extends AbstractTableModel
 
 	protected static final String SORTED_DSE_EVENT = "Sorted";
 	
-	public static final int DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS = 30;
+	public static final int DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS = 10;
+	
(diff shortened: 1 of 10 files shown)


org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -645,7 +647,8 @@ public boolean open (int maxRows)
 			m_buffer = new ArrayList<Object[]>(m_rowCount+10);
 		}
 		m_sort = new ArrayList<MSort>(m_rowCount+10);
-		if (m_rowCount > 0)
+		//actual row count or -1 for timeout
```
hengsin, line 650: I leave a wrong comments here, should be just "//actual row count or timeout"

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
```diff
@@ -916,6 +912,35 @@ public void onEvent(Event event) throws Exception {
...
+	private int getRecordCount(GridTab mTab, StringBuffer where) {
+		StringBuffer sql = new StringBuffer("SELECT COUNT(*) FROM ")
+		        .append(mTab.getTableName());
+		if (where.length() > 0)
+		    sql.append(" WHERE ").append(where);
+		String finalSQL = MRole.getDefault().addAccessSQL(sql.toString(),
+				mTab.getTableName(), MRole.SQL_NOTQUALIFIED, MRole.SQL_RO);
+		int no = -1;
+        int timeout = MSysConfig.getIntValue(MSysConfig.GRIDTABLE_INITIAL_COUNT_TIMEOUT_IN_SECONDS, 
+        		GridTable.DEFAULT_GRIDTABLE_COUNT_TIMEOUT_IN_SECONDS, Env.getAD_Client_ID(Env.getCtx()));
```
hengsin, line 930: The reason I'm using a very short timeout for count here is to make this kind of optional. In other words, the key objective here is thing might still work if record count timeout.

hengsin: The straight way to say this is "if we can get the count in 1 seconds, lets get it, otherwise, we move on". We can increase that 1 seconds but for me, that shouldn't goes higher than 5 seconds.

CarlosRuiz-globalqss: Yes - no problem about that here - in AbstractADWindowContent.getRecordCount happens the initial count when the table is not marked as High Volume, and the short timeout is OK.

When the table is marked as High Volume, the initial count happens on FindWindow.getNoOfRecords, that count is OK to be stopped when opening the window, but then the user fills some parameters and is still interrupted by the short timeout.  In such case the timeout must be longer.  I'll check also what happens with your idea to simply continue without the count ...

org.adempiere.base/src/org/compiere/model/GridTable.java
CarlosRuiz-globalqss: fixed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
hengsin: hi @CarlosRuiz-globalqss , with Findwindow, AbstractADWindowContent.getRecordCount is not call at all. The code that stop the user from leaving the findwindow is the getNoOfRecords call in FindWindow.cmd_ok_Simple and cmd_ok_Advanced. To continue from the timeout error, should be just to ignore that in the 2 FindWindow method above.

CarlosRuiz-globalqss: That's what I tried but ends in an empty window, the loader is loading rows in background but the user cannot interact because a window without records is shown.

hengsin: I try simulate that by replacing line 3024 in Findow with dispose(); and changing the value of m_total to COUNTING_RECORDS_TIMED_OUT (in debug) and the window does still works fine. Is what you are seeing above is just the slowness in getting the first record back from DB ?

hengsin: Found which row count timeout can create a not usable window - the row count timeout inside the  GridTable.Loader.Open method.

hengsin: [pr2340.GridTable.loading.patch.txt](https://github.com/idempiere/idempiere/files/15149971/pr2340.GridTable.loading.patch.txt)
- Fix for the GridTable.Loader.Open issue above.

CarlosRuiz-globalqss: Thanks @hengsin - I'm testing it, it's getting better, but still with some problems on big number of records.

My test case is as follows:
* as role System Administrator
* Table and Column -> AD_AllUsers_V
* in View Component > SQL From > I add the following:
  * `FROM AD_User, generate_series(1,10000000)`
  * I change the 10000000 adding or removing zeroes depending on how slow I want the query to be
* then run "View Validate"
* and then open the window "All Users"

At this moment the server got stuck and I don't see why the timeout for the SQL to load the records is not triggered

CarlosRuiz-globalqss: The test case was incomplete.  Additionally I need to change the view column AD_AllUsers_V_ID to `ROW_NUMBER() OVER ()::numeric(10,0)` and then run the View Validate

CarlosRuiz-globalqss: > [pr2340.GridTable.loading.patch.txt](https://github.com/idempiere/idempiere/files/15149971/pr2340.GridTable.loading.patch.txt)
>     * Fix for the GridTable.Loader.Open issue above.

Thanks @hengsin for the patch, integrated.

I tested thoroughly and added more scenarios.

I think this is practically complete, the only thing that I find annoying is that the UI sometimes react to the DataStatusEvent and show the loading in background, and sometimes it doesn't and just inform the changes when everything is loaded.

hengsin: While testing this, I found out an very "interesting" thing about PostgreSQL query performance with large table. You can get very different result with setFetchSize(0) vs setFetchSize(100) and sometime, surprisingly, things are faster with setFetchSize(0)!

The reason behind this is with setFetchSize(0), PostgreSQL is not using server side cursor and it is with setFetchSize(100). The "interesting" here is PostgreSQL can ends up with a different query execution optimization planner when using server side cursor. You can check this out by comparing "explain SELECT * FROM AD_AllUsers_V WHERE AD_AllUsers_V.AD_Client_ID=0 AND AD_AllUsers_V.AD_Org_ID=0 ORDER BY AD_AllUsers_V.Name FETCH FIRST 100001 ROWS ONLY" with "explain declare ttcur cursor for SELECT * FROM AD_AllUsers_V WHERE AD_AllUsers_V.AD_Client_ID=0 AND AD_AllUsers_V.AD_Org_ID=0 ORDER BY AD_AllUsers_V.Name FETCH FIRST 100001 ROWS ONLY".

hengsin (approved):
tested

PR #2342, opened 2024-04-30
IDEMPIERE-6133 : Support Export Blob Column for Export SQL Insert Scr…

uthadehikaru:
…ipts

https://idempiere.atlassian.net/browse/IDEMPIERE-6133

# Pull Request Checklist

### Tests
### Documentation


org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -3682,14 +3683,7 @@ else if (c == String.class)
...
 					{
-						if (p_info.isColumnMandatory(i))
-						{
-							sqlValues.append("''");		//	no db dependent stuff here -- at this point value is known to be not null
-						}
-						else
-						{
-							sqlValues.append("null");
-						}
+						sqlValues.append("decode('"+Hex.encodeHexString((byte[]) value)+"','hex')");
```
CarlosRuiz-globalqss, line 3686: is this just for Oracle?

hengsin: hi @CarlosRuiz-globalqss , decode(hex encoded data, 'hex') is actually a postgresql only call. need Oracle equivalent implementation here.

uthadehikaru: i check that oracle has hextoraw to convert hex string to blob column but as i see the code. the sql will be same between pgsql and oracle. need a tweak to create different query for each database

hengsin: hi @uthadehikaru , you can do that by add a new method to the AdempiereDatabase interface and implement it in DB_PostgreSQL and DB_Oracle.

uthadehikaru: i see. for temporary, this solution will only affect to postgresql script and for oracle, the value is null or empty.

uthadehikaru: hi @hengsin , i check the Convert class and the database that actively convert is Convert_Postgresql. is idempiere use oracle sql first than when run on postgres, the convert class handle it? i think we can add convert method for blob query for each database.

hengsin: Don't do that, that's the legacy mode and we want to minimize the use of convert for postgresql ( the current default for postgresql is using that at the very minimal level). As I've posted, use the AdempiereDatabase interface instead and it is a very straightforward change.

hengsin (changes requested):
The right way is:
- add `public String TO_Blob(String hexString) `to `AdempiereDatabase` interface.
- that should return `"decode('"+hexString+"','hex')"` for DB_PostgreSQL and return `"hextoraw('"+hexString+"')"` for DB_Oracle.
- Add `String database` argument to `toInsertSQL` and `buildInsertSQL` method
- Inside the `buildInsertSQL` method, you will call `sqlValues.append (DB.getDatabase(database).TO_Blob(Hex.encodeHexString((byte[]) value)));`
- For `addSQLInsert`, `String oracle = Database.getDatabase(Database.DB_ORACLE).convertStatement(po.toInsertSQL(Database.DB_Oracle));` and `String pg = Database.getDatabase(Database.DB_POSTGRESQL).convertStatement(po.toInsertSQL(Database.DB_POSTGRESQL));`

uthadehikaru:
i see. thx @hengsin for your guideline.

org.adempiere.base/META-INF/MANIFEST.MF
```diff
@@ -115,6 +115,7 @@ Import-Package: com.github.benmanes.caffeine;version="3.1.8",
  org.antlr.v4.runtime.tree.pattern;version="4.9.2",
  org.antlr.v4.runtime.tree.xpath;version="4.9.2",
  org.apache.activemq;version="5.3.0",
+ org.apache.commons.codec.binary;version="1.16.1",
```
CarlosRuiz-globalqss, line 118: This is not needed, java 17 has the same:
java.util.HexFormat.of().formatHex((byte[]) value)

org.compiere.db.oracle.provider/src/org/compiere/db/DB_Oracle.java
```diff
@@ -1216,4 +1217,9 @@ public boolean isQueryTimeout(SQLException ex) {
 	public ITablePartitionService getTablePartitionService() {
 		return new TablePartitionService();
 	}
+
+	@Override
+	public String TO_Blob(String hexString) {
```
CarlosRuiz-globalqss, line 1222: I'm not sure if TO_Blob is the best name for this method

@hengsin ?

Also, I think this method requires some sanitization, can be misused if somebody passes a string that is not hexadecimal (f.e. containing quotes)

CarlosRuiz-globalqss (changes requested):

CarlosRuiz-globalqss:
Something to notice:

My tests in oracle shows that this option is not good, it works just with very little files.

I tested with a small image (406 bytes) and the insert was good.

But when testing with a normal image (170KB), the insert generated is useless, it cannot be used in sqlplus or DBeaver, this is an oracle problem that do not allow too long strings, there are several errors trying to do that:
SP2-0027
SP2-0341
ORA-01489
ORA-01704

So, because of that, I would suggest to have a SysConfig to disable exporting BLOBS in oracle.

org.compiere.db.oracle.provider/src/org/compiere/db/DB_Oracle.java
hengsin: I just follow the existing convention like the TO_CHAR method. Probably what we currently have in to_char or to_string. Alternative, is to pass in byte[] instead and perform the conversion of byte[] to hexString inside this method instead.

hengsin:
> Something to notice:
> 
> My tests in oracle shows that this option is not good, it works just with very little files.
> 
> I tested with a small image (406 bytes) and the insert was good.
> 
> But when testing with a normal image (170KB), the insert generated is useless, it cannot be used in sqlplus or DBeaver, this is an oracle problem that do not allow too long strings, there are several errors trying to do that: SP2-0027 SP2-0341 ORA-01489 ORA-01704
> 
> So, because of that, I would suggest to have a SysConfig to disable exporting BLOBS in oracle.

Yes, the size concern is why the original code doesn't export blob.

org.compiere.db.oracle.provider/src/org/compiere/db/DB_Oracle.java
CarlosRuiz-globalqss: OK, understood about the TO_Blob, fine by me.

As you suggest, it sounds safer to receive the byte[] as parameter and do the conversion inside.

org.adempiere.base/META-INF/MANIFEST.MF
hengsin: hi @uthadehikaru, as per Carlos comments above, please use the build in jdk 17 method instead.

org.compiere.db.oracle.provider/src/org/compiere/db/DB_Oracle.java
hengsin: hi @uthadehikaru , as per the discussion above, please change the parameter to byte[] and perform the hex string conversion inside this method.

hengsin (changes requested):
As per @CarlosRuiz-globalqss comments on the size issue, I think we should add a sysconfig to make this optional (or a max size of blob that can be export)

uthadehikaru:
> As per @CarlosRuiz-globalqss comments on the size issue, I think we should add a sysconfig to make this optional (or a max size of blob that can be export)

should we add 2 sysconfigs for both databases? like EXPORT_BLOB_COLUMN_FOR_POSTGRESQL and EXPORT_BLOB_COLUMN_FOR_ORACLE and the default is true? what level this sysconfig? per tenant or system? need suggestion for this.

org.adempiere.base/src/org/compiere/db/AdempiereDatabase.java
```diff
@@ -488,6 +488,13 @@ public default String getSQLCreate(MTable table)
 		return sb.toString();
 	}	//	getSQLCreate
 	
+	/**
+	 * Get SQL Statement to store Blob 
+	 * @param hexString
+	 * @return SQL statement
+	 */
+	public String TO_Blob(String hexString);
```
hengsin, line 496: change "String hexString" to "bytea[] blob"

org.adempiere.base/src/org/compiere/db/AdempiereDatabase.java
```diff
@@ -488,6 +488,13 @@ public default String getSQLCreate(MTable table)
 		return sb.toString();
 	}	//	getSQLCreate
 	
+	/**
+	 * Get SQL Statement to store Blob
```
hengsin, line 492: Replace this with "Convert blob to hex encoded string and return SQL function that will convert the hex encoded string back to blob"

hengsin (changes requested):

CarlosRuiz-globalqss:
> should we add 2 sysconfigs for both databases? like EXPORT_BLOB_COLUMN_FOR_POSTGRESQL and EXPORT_BLOB_COLUMN_FOR_ORACLE and the default is true? what level this sysconfig? per tenant or system? need suggestion for this.

I think just one SysConfig is OK - EXPORT_BLOB_COLUMN_FOR_INSERT
Configurable per tenant (it requires migration script)
default to Yes
I think is enough to have the option to turn it off for oracle installations when needed.

org.adempiere.base/src/org/compiere/db/AdempiereDatabase.java
```diff
@@ -488,6 +488,13 @@ public default String getSQLCreate(MTable table)
 		return sb.toString();
 	}	//	getSQLCreate
 	
+	/**
+	 * Convert blob to hex encoded string and return SQL function that will convert the hex encoded string back to blob
+	 * @param hexString
```
hengsin, line 493: should be `@param blob`

hengsin (changes requested):

CarlosRuiz-globalqss (changes requested):
Hi @uthadehikaru - can you please apply the following patch:

[pr2342PR.patch.txt](https://github.com/idempiere/idempiere/files/15247848/pr2342PR.patch.txt)


Fixes issues I noticed while testing this:

1 - The Test.BinaryData is defined as Image instead of Binary - and it has problems saving records on Test.  This must be some regression bug as I remember fixing that sometime ago, but anyways I think binary is more appropriate.

2 - PO.java is trying to save '' for an empty mandatory BLOB, that's not valid in oracle, oracle treats empty strings as null, saving a '0' fixed the issue

3 - in PO.java too found that if you try to save a Broadcast Message while the preference to Log Migration Script is set, then is not possible to save it, the change in line 5426 solves it


cc: @hengsin

hengsin:
Test.BinaryData doesn't works with Image reference is not a regression. Image reference is for reference to AD_Image and that must be integer instead of blob. If we want to support the use of blob with Image editor, we need to add a new reference for that (for e.g ImageBlob).

hengsin (changes requested):
Please apply the following patch to add support for export of large size Oracle blob data:
[pr2342-oracle-blob.patch.txt](https://github.com/idempiere/idempiere/files/15248373/pr2342-oracle-blob.patch.txt)

hengsin (approved):
tested

hengsin:
hi @CarlosRuiz-globalqss , the latest update seems ok for me, please review again, thanks.

CarlosRuiz-globalqss (approved):
tested successfully in postgresql and oracle

PR #2371, opened 2024-05-18

CarlosRuiz-globalqss:

- avoid calling delete cascade for ID when is UUID table

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -4180,7 +4180,7 @@ public abstract class PO
 					delete_Tree(MTree_Base.TREETYPE_CustomTable);
 				}
 
-				if (m_KeyColumns != null && m_KeyColumns.length == 1) {
+				if (m_KeyColumns != null && m_KeyColumns.length == 1 && !getTable().isUUIDKeyTable()) {
 					//delete cascade only for single key column record
 					PO_Record.deleteModelCascade(p_info.getTableName(), Record_ID, localTrxName);
 					//	Delete Cascade AD_Table_ID/Record_ID (Attachments, ..)


org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -4180,7 +4180,7 @@ else if (savepoint != null)
 					delete_Tree(MTree_Base.TREETYPE_CustomTable);
 				}
 
-				if (m_KeyColumns != null && m_KeyColumns.length == 1) {
+				if (m_KeyColumns != null && m_KeyColumns.length == 1 && !getTable().isUUIDKeyTable()) {
```
hengsin, line 4183: Shouldn't it be something like the following:
```
   if (m_KeyColumns != null && m_KeyColumns.length == 1) {
           if (getTable().isUUIDKeyTable()) {
                    PO_Record.deleteModelCascade(p_info.getTableName(), Record_UU, localTrxName);
                     ...
           } else {
                    PO_Record.deleteModelCascade(p_info.getTableName(), Record_ID, localTrxName);
                    ...
           }
   }
```

CarlosRuiz-globalqss: Hi @hengsin - the lines that manage the UU part start on line 4192

CarlosRuiz-globalqss: ah, I see the "else if" you suggest.

It could be possible that a table has Record_ID against a foreign table, and just Record_UU against another table, so better to check both.
Or in other words, ID based tables can still have Record_UU references (when they have UUID column)

hengsin: That should be very rare. The current code that always execute the Record_UU part is not good for performance.

CarlosRuiz-globalqss: > That should be very rare.

Maybe is rare at this moment because is a new feature, but I think Record_UU is the preferred way over Record_ID because is portable, so it could be that in future, and in customizations will be the preferred way.

> The current code that always execute the Record_UU part is not good for performance.

I think it is necessary to guarantee consistency.

hengsin (approved):
looks good

PR #2379, opened 2024-05-25
IDEMPIERE-6160 Remove Eclipse warnings

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6160


d-ruiz (approved):
I cannot compile iDempiere master in Eclipse because it is showing the following error 

"The project was not built due to "release 17 is not found in the system". Fix the problem, then try refreshing this project and building it since it may be inconsistent"

All problems shown belong to the zk.iceblue package. If I close the project, I can run iDempiere without a problem.

I tested applying this commit and it solved the Problems in Eclipse

PR #2408, opened 2024-07-04
IDEMPIERE-6123 Query in search window causing slowness and load spikes in the database (FHCA-5356)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6123?focusedCommentId=50935

fix new issue found informing timeout errors when no records are found after a failed count


CarlosRuiz-globalqss:
pushing as urgent, as this is an issue found in production r11 environment

PR #2422, opened 2024-07-22
IDEMPIERE-6193: Trial Balance : remove lines where account has no beg…

nmicoud:
…inning balance and movement

https://idempiere.atlassian.net/browse/IDEMPIERE-6193

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/report/TrialBalance.java
+++ b/org.adempiere.base/src/org/compiere/report/TrialBalance.java
@@ -282,6 +282,17 @@ public class TrialBalance extends SvrProcess
 		createBalanceLine();
 		createDetailLines();
 
+		boolean withDel = false;
+		if (withDel) {
+			String sql = "DELETE FROM T_TrialBalance WHERE AD_PInstance_ID = " + getAD_PInstance_ID()
+			+ " AND Account_ID IN (SELECT Account_ID FROM T_TrialBalance WHERE AD_PInstance_ID = " + getAD_PInstance_ID() + " AND AD_Table_ID IS NULL AND AmtAcctBalance = 0)"
+			+ " AND Account_ID IN (SELECT Account_ID FROM T_TrialBalance WHERE AD_PInstance_ID = " + getAD_PInstance_ID() + " GROUP BY Account_ID HAVING COUNT(*) = 1)";
+
+			System.out.println(sql);
+			int no =DB.executeUpdateEx(sql, get_TrxName());
+			System.out.println("del " + no);
+		}
+
 		if (log.isLoggable(Level.FINE)) log.fine((System.currentTimeMillis() - m_start) + " ms");
 		return "";
 	}	//	doIt
--- a/org.adempiere.base/src/org/compiere/report/TrialBalance.java
+++ b/org.adempiere.base/src/org/compiere/report/TrialBalance.java
@@ -282,16 +282,12 @@ public class TrialBalance extends SvrProcess
 		createBalanceLine();
 		createDetailLines();
 
-		boolean withDel = false;
-		if (withDel) {
-			String sql = "DELETE FROM T_TrialBalance WHERE AD_PInstance_ID = " + getAD_PInstance_ID()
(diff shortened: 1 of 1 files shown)


nmicoud:
Hi,
I've created a draft PR as I would like to be sure of the approach.
In the doIt part, you can easily change the value of withDel and see effects.

Also, I'm not sure about the name of the parameter.
The best I could find is IsDelLinesNoBalanceAndMovement -> Delete lines with no beginning balance and movement.

Thanks,

Nicolas

norbertbede:
IncludeAllAccounts or isIncludeAllAccounts ?

nmicoud:
Maybe too generic? but maybe with a good description, that can be enough

CarlosRuiz-globalqss:
Hi @nmicoud,

Here is the patch suggested for the query:
[pr2422_PR0.patch.txt](https://github.com/user-attachments/files/17779106/pr2422_PR0.patch.txt)

I'm wondering if we need a parameter for this, for me it makes sense to have it by default, it doesn't add any value to have a line with zero open balance, and no movement.

I'm also wondering if opening balances with zero value are worthy to have there, even if they have movements (that's something that maybe we can ask Steven).

I would like also to have a parameter that doesn't generate opening lines at all, that would be more useful in some cases where you just want to report the movements.

So, my suggestion:
- do not add a new parameter for this, leave it as default as I think is not adding value to have historical zero accounts there
- we can ask Steven if is worthy to add lines that have movements with a zero beginning balance, if the beginning balance is zero then is unnecessary there, but I don't know if Steven has some usage for those beginning lines
- add a parameter to not add beginning balances at all - if that parameter is set then we simply don't call createBalanceLine method

WDYT?

Regards,

Carlos Ruiz

nmicoud:
Hi @CarlosRuiz-globalqss 

Patch is good 

_>     * do not add a new parameter for this, leave it as default as I think is not adding value to have historical zero accounts there_

Agree to not have a parameter to remove those unnecessary lines. I wanted first to add a parameter to preserve backward compatibility, but indeed, it makes no sense to display them. And if someone complain, we can easily add a SysConfig key later.


_>     * we can ask Steven if is worthy to add lines that have movements with a zero beginning balance, if the beginning balance is zero then is unnecessary there, but I don't know if Steven has some usage for those beginning lines_

I would tend to think that having an opening balance with 0 make sense as it's obvious the beginning balance is 0.
Anyway Steven is the referent so we should wait for his opinion.


_>     * add a parameter to not add beginning balances at all - if that parameter is set then we simply don't call createBalanceLine method_

What would be the interest of having a report without opening lines here?
I mean, it's already there "Accounting Fact Details" :-/ Or reason is to be able to have all from a single menu entry?

Regards,

Nicolas

CarlosRuiz-globalqss:
> Agree to not have a parameter to remove those unnecessary lines. I wanted first to add a parameter to preserve backward compatibility, but indeed, it makes no sense to display them. And if someone complain, we can easily add a SysConfig key later.

OK

> What would be the interest of having a report without opening lines here? I mean, it's already there "Accounting Fact Details" :-/ Or reason is to be able to have all from a single menu entry?

OK, you're right, I was just trying to make the report as it was before, but as you point it is unnecessary for this purpose.

dpansheriya:
> 
> I'm also wondering if opening balances with zero value are worthy to have there, even if they have movements (that's something that maybe we can ask Steven).
> 
As per my experiance working with accountants here, they prefer to have opening balance with 0 when account has movement in given period. So it is good to keep it.

CarlosRuiz-globalqss:
> As per my experiance working with accountants here, they prefer to have opening balance with 0 when account has movement in given period. So it is good to keep it.

Right @dpansheriya - the patch from Nicolas doesn't remove those opening lines when there is movement within the period, it just removes zero opening lines when there is no movement.

CarlosRuiz-globalqss (approved):
Hi @hengsin - I tested this one successfully - do you have any objection or a better idea how to implement it?

At first we thought this required a new parameter, but after testing I think is not needed as I think it doesn't add any value to show a zero being balance with no movements.

hengsin (approved):
looks good

PR #2448, opened 2024-09-06
IDEMPIERE-6223 Add new field JsonData to AD_PInstance - Peer Review Fixes

CarlosRuiz-globalqss:


CarlosRuiz-globalqss:
forced push as is a peer review tested

PR #2459, opened 2024-09-18
IDEMPIERE-4224 : Decluttering configuration using Attribute on record

dpansheriya:
# Pull Request Checklist

### Tests
### Documentation


org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -126,6 +126,23 @@ public abstract class PO
 
 	/** default query/statement timeout, 300 seconds **/
 	private static final int QUERY_TIME_OUT = 300;
+	
+	/** Get value of the attribute for Table ID and Record ID  **/
+	private static final String	TABLE_ATTRIBUTE_VALUE_SQL = "SELECT a.Name, a.AttributeValueType, a.AD_Reference_ID, ta.Value, ta.ValueDate, ta.ValueNumber, ta.M_AttributeValue_ID "
+																		+ "	FROM AD_TableAttribute ta "
+																		+ "	INNER JOIN M_Attribute a ON (a.M_Attribute_ID = ta.M_Attribute_ID) "
+																		+ "	WHERE ta.AD_Table_ID = ? AND Record_ID = ? AND a.IsActive = 'Y' ";
+	/** Get Default value of the attribute **/
```
hengsin, line 135: Not a must, but for new code, it is better to use the """ text block instead of the old string concatenation approach.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -6223,5 +6240,209 @@ public boolean columnExists(String columnName, boolean throwException) {
...
+				DB.close(rs, pstmt);
+				rs = null;
+				pstmt = null;
+			}
+		}
+
+		if (s_tableAttributeDefault.containsKey(key))
+			return s_tableAttributeDefault.get(key);
+		return null;
+	} // getAttributeDefaultValue
```
hengsin, line 6446: Going by the current practise in iDempiere code, I think it is more appropriate to add this new attributes related method to MTableAttribute instead.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/TableAttributePanel.java
```diff
@@ -0,0 +1,373 @@
...
+		{
+			jbInit();
+			dynInit();
+		}
+		catch (Exception e)
+		{
+			log.log(Level.SEVERE, "InfoPAttribute", e);
+		}
+		AEnv.showWindow(this);
+	} // InfoPAttribute
```
hengsin, line 94: copy paster error

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/TableAttributePanel.java
```diff
@@ -0,0 +1,373 @@
...
+
+	private Rows					rows			= null;
+	private ConfirmPanel			confirmPanel	= new ConfirmPanel(true);
+
+	/**
+	 * Layout dialog
+	 * 
+	 * @throws Exception
+	 */
+	private void jbInit() throws Exception
```
hengsin, line 109: understood that this is from copy and paste but it is still weird to have jbInit naming in new code.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/panel/TableAttributePanel.java
```diff
@@ -0,0 +1,373 @@
...
+import org.compiere.util.DisplayType;
+import org.compiere.util.Env;
+import org.compiere.util.Msg;
+import org.compiere.util.Util;
+import org.zkoss.zk.ui.Component;
+import org.zkoss.zk.ui.event.Event;
+import org.zkoss.zk.ui.event.EventListener;
+import org.zkoss.zul.Vbox;
+
+public class TableAttributePanel extends Window implements EventListener<Event>
```
hengsin, line 52: naming and package is kind of misleading, this is clearly a window instead of a panel.

org.adempiere.base/src/org/compiere/model/PO.java
dpansheriya: I think textblocks may not efficient as it uses more space as formatting is considered compared to concatation. As Java compiler normally at compile time treat this concatenations, there is no performance impact of concatenation so I think for sql concatenations is better.

dpansheriya:
@hengsin  Other than using text block, other code review comments are taken care.

org.adempiere.base/src/org/compiere/model/PO.java
hengsin: Performance and space efficiency difference would not be significant here (performance is debatable since string concatenation using + operator is generally consider not efficient), this is for readability and ease of maintenance (that's the key reason why Java added text block feature).

dpansheriya: @hengsin  converted to Text block and committed.

migration/iD12/postgresql/202407231630_IDEMPIERE-4224.sql
```diff
@@ -0,0 +1,483 @@
...
+;
+
+-- Oct 5, 2023, 6:58:10 PM IST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Description,Help,AD_Table_ID,ColumnName,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,IsHtml) VALUES (215945,0,'Record ID','Direct internal record ID','The Record ID is the internal unique identifier of a record. Please note that zooming to the record may not be successful for Orders, Invoices and Shipment/Receipts as sometimes the Sales Order type is not known.',200390,'Record_ID',22,'N','N','N','N','N',0,'N',28,0,0,'Y',TO_TIMESTAMP('2023-10-05 18:58:09','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2023-10-05 18:58:09','YYYY-MM-DD HH24:MI:SS'),100,538,'N','N','D','N','N','N','Y','59a60ff9-fd34-42f0-b51e-01743f493fbb','Y',0,'Y','N','N')
```
CarlosRuiz-globalqss, line 81: This should be of _Record ID_ type instead of button.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java
```diff
@@ -0,0 +1,387 @@
...
+	} // addProductAttributes
+
+	/**
+	 * Get Field editor for attribute
+	 * @param attribute
+	 * @return
+	 */
+	public GridField getGridField(MAttribute attribute)
+	{
+		GridFieldVO vo = GridFieldVO
```
CarlosRuiz-globalqss, line 225: for some reason the Msg.translate here is not working well and the label is not translated.

Maybe this patch is required:
```
diff --git a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java
index 980ef9e079..83b8bff562 100644
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java
@@ -184,7 +184,7 @@ public class WTableAttribute extends Window implements EventListener<Event>
 				if (!Util.isEmpty(attribute.getDescription(), true))
 					label.setValue(attribute.getDescription());
 				else if (Util.isEmpty(label.getValue(), true))
-					label.setValue(attribute.getName());
+					label.setValue(Msg.translate(Env.getCtx(), attribute.get_Translation("Name")));
 
 				row.appendCellChild(label.rightAlign());
 
```

CarlosRuiz-globalqss (changes requested):
Would be good to add a window (probably advanced) that the Admin can use to check and even modify the new AD_TableAttribute table

The current implementation is not supporting UUID based tables (AD_TableAttribute.Record_UU), this can be added later, but better we keep a ticket as reminder.

Please also add NF12 wiki documentation, better to add in the documentation how the Attributes can be translated using Messages.

CarlosRuiz-globalqss:
@dpansheriya - thinking on another use cases, how can a plugin add an attribute for a table?

If a table has an official defined attribute that would be "easy" to add.

But if a table doesn't have officially defined attributes then a plugin would try to add an attribute set and assign it to the table, and another plugin would try the same and both will collide.

Thinking draft here:
What if instead of adding AD_Table.M_AttributeSet_ID, we add a M-N relation between AD_Table and M_AttributeSet
Something like a table AD_TableAttributeSet, so a table can have multiple attribute sets related (from different plugins), and in the WTableAttribute we manage to show all the sets, maybe grouped by AttributeSet name (which we would need to make also translatable using messages maybe).

dpansheriya:
> @dpansheriya - thinking on another use cases, how can a plugin add an attribute for a table?
> 
> If a table has an official defined attribute that would be "easy" to add.
> 
> But if a table doesn't have officially defined attributes then a plugin would try to add an attribute set and assign it to the table, and another plugin would try the same and both will collide.
> 
> Thinking draft here: What if instead of adding AD_Table.M_AttributeSet_ID, we add a M-N relation between AD_Table and M_AttributeSet Something like a table AD_TableAttributeSet, so a table can have multiple attribute sets related (from different plugins), and in the WTableAttribute we manage to show all the sets, maybe grouped by AttributeSet name (which we would need to make also translatable using messages maybe).

To me it sounds good thought but would like to know opinion from @hengsin as it make it little complex

hengsin:
> > @dpansheriya - thinking on another use cases, how can a plugin add an attribute for a table?
> > If a table has an official defined attribute that would be "easy" to add.
> > But if a table doesn't have officially defined attributes then a plugin would try to add an attribute set and assign it to the table, and another plugin would try the same and both will collide.
> > Thinking draft here: What if instead of adding AD_Table.M_AttributeSet_ID, we add a M-N relation between AD_Table and M_AttributeSet Something like a table AD_TableAttributeSet, so a table can have multiple attribute sets related (from different plugins), and in the WTableAttribute we manage to show all the sets, maybe grouped by AttributeSet name (which we would need to make also translatable using messages maybe).
> 
> To me it sounds good thought but would like to know opinion from @hengsin as it make it little complex

Yeah, I think it is better to add a level of grouping.

migration/iD12/postgresql/202407231630_IDEMPIERE-4224.sql
```diff
@@ -0,0 +1,483 @@
...
+;
+
+-- Dec 27, 2023, 2:32:11 PM IST
+INSERT INTO AD_Column (AD_Column_ID,Version,Name,Description,Help,AD_Table_ID,ColumnName,DefaultValue,FieldLength,IsKey,IsParent,IsMandatory,IsTranslated,IsIdentifier,SeqNo,IsEncrypted,AD_Reference_ID,AD_Reference_Value_ID,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Element_ID,IsUpdateable,IsSelectionColumn,ReadOnlyLogic,EntityType,IsSyncDatabase,IsAlwaysUpdateable,IsAutocomplete,IsAllowLogging,AD_Column_UU,IsAllowCopy,SeqNoSelection,IsToolbarButton,IsSecure,FKConstraintType,IsHtml) VALUES (216309,0,'Entity Type','Dictionary Entity Type; Determines ownership and synchronization','The Entity Types "Dictionary", "iDempiere" and "Application" might be automatically synchronized and customizations deleted or overwritten.
```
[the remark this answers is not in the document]
dpansheriya: @CarlosRuiz-globalqss  this U are used in default logic for newly added entity type column.

dpansheriya:
@CarlosRuiz-globalqss in latest push, added a M-N relation between AD_Table and M_AttributeSet along with fixing code review comments

migration/iD12/postgresql/202407231630_IDEMPIERE-4224.sql
```diff
@@ -0,0 +1,483 @@
...
+-- Nov 4, 2023, 9:19:53 AM IST
+UPDATE AD_Column SET FieldLength=2000, AD_Reference_ID=14,Updated=TO_TIMESTAMP('2023-11-04 09:19:53','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=215941
+;
+
+-- Nov 4, 2023, 9:19:58 AM IST
+INSERT INTO t_alter_column values('ad_tableattribute','Value','VARCHAR(2000)',null,'NULL')
+;
+
+-- Dec 27, 2023, 2:31:52 PM IST
+INSERT INTO AD_Wlistbox_Customization (AD_Client_ID,AD_Org_ID,AD_Wlistbox_Customization_ID,WlistboxName,AD_User_ID,Created,CreatedBy,Custom,IsActive,Updated,UpdatedBy,AD_Wlistbox_Customization_UU) VALUES (0,0,200000,'AD_InfoPanel|AD_Element',100,TO_TIMESTAMP('2023-12-27 14:31:52','YYYY-MM-DD HH24:MI:SS'),100,'AD_Element_ID=null,AD_Client_ID=null,AD_Org_ID=null,ColumnName=null,Name=null,PrintName=null,Description=null,Help=null,Placeholder=null,IsActive=null,EntityType=null,PO_Name=null,PO_PrintName=null,PO_Description=null,PO_Help=null','Y',TO_TIMESTAMP('2023-12-27 14:31:52','YYYY-MM-DD HH24:MI:SS'),100,'8a63367f-1acf-4e6f-acbf-c364fab40cd4')
```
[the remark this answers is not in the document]
dpansheriya: This is fixed.

org.adempiere.base/src/org/compiere/model/MTableAttribute.java
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	public MTableAttribute(Properties ctx, ResultSet rs, String trxName)
+	{
+		super(ctx, rs, trxName);
+	}
+
+	public static MTableAttribute get(int tableID, int recordID, int attrID)
+	{
+		String whereClause = MTableAttribute.COLUMNNAME_AD_Table_ID + " = ? AND " + MTableAttribute.COLUMNNAME_Record_ID + " = ? AND " + MTableAttribute.COLUMNNAME_M_Attribute_ID + " = ? ";
```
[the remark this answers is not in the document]
dpansheriya: This is fixed.

org.idempiere.zk.breeze.theme/src/web/theme/default/css/fragment/font-icons.css.dsp
```diff
@@ -410,3 +410,7 @@
 	content: "\f00e";
 	font-weight: 900;
 }
+
+.z-icon-AttributeForm:before {
+	content: "\f29e";
```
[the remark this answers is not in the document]
dpansheriya: This is fixed. image committed in both theme plugin

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/AbstractADWindowContent.java
```diff
@@ -1760,6 +1770,8 @@ private void updateToolbar()
 
 		toolbar.enableQuickForm(adTabbox.getSelectedTabpanel().isEnableQuickFormButton() && !adTabbox.getSelectedGridTab().isReadOnly());
 
+		toolbar.enableAttributeForm((MTable.get(Env.getCtx(), adTabbox.getSelectedGridTab().getAD_Table_ID()).getM_AttributeSet_ID() > 0));
```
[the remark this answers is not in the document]
dpansheriya: This is fixed.

migration/iD12/postgresql/202407231630_IDEMPIERE-4224.sql
dpansheriya: Changed to Record ID type

org.adempiere.base/src/org/compiere/model/MTableAttribute.java
```diff
@@ -0,0 +1,154 @@
...
+					else
+					{
+						value = DefaultValue;
+					}
+					s_tableAttributeDefault.put(key, value);
+				}
+			}
+			catch (Exception e)
+			{
+				CLogger.get().log(Level.SEVERE, "Failed: Get Attribute = " + attributeName, e);
```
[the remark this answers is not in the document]
dpansheriya: This is fixed.

migration/iD12/postgresql/202407231630_IDEMPIERE-4224.sql
```diff
@@ -0,0 +1,483 @@
...
+;
+
+-- Oct 5, 2023, 7:15:34 PM IST
+UPDATE AD_Column SET IsUpdateable='Y', IsAlwaysUpdateable='Y',Updated=TO_TIMESTAMP('2023-10-05 19:15:34','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_Column_ID=215948
+;
+
+-- Oct 25, 2023, 3:32:17 PM IST
+INSERT INTO AD_ToolBarButton (AD_Client_ID,AD_Org_ID,Created,CreatedBy,ComponentName,IsActive,AD_ToolBarButton_ID,Name,Updated,UpdatedBy,IsCustomization,KeyStroke_KeyCode,KeyStroke_Modifiers,AD_ToolBarButton_UU,"action",IsAdvancedButton,IsAddSeparator,EntityType,IsShowMore) VALUES (0,0,TO_TIMESTAMP('2023-10-25 15:32:16','YYYY-MM-DD HH24:MI:SS'),100,'AttributeForm','Y',200133,'Window - AttributeForm',TO_TIMESTAMP('2023-10-25 15:32:16','YYYY-MM-DD HH24:MI:SS'),100,'N',0,0,'563fb0ac-a92d-482e-b6b3-dd3f9970cd96','W','N','N','D','N')
```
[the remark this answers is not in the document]
dpansheriya: As now visible when configuration attribute is set is set on table, so now keeping it in main toolbar may correct.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTableAttribute.java
dpansheriya: This patch was applied.

org.adempiere.base/src/org/compiere/model/X_AD_Table.java
```diff
@@ -31,7 +31,7 @@ public class X_AD_Table extends PO implements I_AD_Table, I_Persistent
 	/**
 	 *
 	 */
-	private static final long serialVersionUID = 20241222L;
+	private static final long serialVersionUID = 20250101L;
```
CarlosRuiz-globalqss, line 34: this change is not required

dpansheriya: fixed

CarlosRuiz-globalqss (changes requested):
Hi @dpansheriya 

Can you please apply the following patch:
[pr2459PR1.patch.txt](https://github.com/user-attachments/files/19039138/pr2459PR1.patch.txt)

Changes:
* Set AccessLevel to All instead of System+Client for attribute tables
  * this is because before the tables had Client+Org, so we would be removing the Org access
* minor remove of extra commands not necessary in migration scripts
* move the new Table Attribute tab to the window "Table and Column" instead of having its own window
* add a missing message
* regenerate model classes
* add header to MTableAttributeSet
* rename the methods getAttribute* in PO to get_TableAttribute*
  * this is because the PO methods must have the underscore prefix to avoid colliding with M class methods
  * and added the Table prefix to avoid confusion because there is already get_Attribute from IDEMPIERE-1231

dpansheriya:
> Hi @dpansheriya
> 
> Can you please apply the following patch: [pr2459PR1.patch.txt](https://github.com/user-attachments/files/19039138/pr2459PR1.patch.txt)
> 
> Changes:
> 
> * Set AccessLevel to All instead of System+Client for attribute tables
>   
>   * this is because before the tables had Client+Org, so we would be removing the Org access
> * minor remove of extra commands not necessary in migration scripts
> * move the new Table Attribute tab to the window "Table and Column" instead of having its own window
> * add a missing message
> * regenerate model classes
> * add header to MTableAttributeSet
> * rename the methods getAttribute* in PO to get_TableAttribute*
>   
>   * this is because the PO methods must have the underscore prefix to avoid colliding with M class methods
>   * and added the Table prefix to avoid confusion because there is already get_Attribute from IDEMPIERE-1231
> * improved methods AsInt and AsBoolean to behave same as the get_Value counterpart
> * changed the default discovery on MTableAttribute to match the same boolean type for get_TableAttribute

Applied above PR

dpansheriya:
> M_AttributeUse

I Tested from Export button on tab as well as from packout. I did not get any stack overflow.

CarlosRuiz-globalqss:
> > M_AttributeUse
> I Tested from Export button on tab as well as from packout. I did not get any stack overflow.

Hi @dpansheriya - probably is not caused by this ticket but a general bug of 2Pack?

This is the test case:
As System, import the following:
[Test4224BP-IsGov.zip](https://github.com/user-attachments/files/19043155/Test4224BP-IsGov.zip)

That test case creates a new attribute for C_BPartner.

Then import this packout definition:
[Test4224-StackOverflow.zip](https://github.com/user-attachments/files/19043156/Test4224-StackOverflow.zip)

When you try to pack out that it throws the StackOverflow

The packout tries to export:
`SELECT * FROM M_AttributeSet WHERE M_AttributeSet_UU='c03980ed-1a6b-4187-b9bd-d373c85344f8';M_AttributeUse`

Maybe is because adding the M_AttributeUse is redundant, but strangely using "Data Single" instead of "Data" also fails.  In theory "Data Single" must not use AttributeUseElementHandler, it must always use GenericPOElementHandler

dpansheriya:
> > > M_AttributeUse
> > > I Tested from Export button on tab as well as from packout. I did not get any stack overflow.
> 
> Hi @dpansheriya - probably is not caused by this ticket but a general bug of 2Pack?
> 
> This is the test case: As System, import the following: [Test4224BP-IsGov.zip](https://github.com/user-attachments/files/19043155/Test4224BP-IsGov.zip)
> 
> That test case creates a new attribute for C_BPartner.
> 
> Then import this packout definition: [Test4224-StackOverflow.zip](https://github.com/user-attachments/files/19043156/Test4224-StackOverflow.zip)
> 
> When you try to pack out that it throws the StackOverflow
> 
> The packout tries to export: `SELECT * FROM M_AttributeSet WHERE M_AttributeSet_UU='c03980ed-1a6b-4187-b9bd-d373c85344f8';M_AttributeUse`
> 
> Maybe is because adding the M_AttributeUse is redundant, but strangely using "Data Single" instead of "Data" also fails. In theory "Data Single" must not use AttributeUseElementHandler, it must always use GenericPOElementHandler

Stack overflow issue fixed, Can you please verify?

CarlosRuiz-globalqss (approved):
tested

PR #2550, opened 2024-11-20
IDEMPIERE-6309 Output of task is better formatted as a console font

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6309

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTask.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTask.java
@@ -123,12 +123,12 @@ public class WTask extends Window implements EventListener<Event>, IHelpContext
 						Executions.activate(desktop, 500);
 						try {
 							StringBuilder sb = new StringBuilder();
-							sb.append("<pre>")
+							sb.append("<pre><code>")
 							.append(osTask.getOut())
-							.append("<br>-----------<br>")
+							.append("-----------<br><p style=\"color:red\">")
 							.append(osTask.getErr())
-							.append("<br>-----------<br>")
-							.append("</pre>");
+							.append("</p>-----------<br>")
+							.append("</code></pre>");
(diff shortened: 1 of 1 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WTask.java
```diff
@@ -123,10 +123,12 @@ public void run() {
...
 						try {
 							StringBuilder sb = new StringBuilder();
-							sb.append(osTask.getOut())
+							sb.append("<pre>")
+							.append(osTask.getOut())
 							.append("<br>-----------<br>")
 							.append(osTask.getErr())
-							.append("<br>-----------");
+							.append("<br>-----------<br>")
+							.append("</pre>");
```
hengsin, line 131: hmm .. how is that ensure the output is using console font ?

hengsin:
The usual web font family for console font: https://www.w3schools.com/css/tryit.asp?filename=trycss_font-family (see .p3)

hengsin:
According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre , monospace is typically use for pre but it is not something guarantee by any means.

CarlosRuiz-globalqss:
> According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre , monospace is typically use for pre but it is not something guarantee by any means.

Hi @hengsin - I tried different options for fonts for the following script:
[test.sh.txt](https://github.com/user-attachments/files/17849954/test.sh.txt)

The best about showing the information in firefox and vivaldi was with \<pre\>\<code\> (this was also the recommendation from chatgpt)

It looks like this:

<kbd>test</kbd>

Other combinations of fonts didn't show the format as expected.

hengsin (approved):
looks good

PR #2560, opened 2024-11-25
IDEMPIERE-6316: Wrong Name on RecordInfo panel when the value of the …

nmicoud:
…column is Context

https://idempiere.atlassian.net/browse/IDEMPIERE-6316

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordInfo.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordInfo.java
@@ -177,6 +177,7 @@ public class WRecordInfo extends Window implements EventListener<Event>
 	private Component tabPanels;
 	private UserPreference userPreference;
 	private Tabbox tabbox;
+	private int windowNo;
 
 	/**
 	 * 	Layout dialog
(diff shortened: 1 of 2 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordInfo.java
```diff
@@ -481,7 +481,7 @@ private void addLine (int AD_Column_ID, Timestamp Updated, int UpdatedBy,
 		Vector<String> line = new Vector<String>();
 		//	Column
 		MColumn column = MColumn.get (Env.getCtx(), AD_Column_ID);
-		line.add(Msg.translate(Env.getCtx(), column.getColumnName()));
+		line.add(Msg.getElement(Env.getCtx(), column.getColumnName()));
```
hengsin, line 484: I think we should replace other `translate` in this class with `getElement` call as well.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/window/WRecordInfo.java
nmicoud: done

CarlosRuiz-globalqss:
Hi @nmicoud and @hengsin 

I have seen this problem before in other parts of the code.

What if change the translate method to look for element when the message is not found?

hengsin:
Msg.translate check context, follow by ad_message and finally ad_element. I think this is wrong for this particular use case since translation of column name should comes from ad_element, not ad_message or context. There's another issue need checking here - the IsSOTrx property. The current getElement call means IsSOTrx is hard coded to true and I think that's not right for some window (for e.g the Purchase Order window).

CarlosRuiz-globalqss:
Hi, the following patch solves the problem of IsSOTrx mentioned by @hengsin :
[pr2560PR0.patch.txt](https://github.com/user-attachments/files/17918036/pr2560PR0.patch.txt)

I added a comment in the ticket for further discussion as this problem seems to be bigger than what is fixed here:
https://idempiere.atlassian.net/browse/IDEMPIERE-6316?focusedCommentId=51315

hengsin (approved):
looks good

PR #2567, opened 2024-11-28
IDEMPIERE-6314 Issues with attachment on System records

CarlosRuiz-globalqss:
- IDEMPIERE-5567 fix for the PO.is_new method


CarlosRuiz-globalqss:
forced push as this is an urgent fix

PR #2568, opened 2024-11-29
Idempiere 6289 missing icon on tab list

uthadehikaru:
https://idempiere.atlassian.net/browse/IDEMPIERE-6289

# Pull Request Checklist

### Tests
### Documentation


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
```diff
@@ -299,13 +299,13 @@ else if(Window.REPLACE.equals(window.getAttribute(Window.INSERT_POSITION_KEY)))
...
 			if (refTab == null)
-				windowContainer.addWindow(tabPanel, title, true, null);
+				windowContainer.addWindow(tabPanel, title, true, new DecorateInfo("Window"));
 			else
 				windowContainer.replace(refTab, window, title);
 
 		}
 		else {
-	    	windowContainer.addWindow(tabPanel, title, true, null);
+	    	windowContainer.addWindow(tabPanel, title, true, new DecorateInfo("Window"));
```
hengsin, line 308: hmm .. can we be sure that this "Window" and not "Report", "Process", "Info" or "Form" ?

uthadehikaru: Yeah, I'm not sure how to check what type of window here. Need deeper testing. Just set default window icon to make sure the tab has icon.

hengsin: that could be misleading to always use the "Window" icon.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/desktop/TabbedDesktop.java
uthadehikaru: Hi @hengsin . i Tried to trace back method showEmbedded and i think it's impossible to detect which is the Window child class extends Window class. If we want to set icon properly, we need to pass window attribute for the icon to set correctly. lets take WAcctViewer that trigger this method. when showWindow is called, we cannot information this is custom form. on DPViews, the icon for Account Info is hardcoded on the class. I just take Window as default icon to make sure the tab has icon.

CarlosRuiz-globalqss: Hi @uthadehikaru - as you said "we need to pass window attribute" - I tested the following patch to open the "Workflow Activities" form from the "Activities" dashboard with success:
[pr2568PR0.patch.txt](https://github.com/user-attachments/files/17992158/pr2568PR0.patch.txt)

I think the same can be done for the other types.

NOTE aside: It would be good to define constants somewhere for the "Window", "Report", "Process", "Info" and "Form" - and use the constants throughout the system instead of those hardcoded values

uthadehikaru: Hi @CarlosRuiz-globalqss , that is what i mean. we need to send attribute like patch you sent. but it's not answer @hengsin concern, how can be sure what type of the window if no attribute defined. can we use window icon or something else? or just leave it empty?

hengsin: Regardless, we should integrate the patch from Carlos above. For existing case where we can't determine the type of the window, we can use a default icon but that shouldn't be the "Window" icon since that would be misleading. My preference would be to just leave it empty.

uthadehikaru: Can we define default icon for window that opened from this method? I tried to traceback and my conclusion is window that opened here is mostly from custom form. we can set DecorateInfo on All custom form of idempiere core class but custom form from plugin will be empty. I suggest this icon as default icon
https://static.vecteezy.com/system/resources/previews/000/575/451/original/windows-tab-icon-vector.jpg
WDYT

hengsin: Yes, I'm ok with that.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/util/Icon.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the GNU General Public License for more details.                       *
+ * You should have received a copy of the GNU General Public License along    *
+ * with this program; if not, write to the Free Software Foundation, Inc.,    *
+ * 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA.                     *
+ *****************************************************************************/
+package org.adempiere.webui.util;
+
+import java.io.Serializable;
+
+public class Icon implements Serializable {
```
hengsin, line 17: It is a class with only constant, should be final (i.e `public final class`) and shouldn't implement Serializable.

Also, should have a private constructor (`private Icon() {}`) since you would never wants to create a new instance of this class.

org.idempiere.zk.iceblue_c.theme/src/web/theme/iceblue_c/css/fragment/font-icons.css.dsp
```diff
@@ -358,6 +358,9 @@
 .z-icon-StepForward:before {
 	content: "\f04e";
 }
+.z-icon-Tab::before {
+	content: "\f24d";
+}
```
hengsin, line 363: Since we still include the breeze theme, please add this to the breeze theme too.
Also, icon image needs to be added for use case where user is using image instead of font icon.

hengsin: hi @uthadehikaru , still missing the image file. If you can't find a suitable one, you can create one from the font icon.

hengsin (approved):
looks good

PR #2573, opened 2024-12-01
IDEMPIERE-6328 : Improve Feedback UI

uthadehikaru:
https://idempiere.atlassian.net/browse/IDEMPIERE-6328

# Pull Request Checklist

### Tests
### Documentation


org.adempiere.ui.zk/js/ckeditor/config.js
```diff
@@ -1,21 +1,11 @@
...
-	'/',
-	{ name: 'basicstyles', items : [ 'Bold','Italic','Underline','Strike','Subscript','Superscript','-','RemoveFormat' ] },
-	{ name: 'paragraph', items : [ 'NumberedList','BulletedList','-','Outdent','Indent','-','Blockquote','-','JustifyLeft','JustifyCenter','JustifyRight','JustifyBlock' ] },
-	{ name: 'links', items : [ 'Link','Unlink' ] },
-	{ name: 'insert', items : [ 'Image','Table','HorizontalRule','SpecialChar' ] }
-];
+    config.toolbar = [
+		{ name: 'styles', items : [ 'Styles','Format','Font','FontSize' ] },
+		{ name: 'colors', items : [ 'TextColor','BGColor' ] },
+	];
```
hengsin, line 10: this change is not right. the idea is config-min.js is for simple/minimum toolbar and config.js for the full toolbar.

hengsin (changes requested):

org.adempiere.ui.zk/js/ckeditor/config.js
uthadehikaru: do we have to show all toolbar buttton? The current toolbar button is too much and not neccessary for basic user when writing email message.

hengsin: I means if we were to show a basic toolbar, we should just use the config-min.js instead of removing config-min.js and modify config.js. That would effect other parts of the application.

uthadehikaru: I'm sorry. I mischeck when searching which use ckeditor/config.js.

hengsin:
hi @uthadehikaru , isn't the original idea is to always use the simplify toolbar (config-min.js) for the feedback email dialog ?

uthadehikaru:
I think it's better to create separated config for email dialog. WDYT?

hengsin:
> I think it's better to create separated config for email dialog. WDYT?

If you feel both existing configuration (config-min.js and config.js) doesn't meet your need, that's fine for me too.

CarlosRuiz-globalqss:
Please consider my comments at
https://idempiere.atlassian.net/browse/IDEMPIERE-6328?focusedCommentId=51336

hengsin:
hi @uthadehikaru , please fix the "never used" Eclipse warnings.

Also, please consider the following patch which should make the email dialog more usable on mobile devices:
[smallscreen.patch.txt](https://github.com/user-attachments/files/17986969/smallscreen.patch.txt)

hengsin (approved):
tested

CarlosRuiz-globalqss (changes requested):
Hi @uthadehikaru 

Can you please consider adding the following patch:
<strike>pr2573PR0.patch.txt</strike>

it shows the "Processing..." dialog while waiting for the screenshot to be collected.


There are still a couple of things here that are not usable in mobile:

- The Mail Dialog doesn't have a scrollbar, the mail text cannot be accessed in mobile on landscape mode, if we can add a scrollbar to this dialog would be great.  The Feedback Create New Request do have a scrollbar and is usable there.
- but also strangely the create request have maximizable button, but it cannot be resized

hengsin:
For the "the "Processing..." dialog, we need to check whether that would be captured as part of the screen shot. If I'm not mistaken, that's the reason why it wasn't there before this. Also, we might want to update html2canvas to the latest available.

hengsin:
Looks like there's a more actively maintain fork at https://github.com/yorickshan/html2canvas-pro

CarlosRuiz-globalqss:
> For the "the "Processing..." dialog, we need to check whether that would be captured as part of the screen shot. If I'm not mistaken, that's the reason why it wasn't there before this. Also, we might want to update html2canvas to the latest available.

You're right - I deleted my suggested patch, the captured screenshot ends with the processing wheel.

org.adempiere.ui.zk/WEB-INF/src/web/js/html2canvas/zk.wpd
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <package name="html2canvas" language="xul/html">
-	<script src="html2canvas.js" />
+	<script src="html2canvas.min.js" />
```
hengsin, line 3: This is not correct. With Zk naming convention, we have to rename html2canvas.js to html2canvas.src.js (for source) and html2canvas.min.js to html2canvas.js

The convention is fix and can't be changed.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/web/js/html2canvas/zk.wpd
uthadehikaru: ah i see. at first, i wanna use html2canvas-pro but it make the build error. can we use html2canvas-pro? it's better to make different with old html2canvas library.

hengsin: of course, the naming is not important.

hengsin:
hi @uthadehikaru , the following patch uses browser display media api (if available) to capture screenshot:

[browserapi.patch.txt](https://github.com/user-attachments/files/18005221/browserapi.patch.txt)

It fallbacks to existing html2canvas if browser display media api is not available.

Please merge it if tested ok, thanks.

uthadehikaru:
> hi @uthadehikaru , the following patch uses browser display media api (if available) to capture screenshot:
> 
> [browserapi.patch.txt](https://github.com/user-attachments/files/18005221/browserapi.patch.txt)
> 
> It fallbacks to existing html2canvas if browser display media api is not available.
> 
> Please merge it if tested ok, thanks.

how to test fallback

> hi @uthadehikaru , the following patch uses browser display media api (if available) to capture screenshot:
> 
> [browserapi.patch.txt](https://github.com/user-attachments/files/18005221/browserapi.patch.txt)
> 
> It fallbacks to existing html2canvas if browser display media api is not available.
> 
> Please merge it if tested ok, thanks.

I cannot confirm the fallback is working. please test when browser api is not available

CarlosRuiz-globalqss (approved):
Hi @hengsin - I tested this successfully.

I added two commits.

1 - I think the change on lang-addon.xml for iceblue_c is not necessary because there are no changes to the theme - right?

2 - I refactored the DefaultFeedbackService.java to create a method getCaptureScreenshotScript and unify the method for Request and EMail (request was forgotten in previous commit)
I also changed the script to capture the screen using htmlcanvas when the user doesn't allow to capture the screen using media.
One side effect I noticed with the media capture is that the user can capture a different window than iDempiere, but I think that's minor probably.

uthadehikaru:
Hi @hengsin , I already solve mobile issue that carlos addressed

hengsin (approved):
tested

PR #2612, opened 2024-12-27
IDEMPIERE-6348 Change oomph to 13

CarlosRuiz-globalqss:


CarlosRuiz-globalqss:
Hi @hengsin - can you please take a look to this?  I'm not sure if that's the correct way to change the oomph file

utils_dev/oomph/idempiere.setup
```diff
@@ -26,7 +26,7 @@
   <setupTask
       xsi:type="setup:VariableTask"
       name="idempiere.target.platform"
-      defaultValue="idempiere-12-local-mirror"
+      defaultValue="idempiere-13-local-mirror"
```
hengsin, line 29: defaultValue must equals to the target name in https://github.com/idempiere/idempiere/blob/master/org.idempiere.p2.targetplatform/org.idempiere.p2.targetplatform.mirror.target

hengsin (changes requested):

utils_dev/oomph/idempiere.setup
CarlosRuiz-globalqss: done

hengsin (approved):
looks good

PR #2614, opened 2024-12-29
IDEMPIERE-6356 Theme Selector on Login

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6356

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/migration/iD13/oracle/202412292146_IDEMPIERE-6356.sql
+++ b/migration/iD13/oracle/202412292146_IDEMPIERE-6356.sql
@@ -1,0 +1,10 @@
+-- IDEMPIERE-6356 Theme Selector on Login
+SELECT register_migration_script('202412292146_IDEMPIERE-6356.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Dec 29, 2024, 9:46:53 PM CET
+UPDATE AD_SysConfig SET Value='iceblue_c,default',Updated=TO_TIMESTAMP('2024-12-29 21:46:53','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_SysConfig_ID=200021
+;
+
--- a/migration/iD13/postgresql/202412292146_IDEMPIERE-6356.sql
+++ b/migration/iD13/postgresql/202412292146_IDEMPIERE-6356.sql
@@ -1,0 +1,7 @@
+-- IDEMPIERE-6356 Theme Selector on Login
+SELECT register_migration_script('202412292146_IDEMPIERE-6356.sql') FROM dual;
+
+-- Dec 29, 2024, 9:46:53 PM CET
+UPDATE AD_SysConfig SET Value='iceblue_c,default',Updated=TO_TIMESTAMP('2024-12-29 21:46:53','YYYY-MM-DD HH24:MI:SS'),UpdatedBy=100 WHERE AD_SysConfig_ID=200021
+;
+
--- a/org.adempiere.base/src/org/compiere/model/SystemProperties.java
+++ b/org.adempiere.base/src/org/compiere/model/SystemProperties.java
@@ -256,6 +256,14 @@ public class SystemProperties {
 	 * ZK_THEME allows to define the theme to use, it overrides the SysConfig ZK_THEME so it can be used for testing purposes
 	 * @return
 	 */
+	public static String setZkTheme(String theme) {
+		return System.setProperty(ZK_THEME, theme);
+	}
+
+	/**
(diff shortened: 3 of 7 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I have a doubt with this pull request.

Was there a reason why org.idempiere.zk.breeze.theme/src/web/theme/default/preference.zul was setting the zktheme cookie name as "breeze" instead of default?

I'm setting the pull request in draft because I think if approved, we better remove the migration script and keep just this as an option for implementors to configure, not as default, mostly because the "breeze" theme is deprecated and I think we must not encourage its usage.

hengsin:
> Hi @hengsin - I have a doubt with this pull request.
> 
> Was there a reason why org.idempiere.zk.breeze.theme/src/web/theme/default/preference.zul was setting the zktheme cookie name as "breeze" instead of default?
> 
> I'm setting the pull request in draft because I think if approved, we better remove the migration script and keep just this as an option for implementors to configure, not as default, mostly because the "breeze" theme is deprecated and I think we must not encourage its usage.

Don't think it will work if you use "default" instead. As it is, iDempiere theme is implemented as customization of zk theme and the underlying zk theme for "default" is "breeze".

CarlosRuiz-globalqss:
> Don't think it will work if you use "default" instead. As it is, iDempiere theme is implemented as customization of zk theme and the underlying zk theme for "default" is "breeze".

Thanks @hengsin - in test-feature it's working fine, what could be the bad consequences?

I tried also renaming default to breeze, but that seems like a very big change and I think you didn't want to do it because it can broke other things.

I can hardcode that when reading the cookie "breeze" it will set the theme to default, but not sure if that's a good idea instead of what's already working.

CarlosRuiz-globalqss:
Hmm no - it seems is not working 100% - I'll do another check

CarlosRuiz-globalqss:
OK, in last commit I left the preference.zul back to save the zktheme cookie as breeze and controlled the situation at org.adempiere.webui.panel.LoginPanel.initComponents

CarlosRuiz-globalqss:
Hi @hengsin - please check my last commit:

I changed org.adempiere.ui.zk/WEB-INF/zk.xml - I think maybe it was something forgotten from the change to iceblue_c

I also added a comment in org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/theme/ThemeManager
do you think is worthy to change the code to use Themes.hasTheme instead?

Also, I'll keep this PR in draft until the finding from Diego is solved, I'll add a comment in the ticket about that.

hengsin:
> Hi @hengsin - please check my last commit:
> 
> I changed org.adempiere.ui.zk/WEB-INF/zk.xml - I think maybe it was something forgotten from the change to iceblue_c
> 
> I also added a comment in org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/theme/ThemeManager do you think is worthy to change the code to use Themes.hasTheme instead?
> 
> Also, I'll keep this PR in draft until the finding from Diego is solved, I'll add a comment in the ticket about that.

Yes, forgot to change that. Perhaps that should be removed now you want to use cookies only.

CarlosRuiz-globalqss:
> Yes, forgot to change that. Perhaps that should be removed now you want to use cookies only.

Ah OK, removed.

I tried that and still doesn't solve the issue described in the ticket.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/theme/ThemeManager.java
```diff
@@ -79,11 +79,14 @@ public static String getTheme() {
...
+			theme = theme.split(",")[0];
 		if (theme.equals(m_brokenTheme)) {
 			theme = ITheme.ZK_THEME_DEFAULT;
 		} else {
 			if (! theme.equals(m_theme)) {
 				if (! ITheme.ZK_THEME_DEFAULT.equals(theme)) {
+					// NOTE: Here we could use also Themes.hasTheme instead, but be careful that default is registered as breeze
 					// Verify the theme.css.dsp exists in the theme folder
 					String themeCSSURL = getStyleSheet(theme);
 					if (ThemeManager.class.getResource(toClassPathResourcePath(themeCSSURL)) == null) {
```
hengsin, line 92: to implement cookie base user theme selection, we need to change ThemeManager.getTheme to get from cookie (if available) as first priority. We can't use Theme.setTheme to change iDempiere theme, that wouldn't work.

d-ruiz:
Hi @CarlosRuiz-globalqss  ,

**Quick status check:**
- [x] Still working on this and plan to complete soon? 
- [ ] Ready to close for now?

CarlosRuiz-globalqss:
Hi @d-ruiz - this is a draft request - like an initial experiment, I think draft requests maybe can be left there.

PR #2623, opened 2025-01-03
IDEMPIERE-2449 NPE on GridTab.m_DataStatusEvent / change approach from volatile to synchronized

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-2449

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTab.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTab.java
@@ -182,7 +182,7 @@ public class GridTab implements DataStatusListener, Evaluatee, Serializable
     /** A list of event listeners for this component.	*/
     protected EventListenerList m_listenerList = new EventListenerList();
     /** Current Data Status Event						*/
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - can you please check if the approach is valid?

I changed the m_DataStatusEvent variable to volatile 9 years ago, and it seems it has worked fine since that time.  But maybe some changes in java 17 are bringing back this issue and the volatile approach seems to have problems.

I tested changing the volatile to synchronized in the methods, and it seems to be working fine now, but maybe concerning is that it must add some slowness to the app.

hengsin (approved):
looks good

hengsin:
@CarlosRuiz-globalqss , it will add some overheat but I think it is not that significant since this is not static methods. There are other possible approach (like remove m_dataStatusEvent variable and use atomic fo m_last*) but will require larger scale changes and more testing.

PR #2630, opened 2025-01-07
IDEMPIERE-5598 : Add new accounging dimensions for

dpansheriya:
Charge. Asset, Department, Warehouse, Employee, Cost Center, CustomField1 to 4

[ Please copy here the link to the related Jira ticket ]

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/migration/iD13/oracle/202405141359_IDEMPIERE-5598.sql
+++ b/migration/iD13/oracle/202405141359_IDEMPIERE-5598.sql
@@ -1,0 +1,3536 @@
+-- IDEMPIERE-5598: Add new Accounting Dimensions
+SELECT register_migration_script('202405141359_IDEMPIERE-5598.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+-- Jan 22, 2024, 1:59:10 PM IST
+INSERT INTO AD_Ref_List (AD_Ref_List_ID,Name,Description,AD_Reference_ID,Value,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,EntityType,AD_Ref_List_UU) VALUES (200692,'Charge','Additional document charges',181,'CH',0,0,'Y',TO_TIMESTAMP('2024-01-22 13:59:10','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2024-01-22 13:59:10','YYYY-MM-DD HH24:MI:SS'),100,'D','ff70fb90-e8eb-4db1-9239-63d31cad4a74')
+;
+
+-- Jan 22, 2024, 1:59:10 PM IST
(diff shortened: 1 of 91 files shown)


hengsin:
so we are deprecating the use of C_ValidCombination here ?

dpansheriya:
> so we are deprecating the use of C_ValidCombination here ?

Not as part of this ticket, but as we decided, the new dimension we are adding are not added in C_ValidCombination . Better we create another ticket for depricating C_ValidCombination

hengsin:
> > so we are deprecating the use of C_ValidCombination here ?
> 
> Not as part of this ticket, but as we decided, the new dimension we are adding are not added in C_ValidCombination . Better we create another ticket for depricating C_ValidCombination

Ok but doing that is effectively deprecating the use of C_ValidCombination.

dpansheriya:
> C_ValidCombination

C_ValidCombination is used only in GL Distribution only. While all other places the use is to get on Account id.
We are deprecating here the us of it but we needs to create an ticket to drop C_ValidCombination and all column referencing it replace by account_ID means to C_ElementValue table with account type element.
That will be another larger refactoring, so instead of mixing, better we do that as separate ticket.

nmicoud:
ValidCombination is also used as a shortcut on JournalLine (see https://idempiere.atlassian.net/browse/IDEMPIERE-5598?focusedCommentId=51466)

CarlosRuiz-globalqss:
This is not deprecating the combination, because after this ticket it can still be used.

What I understand is that we are limiting the combination to the 15 dimensions that it actually manages.

To make it clear, I think is worthy to add a Field Group on the Journal Line tabs (journal and batch) to separate the dimensions that are managed with combination from the new dimensions that are not.

dpansheriya:
> This is not deprecating the combination, because after this ticket it can still be used.
> 
> What I understand is that we are limiting the combination to the 15 dimensions that it actually manages.
> 
> To make it clear, I think is worthy to add a Field Group on the Journal Line tabs (journal and batch) to separate the dimensions that are managed with combination from the new dimensions that are not.

What should we name this field group?

CarlosRuiz-globalqss:
> What should we name this field group?

Maybe "Additional Dimensions" ?

dpansheriya:
> > What should we name this field group?
> 
> Maybe "Additional Dimensions" ?

This changes done on both GL Journal and GL Journal Batch window.

CarlosRuiz-globalqss:
xref -> https://mattermost.idempiere.org/idempiere/pl/8c4ucsxenbbyze568cxht7uoao

CarlosRuiz-globalqss:
All unit tests passed.

I notice the model classes are not generated with the core version, so here is the patch:
[PR2630-model.patch.txt](https://github.com/user-attachments/files/19126080/PR2630-model.patch.txt)

I installed this new feature in https://test-feature.idempiere.org:3243/webui/ for calling community tests

dpansheriya:
> All unit tests passed.
> 
> I notice the model classes are not generated with the core version, so here is the patch: [PR2630-model.patch.txt](https://github.com/user-attachments/files/19126080/PR2630-model.patch.txt)
> 
> I installed this new feature in https://test-feature.idempiere.org:3243/webui/ for calling community tests

Patch applied

nmicoud:
Hi,
I've made some tests and wondering why C_Employee_ID is not (at least) on C_Invoice/C_InvoiceLine (and I guess on C_Payment too)?
IMO, that makes sense if you want to register an invoice from your phone supplier and have 1 line per employee.
wdyt?

dpansheriya:
> wondering

@nmicoud  I agree, I think good to add C_Employee_ID in reference bart of C_InvoiceLine, C_Payment as it needs to record when we are reembersing expense or paying salary using invoice.

dpansheriya:
> Hi, I've made some tests and wondering why C_Employee_ID is not (at least) on C_Invoice/C_InvoiceLine (and I guess on C_Payment too)? IMO, that makes sense if you want to register an invoice from your phone supplier and have 1 line per employee. wdyt?

@CarlosRuiz-globalqss  @nmicoud  I pushed migration script to add C_Employee_ID into invoice and payemnt window.

PR #2641, opened 2025-01-14
IDEMPIERE-5651 - Fix vFormat in field and findWindow

joseleite550:
https://idempiere.atlassian.net/browse/IDEMPIERE-5651

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
@@ -218,13 +218,21 @@ public class WStringEditor extends WEditor implements ContextMenuListener
 	        }
 
 	        // Validate VFormat with regular expression
-	        if (!Util.isEmpty(vFormat) && vFormat.startsWith("~")) {
-	        	String regex = gridField.getVFormat().substring(1); // remove the initial ~
-	        	if (!newValue.matches(regex)) {
-	        		String msgregex = Msg.getMsg(Env.getCtx(), regex);
-	        		throw new WrongValueException(component, Msg.getMsg(Env.getCtx(), "InvalidFormatRegExp", new Object[] {msgregex}));
-	        	}
-	        }
+			if (!Util.isEmpty(vFormat)) {
+				if (vFormat.startsWith("~")) {
+					String regex = gridField.getVFormat().substring(1); // remove the initial ~
+					if (!newValue.matches(regex)) {
+						String msgregex = Msg.getMsg(Env.getCtx(), regex);
+						throw new WrongValueException(component,
+								Msg.getMsg(Env.getCtx(), "InvalidFormatRegExp", new Object[] { msgregex }));
+					}
+				} else {
+					String regex = vFormatToRegex(vFormat);
+					if (!newValue.matches(regex)) {
+						newValue = null;
+					}
+				}
+			}
 
 	        ValueChangeEvent changeEvent = new ValueChangeEvent(this, this.getColumnName(), oldValue, newValue);
 	        
@@ -380,5 +388,42 @@ public class WStringEditor extends WEditor implements ContextMenuListener
 		}
 		return null;
 	}
+	
+	private  String vFormatToRegex(String vFormat) {
+        StringBuilder regex = new StringBuilder();
+        for (char c : vFormat.toCharArray()) {
+            switch (c) {
+                case '0':
+                    regex.append("[0-9]");
+                    break;
+                case '9':
+                    regex.append("[ 0-9]");
+                    break;
+                case 'a':
+                    regex.append("[A-Za-z0-9]");
+                    break;
+                case 'A':
+                    regex.append("[A-Z0-9]");
+                    break;
+                case 'l':
+                    regex.append("[A-Za-z]");
+                    break;
+                case 'L':
+                    regex.append("[A-Z]");
+                    break;
+                case 'U':
+                    regex.append("[^a-z]");
+                    break;
+                default:
+                    if ("()[]{}.+*?^$|\\".indexOf(c) != -1) {
+                        regex.append("\\").append(c);
+                    } else {
+                        regex.append(c);
+                    }
+                    break;
+            }
+        }
+        return regex.toString();
+    }
 
 }
--- a/org.adempiere.ui.zk/WEB-INF/src/web/js/jquery/maskedinput/jquery.maskedinput.js
+++ b/org.adempiere.ui.zk/WEB-INF/src/web/js/jquery/maskedinput/jquery.maskedinput.js
@@ -142,7 +142,10 @@
                     for (i = start; end > i && len > i; i++) tests[i] && (buffer[i] = getPlaceholder(i));
                 }
                 function writeBuffer() {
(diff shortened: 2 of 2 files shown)


dpansheriya:
My thought is on lookup dialog, there should not be any VFormat apply. Applying vformat in lookup dialog not let user type % and so like search can not done. @hengsin @CarlosRuiz-globalqss  please confirm are you agree?

CarlosRuiz-globalqss:
> My thought is on lookup dialog, there should not be any VFormat apply. Applying vformat in lookup dialog not let user type % and so like search can not done. @hengsin @CarlosRuiz-globalqss please confirm are you agree?

Agree with @dpansheriya - I think the correct fix is to disable the VFormat on the fields on Simple and Advanced search.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
```diff
@@ -380,5 +388,42 @@ private AbstractADWindowContent findADWindowContent() {
 		}
 		return null;
 	}
+	
+	private  String vFormatToRegex(String vFormat) {
```
CarlosRuiz-globalqss, line 392: I like the idea of validating also the Format, but this function looks incomplete against the definitions on jquery.maskedinput.js lines 13 to 24

Also, I have not tested if the function copes with constants.

There is another potential big improvement we have seen here: as the VFormat is defined in Column, this validation could happen also on PO, in this way we can define a VFormat and force it to be respected in data coming from Web Services (f.e. REST/SOAP)

Also, I just noticed the help of VFormat element is missing the 'U' notation.  Sorry that I added that in IDEMPIERE-2020 and forgot to change the Help, but we can use this ticket to fix it).

org.adempiere.ui.zk/WEB-INF/src/web/js/jquery/maskedinput/jquery.maskedinput.js
```diff
@@ -142,7 +142,10 @@
                     for (i = start; end > i && len > i; i++) tests[i] && (buffer[i] = getPlaceholder(i));
                 }
                 function writeBuffer() {
-                    input.val(buffer.join(""));
+					if(defaultBuffer == buffer.join(""))
```
CarlosRuiz-globalqss, line 145: Why is this required?  Can you please add a test case where we can find the fix here?

CarlosRuiz-globalqss (changes requested):
I tested the scenario depicted in the description on the ticket, but the pull request is not solving it, as advised better the fix must be to remove the VFormat validation when the field is on Find dialog.

org.adempiere.ui.zk/WEB-INF/src/web/js/jquery/maskedinput/jquery.maskedinput.js
joseleite550: If you keep this part as it is, when you click on the field and then remove its focus, the visual format (vformat) is applied. Later, this format is used for the search, but no corresponding value is found since only the mask is taken into account.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
joseleite550: I added the missing validations from lines 13 to 24

CarlosRuiz-globalqss:
@joseleite550 - as advised - we think is better to implement a different solution here, so I think this ticket can be split maybe in three parts or tickets:

1) do not apply the VFormat on find fields

2) validate also legacy VFormat same as regex on WStringEditor

3) validate VFormat in PO (model layer) - this would enable applying VFormat validation also in REST or other sources

The changes here for WStringEditor are covering the 2)

The changes on jquery.maskedinput.js doesn't seem necessary if the implementation must be the 1)

What do you prefer?  to close this and tackle those three tickets in a different pull request?  Or change this PR to cover just the second case?

joseleite550:
> @joseleite550 - as advised - we think is better to implement a different solution here, so I think this ticket can be split maybe in three parts or tickets:
> 
> 1. do not apply the VFormat on find fields
> 2. validate also legacy VFormat same as regex on WStringEditor
> 3. validate VFormat in PO (model layer) - this would enable applying VFormat validation also in REST or other sources
> 
> The changes here for WStringEditor are covering the 2)
> 
> The changes on jquery.maskedinput.js doesn't seem necessary if the implementation must be the 1)
> 
> What do you prefer? to close this and tackle those three tickets in a different pull request? Or change this PR to cover just the second case?


@CarlosRuiz-globalqss Let's change this PR to just the second case and I will open other tickets later for the other improvements, i'ts ok?

org.adempiere.ui.zk/WEB-INF/src/web/js/jquery/maskedinput/jquery.maskedinput.js
CarlosRuiz-globalqss: @joseleite550 - as I understand the scope of the current fix is not about the search - it will be solved in a different way.   So, I think this part is not necessary, we'll review its need in the other ticket, agree?

joseleite550: removed

CarlosRuiz-globalqss (changes requested):
Hi @joseleite550 / @VitorHenri 

I did some tests.
Can you please apply the following patch on top of this and conduct your tests to check that everything keeps going well?
[PR2641pr0.patch.txt](https://github.com/user-attachments/files/18849023/PR2641pr0.patch.txt)

Regards,

Carlos Ruiz

CarlosRuiz-globalqss:
Hi @joseleite550 - have you found some time to check the suggested patch?

CarlosRuiz-globalqss:
ping :blush:

joseleite550:
@CarlosRuiz-globalqss This change did not work, when the field has focus on it it does not remove the vformat in the js, for this reason it would be necessary to validate the JS file that asked me to remove it

joseleite550:
This way, with the popup you added, in addition to not searching, it keeps giving an error preventing the search.

image

CarlosRuiz-globalqss:
> @CarlosRuiz-globalqss This change did not work, when the field has focus on it it does not remove the vformat in the js, for this reason it would be necessary to validate the JS file that asked me to remove it

I don't understand what you mean.

Why do you want to remove the VFormat?

Can you please describe your failing test case so I can reproduce it?

I tested and in my case is I don't like that the save button is not showing the real value when there is an error.  Something that still need to be fixed

> This way, with the popup you added, in addition to not searching, it keeps giving an error preventing the search.

Also don't understand this comment, I just tested and the popup doesn't stop from searching.


Case is, the current pull request is not working fine, and the patch I provided improved it a bit, but is still not complete from what I see (the save button is leaving the invalid value in the field and the correct value in the database).

We need to find a pull request that works correctly.

joseleite550:
@CarlosRuiz-globalqss any position on?

joseleite550:
@CarlosRuiz-globalqss  any position on? I applied the patch

CarlosRuiz-globalqss (approved):
tested

PR #2643, opened 2025-01-16
IDEMPIERE-6386 Detail tabs don't have control over slow or big queries (FHCA-6139)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6386

# Pull Request Checklist

### Tests

### DIFF
--- a/migration/iD12/oracle/202501170030_IDEMPIERE-6386.sql
+++ b/migration/iD12/oracle/202501170030_IDEMPIERE-6386.sql
@@ -5,6 +5,6 @@ SET SQLBLANKLINES ON
 SET DEFINE OFF
 
 -- Jan 17, 2025, 12:30:29 AM CET
-INSERT INTO AD_Message (MsgType,MsgText,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Message_ID,Value,EntityType,AD_Message_UU) VALUES ('I','The background loading of the records have finished.',0,0,'Y',TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,200927,'BackgroundLoadFinished','D','4ba78ee4-45c3-4359-92e1-f42e92f26a0e')
+INSERT INTO AD_Message (MsgType,MsgText,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Message_ID,Value,EntityType,AD_Message_UU) VALUES ('I','The background loading of the records has finished.',0,0,'Y',TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,200927,'BackgroundLoadFinished','D','4ba78ee4-45c3-4359-92e1-f42e92f26a0e')
 ;
 
--- a/migration/iD12/oracle/202501170030_IDEMPIERE-6386.sql
+++ b/migration/iD12/oracle/202501170030_IDEMPIERE-6386.sql
@@ -1,0 +1,10 @@
+-- IDEMPIERE-6386 Detail tabs don't have control over slow or big queries (FHCA-6139)
+SELECT register_migration_script('202501170030_IDEMPIERE-6386.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Jan 17, 2025, 12:30:29 AM CET
+INSERT INTO AD_Message (MsgType,MsgText,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Message_ID,Value,EntityType,AD_Message_UU) VALUES ('I','The background loading of the records have finished.',0,0,'Y',TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,200927,'BackgroundLoadFinished','D','4ba78ee4-45c3-4359-92e1-f42e92f26a0e')
+;
+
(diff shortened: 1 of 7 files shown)


org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -1168,7 +1171,6 @@ public void waitLoadingForRow(int row) {
 			log.warning("Reached " + timeout + " seconds timeout loading row " + (row+1) + " for SQL=" + m_SQL);
 			//adjust row count
 			m_rowCount = m_sort.size();
-			throw new AdempiereException(Msg.getMsg(Env.getCtx(), LOAD_TIMEOUT_ERROR_MESSAGE));
```
hengsin, line 1171: why remove this ?

CarlosRuiz-globalqss: > why remove this ?

Raising this exception is leaving the window in a bad state, the user cannot search again, it's forced to close the window and reopen it.

So, instead of raising the exception this is managed in line 3232 below:
`} else if (m_rowLoadTimeout)`
which shows the message to the user gracefully and allows to keep using the window.

hengsin: ic but I think it is quite risky to continue to use it without the complete loading of detail tab, can lead to data corruption.

CarlosRuiz-globalqss: > ic but I think it is quite risky to continue to use the window without the complete loading of detail tab, can lead to data corruption.

not sure what you mean.

At this moment we have two scenarios:

1. what is happening now, suppose the detail tab has 3.000.000, when you open the window it shows nothing, no detail at all, the 3M records start loading in background, if you don't have enough memory the server crashes, if have a lot of memory it will take time, and the details are not shown at the end.  I remember somebody reported this to me saying that the details were not appearing in some window in a server that was slow, it was reproducible.  Also, closing the window doesn't stop the background load.
2. the behavior with this patch is different, instead of loading 3M records, is just load the max allowed (100K by default), and it shows the progress in the detail status bar, so the user can see what's happening there, and also if notifies the user (with a warning dialog) that there are more records that were not loaded.  In my tests also closing the window stopped the background load because it threw some error.

So, if you think partial detail load is dangerous, I'm not sure which would be the right solution, must we notify and close the window when detail number of records is > 100K?
Must we stop showing the header until the detail loads?  Because at this moment the header is shown immediately and the detail is loaded in background.

I don't feel that partial detail loading is dangerous, but I can imagine scenarios where people code can make it dangerous.

hengsin: hi @CarlosRuiz-globalqss , my comments above is for the removal of exception. I means it could be dangerous if the detail tab fail to load all the rows due to time out and we allow the user to keep using the window. Anyway, if you feel that's fine, I think we should at least show a warning message to the user. Same goes for the > 100k issue, I don't think we can be 100% sure that it would not cause any data issues. At the minimum, should show a warning and let user decide whether to continue with it.

CarlosRuiz-globalqss: Ah OK, then that's not a problem, after this patch a Warning is raised in both cases, when the header times out and when the detail times out.

When the timeout happens there is a Dialog showing the warning, and the message is also set in the status line.

If the timeout happens in both (header and detail) the user will see and need to close two dialogs.

The Dialog is raised in the line 3232 through a DataStatusEvent, the only danger I see is in case the Event doesn't reach the UI, because the Event is captured in AbstractADWindowContent.dataStatusChanged and re-scheduled asynchronously to reach the UI thread.

That's also the reason why I changed the `if` on line 1833, the event was just being re-scheduled for messages on the header tab, but now it re-schedule it also for the selected detail tab.

migration/iD12/oracle/202501170030_IDEMPIERE-6386.sql
```diff
@@ -0,0 +1,10 @@
+-- IDEMPIERE-6386 Detail tabs don't have control over slow or big queries (FHCA-6139)
+SELECT register_migration_script('202501170030_IDEMPIERE-6386.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- Jan 17, 2025, 12:30:29 AM CET
+INSERT INTO AD_Message (MsgType,MsgText,AD_Client_ID,AD_Org_ID,IsActive,Created,CreatedBy,Updated,UpdatedBy,AD_Message_ID,Value,EntityType,AD_Message_UU) VALUES ('I','The background loading of the records have finished.',0,0,'Y',TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,TO_TIMESTAMP('2025-01-17 00:30:28','YYYY-MM-DD HH24:MI:SS'),100,200927,'BackgroundLoadFinished','D','4ba78ee4-45c3-4359-92e1-f42e92f26a0e')
+;
```
hengsin, line 9: should be "The background loading of the records has finished." instead.

CarlosRuiz-globalqss: Thanks @hengsin - fixed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/CompositeADTabbox.java
```diff
@@ -1067,7 +1067,7 @@ private void onActivateDetail(IADTabpanel tabPanel) {
 			if (!tabPanel.getGridTab().isSortTab()) {
 				currentRow = tabPanel.getGridTab().getCurrentRow();
 			}
-			tabPanel.query(false, 0, 0);			
+			tabPanel.query(false, 0, tabPanel.getGridTab().getMaxQueryRecords());
```
hengsin, line 1070: I did a workspace reference search for "IADTabpanel.query(boolean currentRows, int currentDays, int maxRows)" and it seems we are doing this consistently (using GridTab.getMaxQueryRecords()). I think those call needs to be fixed as well.

CarlosRuiz-globalqss: Thanks @hengsin - I just found one more - fixed

hengsin: AbstractADWindowContext
- 651 inside doZoomToDetail
WQuickForm
- 460, inside dispose

CarlosRuiz-globalqss: > AbstractADWindowContext
>     * 651 inside doZoomToDetail
> WQuickForm
>     * 460, inside dispose

Those were changed on the first commit.

hengsin (approved):
looks good

PR #2652, opened 2025-01-23
IDEMPIERE-4827 Unify Evaluatee Implementation

CarlosRuiz-globalqss:
- fix case of @AD_Client_ID@ not being discovered on window

https://idempiere.atlassian.net/browse/IDEMPIERE-4827?focusedCommentId=51471

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
+++ b/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
@@ -237,9 +237,7 @@ public class DefaultEvaluatee implements Evaluatee {
 		
 		//try context if no data provider and not only window and not only tab
 		if (!globalVariable && m_dataProvider == null && Util.isEmpty(value) && !m_onlyWindow && (m_onlyTab == null || !m_onlyTab.booleanValue())) {
-			value = Env.getContext(ctx, m_windowNo, variableName);
-			if (Util.isEmpty(value))
-				value = Env.getContext(ctx, variableName);
+			value = Env.getContext(ctx, variableName);
 		}
 		
 		//fall back to default (if define)
--- a/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
+++ b/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
@@ -237,7 +237,9 @@ public class DefaultEvaluatee implements Evaluatee {
 		
 		//try context if no data provider and not only window and not only tab
 		if (!globalVariable && m_dataProvider == null && Util.isEmpty(value) && !m_onlyWindow && (m_onlyTab == null || !m_onlyTab.booleanValue())) {
-			value = Env.getContext(ctx, variableName);
+			value = Env.getContext(ctx, m_windowNo, variableName);
+			if (Util.isEmpty(value))
+				value = Env.getContext(ctx, variableName);
 		}
 		
 		//fall back to default (if define)
--- a/org.idempiere.test/src/org/idempiere/test/base/EnvTest.java
+++ b/org.idempiere.test/src/org/idempiere/test/base/EnvTest.java
@@ -258,6 +258,11 @@ public class EnvTest extends AbstractTestCase {
 		expr = "@C_BPartner_ID:0.Created<yyyy-MM-dd>@='"+created+"'";
 		evaluation = Evaluator.evaluateLogic(evaluatee, expr);
 		assertTrue(evaluation, "Unexpected logic evaluation result");
+
+		//@AD_Client_ID@
+		expr = "@AD_Client_ID@";
+		parsedText = Env.parseContext(Env.getCtx(), -1, expr, false, true); // like in GridFieldVO parsing ColumnSQL
+		assertEquals("11", parsedText, "Unexpected parsed text for "+expr);
 		
 		//custom context
 		Properties ctx = new Properties();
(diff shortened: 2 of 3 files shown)


org.idempiere.test/src/org/idempiere/test/base/EnvTest.java
```diff
@@ -258,6 +258,11 @@ public void testParseContext() {
 		expr = "@C_BPartner_ID:0.Created<yyyy-MM-dd>@='"+created+"'";
 		evaluation = Evaluator.evaluateLogic(evaluatee, expr);
 		assertTrue(evaluation, "Unexpected logic evaluation result");
+
+		//@AD_Client_ID@
+		expr = "@AD_Client_ID@";
+		parsedText = Env.parseContext(Env.getCtx(), -1, expr, false, true); // like in GridFieldVO parsing ColumnSQL
+		assertEquals("11", parsedText, "Unexpected parsed text for "+expr);
```
hengsin, line 265: I don't get this, the context variable is "#AD_Client_ID", not "AD_Client_ID"

org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
```diff
@@ -237,7 +237,9 @@ else if (m_tabNo <= 0)
 		
 		//try context if no data provider and not only window and not only tab
 		if (!globalVariable && m_dataProvider == null && Util.isEmpty(value) && !m_onlyWindow && (m_onlyTab == null || !m_onlyTab.booleanValue())) {
-			value = Env.getContext(ctx, variableName);
+			value = Env.getContext(ctx, m_windowNo, variableName);
```
hengsin, line 240: with m_windowNo should have been tried before reaching here.

hengsin:
"New and existing unit tests pass locally with my changes" is not tick, please do make sure your changes doesn't break any existing test cases.

CarlosRuiz-globalqss:
> "New and existing unit tests pass locally with my changes" is not tick, please do make sure your changes doesn't break any existing test cases.

Checked, all tests passed

org.idempiere.test/src/org/idempiere/test/base/EnvTest.java
CarlosRuiz-globalqss: The query is using @AD_Client_ID@ as the field context variable, not as the login client.

I set an example on test.idempiere.org at:
https://test.idempiere.org/webui/index.zul?Action=Zoom&TableName=AD_Column&Record_ID=1000380

Column = ClientName
SQL = `(SELECT Name FROM AD_Client WHERE AD_Client_ID=@AD_Client_ID@)`

In release-12, opening the Test window fails, because the @AD_Client_ID@ is replaced on the virtual column by empty string.

While in previous release it was working, for example I set the same test case at https://demo.ingeint.com/webui/index.zul (running release-9) and there the window opens without problem.


We know the issue can be easily fixed using `Test.AD_Client_ID` instead of the context variable, and probably that would be a better way, but anyways what this is showing is that backward compatibility was broken in DefaultEvaluatee and there are issues with virtual columns, and can be other collateral issues in other parts of the system.

org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
CarlosRuiz-globalqss: > with m_windowNo should have been tried before reaching here.

No, because GridFieldVO is parsing the context with this:
`vo.ColumnSQL = Env.parseContext(ctx, -1, vo.ColumnSQL, false, true);`
so, the windowNo is -1 and is not used above in the DefaultEvaluatee

hengsin: ok, will test, I'm completely lost on what's going on here ...

hengsin: Found out why, it is due to line 633 in Env.java where it auto add "#" to the context variable. In other word, `Env.getContext(Env.getCtx(), -1, "#AD_Client_ID", false)` and `Env.getContext(Env.getCtx(), -1, "AD_Client_ID", false)` will give you the same result. It is a special thing though in that method, as the same doesn't happens with `Env.getContext(Env.getCtx(), "#AD_Client_ID")` and `Env.getContext(Env.getCtx(), "AD_Client_ID")`.

hengsin: I think if we want to keep this behaviour for backward compatibility, instead of the changes above, we should duplicate the auto add "#" logic to the `Env.getContext(Env.getCtx(), "AD_Client_ID")` method.

hengsin: Propose alternative patch:
[env.globalcontext.patch.txt](https://github.com/user-attachments/files/18544710/env.globalcontext.patch.txt)

CarlosRuiz-globalqss: Thanks @hengsin 

At first I was not convinced by your patch because it creates another backward incompatibility.

But after analyzing potential scenarios it seems that's harmless and your patch makes more consistent the two methods in the Env class.

The new test case passed correctly, and it solves the problem in our migration.

Approving this.

PR #2653, opened 2025-01-24
IDEMPIERE-6386 Detail tabs don't have control over slow or big queries (FHCA-6139)

CarlosRuiz-globalqss:
- fix misleading "no rows found" message instead of timeout
- optimize waitLoadingForRow heavily called
- log info level queries that timeout
- show correctly the final number of records when background loading takes place

https://idempiere.atlassian.net/browse/IDEMPIERE-6386?focusedCommentId=51472

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/GridTable.java
+++ b/org.adempiere.base/src/org/compiere/model/GridTable.java
@@ -1143,11 +1143,19 @@ public class GridTable extends AbstractTableModel
 	 * @param row
 	 */
 	public void waitLoadingForRow(int row) {
+		// optimization - this method is called too frequently
+		// avoid processing when the criteria of whiles and if's below is not  met
+		if ( ! (row >= m_sort.size() || m_sort.size() == 0) )
+			return;
 		//	need to wait for data read into buffer
 		int loops = 0;
 		//wait for [timeout] seconds
 		int timeout = MSysConfig.getIntValue(MSysConfig.GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS, DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS, Env.getAD_Client_ID(Env.getCtx()));
-		while (row >= m_sort.size() && m_loaderFuture != null && !m_loaderFuture.isDone() && loops < timeout)
+		if (timeout <= 0)
+			timeout = 120; /* wait max 2 minutes if the load timeout is not set, WARNING, this is not recommended as the query keeps running in the database */
+		else
+			timeout += 5; /* give 5 seconds extra for the query to timeout first in the database */
+		while (row >= m_sort.size() && m_loaderFuture != null && !m_loaderFuture.isDone() && !m_rowLoadTimeout && loops < timeout)
 		{
 			if (log.isLoggable(Level.FINE)) log.fine("Waiting for loader row=" + row + ", size=" + m_sort.size());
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -3216,14 +3230,15 @@ private void doRun() {
...
+				if (log.isLoggable(Level.INFO)) log.info("Query timed out -> " + m_SQL);
 			}
 			finally
 			{
 				close();
 			}
 			
 			// Background loading without initial rowCount, inform final loaded rows
-			if (m_rowCountTimeout && m_sort.size() > 0)
+			if (backgroundLoad && m_sort.size() > 0)
```
hengsin, line 3241: this is changing the original logic, perhaps it should be `((m_rowCountTimeout || backgroundLoad) && m_sort.size() > 0)` instead.

org.adempiere.base/src/org/compiere/model/GridTable.java
```diff
@@ -1143,11 +1143,19 @@ public Object getValueAt (int row, int col)
...
+		// avoid processing when the criteria of whiles and if's below is not  met
+		if ( ! (row >= m_sort.size() || m_sort.size() == 0) )
+			return;
 		//	need to wait for data read into buffer
 		int loops = 0;
 		//wait for [timeout] seconds
 		int timeout = MSysConfig.getIntValue(MSysConfig.GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS, DEFAULT_GRIDTABLE_LOAD_TIMEOUT_IN_SECONDS, Env.getAD_Client_ID(Env.getCtx()));
-		while (row >= m_sort.size() && m_loaderFuture != null && !m_loaderFuture.isDone() && loops < timeout)
+		if (timeout <= 0)
+			timeout = 120; /* wait max 2 minutes if the load timeout is not set, WARNING, this is not recommended as the query keeps running in the database */
```
hengsin, line 1155: If it is a "warning", perhaps we should log it as a warning message. If we log this, would be better if we don't log the same warning > 1 times (I think we don't have such support now).

org.adempiere.base/src/org/compiere/model/GridTable.java
CarlosRuiz-globalqss: > this is changing the original logic, perhaps it should be `((m_rowCountTimeout || backgroundLoad) && m_sort.size() > 0)` instead.

I noticed that informing the final rows is just needed when loading in background, not when the count times out.

The scenario is like this:
* I opened "Change Audit" window with 27 million records - the count times out showing "Records = ?" in the Find dialog
* then I search for changes on a table which returns few records, the query is very fast and doesn't require background loading
* in this case, there is no need to inform at the end the final loaded rows, because there was no background load, the system shows correctly the number of records found

org.adempiere.base/src/org/compiere/model/GridTable.java
CarlosRuiz-globalqss: > If it is a "warning", perhaps we should log it as a warning message. If we log this, would be better if we don't log the same warning > 1 times (I think we don't have such support now).

done, I think is OK to flood the log with these warnings as this becomes a really bad practice that can potentially crash the server

hengsin (approved):
looks good

PR #2655, opened 2025-01-26
IDEMPIERE-6398 Server crash after a user without report permission tries to run a report

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6398

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/fi/jawsy/jawwa/zk/atmosphere/AtmosphereServerPush.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/fi/jawsy/jawwa/zk/atmosphere/AtmosphereServerPush.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin 

I would like to ask your code review and tests here.

The patch here solves the immediate problem, but I feel there can be something deeper that could cause trouble in other parts of the system.

The thing goes like this:
* The ZkReportViewer constructor is called
* because the user doesn't have permissions to execute the report a Dialog is raised and onClose() is called
* the onClose call clears sets m_reportEngine variable to null
* after that there is a setTitle using m_reportEngine that throws a Null Pointer Exception (seems like the code would assume that the onClose does a return, but it doesn't, so I added it explicitly here)

What I found really strange is that if you set a debug breakpoint on `org.zkoss.zk.ui.impl.UiEngineImpl.execUpdate` line 1323, there is an endless loop at this point.  For some reason the `nextEvent` there keeps processing the same event again and again.

I tested the exactly same scenario on release-10 and the code is the same but there is no endless loop in r10.  In release-10 two dialogs are shown, the NPE dialog and the cannot-report dialog.

So, as the code is same but the behavior is different, I'm afraid it could be a zk bug?

NOTE: The bug was discovered by a customer in a release-11 installation, so it's affecting 11,12,13

PR #2681, opened 2025-02-14
IDEMPIERE-6429: Missing images in iceblue_c theme

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-6429

# Pull Request Checklist

### Tests
### Documentation


uthadehikaru (commented):
hi @nmicoud , check your PR. looks good. just missing field suggestion icon. I created patch this patch to use font icon for field suggestion.
[IDEMPIERE-6429.diff.txt](https://github.com/user-attachments/files/18797068/IDEMPIERE-6429.diff.txt)

nmicoud:
Hi @uthadehikaru 
The image for FieldSuggession is missing from the beginning, and I think that should be addressed in a separate ticket.
And ideally we should find a dedicated image. I mean, the popup is "weird" as all images are colorized but this one

image

But ATM, I didn't find an appropriate image that could be use for FieldSuggestion :-/

hengsin:
@nmicoud , my su
hint-16
hint-22

CarlosRuiz-globalqss:
>> hi @nmicoud , I checked some image that is not used on any code and why some image has three different size and other only one size. should we remove any image that not used?

> hi @uthadehikaru I think we should let them in as they could be used by external plugins

Hi, I checked in core and it seems for webui just the size 24px is being used.

The only point where I see the 16px icons are used is probably in org.compiere.install.util.AppsAction

The 24px can be switched to 16px in ZK_TOOLBAR_BUTTON_SIZE in the theme preference.zul - but core uses just 24

Also org.adempiere.webui.component.WAppsAction.getImage is reading 16px images, but is never called with the small=true parameter in core.

So, for the sake of simplicity I think the 16px could be deprecated, but that would be made in a different ticket, and with a migration note for plugin maintainers.  If a plugin uses 16px images then it needs to switch to 24px or implement their own theme fragment.

WDYT?  

cc: @hengsin ?

nmicoud:
Hi @CarlosRuiz-globalqss 
16px icons are used on popup for Zoom, Refresh, ... (see WEditorPopupMenu.init)

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss 16px icons are used on popup for Zoom, Refresh, ... (see WEditorPopupMenu.init)

ah I see, looking for 16.png in core brings 136 matches with 56 hardcoded images.

So, not a big gain to change that.  Agree to keep it as it is now.

uthadehikaru (approved):
I think we can separate the cleaning for other ticket. If there is no any additional concern, we can merge this ticket

PR #2685, opened 2025-02-20
IDEMPIERE-6432 Fix links in Home server page

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6432


d-ruiz:
Hi @CarlosRuiz-globalqss, 

I think it is important to keep a link to GitHub, after all, we're a software project, so code is king. 

So, I would keep the changes you made. But then there are two links to the website, one to get Involved and one for community resources. My suggestion is to replace the community resources button with one that redirects to GitHub, it can be named: Source Code, Code base, GitHub ...

What do you think?

CarlosRuiz-globalqss:
> I think it is important to keep a link to GitHub, after all, we're a software project, so code is king.
> So, I would keep the changes you made. But then there are two links to the website, one to get Involved and one for community resources. My suggestion is to replace the community resources button with one that redirects to GitHub, it can be named: Source Code, Code base, GitHub ...

Thanks @d-ruiz - committed

d-ruiz (approved):
Looks good

PR #2688, opened 2025-02-23
IDEMPIERE-6435 Add "Hide on Zero" to Document Status (Activities)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6435

- Implement hide when zero

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/WDocumentStatusPanel.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/WDocumentStatusPanel.java
@@ -122,6 +122,15 @@ public class WDocumentStatusPanel extends Panel {
 		lastRefreshCount = 0;
 		for (WDocumentStatusIndicator indicator : indicatorList) {
 			indicator.refresh();
+			boolean isVisible = indicator.isVisible();
+			boolean newVisible = (indicator.getStatusCount() != 0 || !indicator.getDocumentStatus().isHideWhenZero());
+			if (isVisible != newVisible) {
+				try {
+					indicator.setVisible(newVisible);
+				} catch (Exception e) {
+					e.printStackTrace();
+				}
+			}
 			if (indicator.getDocumentStatus().getAD_Client_ID() == 0)
 				lastRefreshCount += indicator.getStatusCount();
 		}
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/dashboard/DPDocumentStatus.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/dashboard/DPDocumentStatus.java
@@ -39,6 +39,8 @@ import org.compiere.Adempiere;
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
 import org.compiere.util.Util;
+import org.zkoss.zk.ui.DesktopUnavailableException;
+import org.zkoss.zk.ui.Executions;
 import org.zkoss.zk.ui.event.Event;
 import org.zkoss.zk.ui.event.EventListener;
 import org.zkoss.zk.ui.event.Events;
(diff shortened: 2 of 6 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/dashboard/DPDocumentStatus.java
```diff
@@ -57,8 +59,15 @@ public class DPDocumentStatus extends DashboardPanel implements EventListener<Ev
...
-		statusPanel.refresh();
-		template.executeAsync(this);
+		try {
+			Executions.activate(template.getDesktop());
+			statusPanel.refresh();
+			template.executeAsync(this);
+		} catch (DesktopUnavailableException | InterruptedException e) {
+			e.printStackTrace();;
+		} finally {
+			Executions.deactivate(template.getDesktop());
```
hengsin, line 69: this is highly not recommended.

CarlosRuiz-globalqss: Thanks @hengsin 

The issue is because I needed to change the visibility of the indicator.

When the code indicator.setVisible was reached, an error was throws stating that the component can only be accessed on listeners.  I think that's caused because we're running the refresh on a non-UI thread.

Googling around the solution I found in stackoverflow (or zk forums) was to use the Executions.activate/deactivate

If that's not a good idea, how could we solve the issue, to change visibility and color running on non-UI thread?

hengsin: you should do that in the updateUI method instead. Using activate/deactivate means the whole UI is lock while the refresh method is executing.

CarlosRuiz-globalqss: > you should do that in the updateUI method instead. Using activate/deactivate means the whole UI is lock while the refresh method is executing.

ready @hengsin - solved in last commit

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/WDocumentStatusPanel.java
```diff
@@ -131,12 +144,46 @@ public void refresh() {
...
+			for (Group group : groups) {
+				if (visibleGroups.contains(group))
+					group.setVisible(true);
+				else
+					group.setVisible(false);
+			}
+		}
+		// hide the Activities dashboard if all indicators are hidden, otherwise show it
+		if (getParent() != null && getParent().getParent() != null && getParent().getParent().getParent() != null)
+			getParent().getParent().getParent().setVisible(isPanelVisible);
```
hengsin: This is fragile especially since it is not cover by automated test. Need more robust approach.

CarlosRuiz-globalqss: > This is fragile especially since it is not cover by automated test. Need more robust approach.

thanks @hengsin - fixed

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/graph/WDocumentStatusPanel.java
```diff
@@ -131,12 +145,54 @@ public void refresh() {
...
 		for (WDocumentStatusIndicator indicator : indicatorList) {
+			boolean isVisible = indicator.isVisible();
+			boolean newVisible = (indicator.getStatusCount() != 0 || !indicator.getDocumentStatus().isHideWhenZero());
+			if (isVisible != newVisible) {
+				indicator.setVisible(newVisible);
+				if (indicator.getParent() instanceof Row)
+					indicator.getParent().setVisible(newVisible);
+			}
+			if (newVisible && indicator.getDocumentStatus().getName_PrintColorZero_ID() > 0)
+				indicator.decorateNameLabel();
```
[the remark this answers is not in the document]
CarlosRuiz-globalqss: no, because the decoration must also be re-applied when the count changes from 0 to another value

PR #2690, opened 2025-02-24
IDEMPIERE-6436 Error inserting permission record for System Administrator role (FHCA-6351)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6436

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
+++ b/org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
@@ -229,7 +229,12 @@ public class PoFiller{
 						}
 					}
 				}
-				if (id != null && refTableName != null) {
+				if (id instanceof Number && ((Number)id).intValue() == 0) {
+					if (refTableName != null && MTable.isZeroIDTable(refTableName)) {
+						po.set_ValueNoCheck(columnName, id);
+						return id;
+					}
+				} else if (id != null && refTableName != null) {
 					if (foreignTable != null) {
 						if (isMulti) {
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
Forced push as is a fix for a broken important process

org.adempiere.pipo/src/org/adempiere/pipo2/PoFiller.java
```diff
@@ -229,7 +229,12 @@ public Object setTableReference(String qName) {
 						}
 					}
 				}
-				if (id != null && refTableName != null) {
+				if (id instanceof Number && ((Number)id).intValue() == 0) {
+					if (refTableName != null && MTable.isZeroIDTable(refTableName)) {
+						po.set_ValueNoCheck(columnName, id);
+						return id;
+					}
+				} else if (id != null && refTableName != null) {
```
hieplq, line 237: The check for null should come first, or it should be removed.

hengsin: This is fine since "id instanceof Number" will evaluate  to false if id is null

hieplq: but this one make NPE 
(Number)id).intValue()

hengsin: no, that's not the case here. Java do short circuit evaluation, if the first part is false, it wouldn't evaluate the second part of the logic expression.

PR #2749, opened 2025-04-17
IDEMPIERE-6498: Error when trying to register a time slot to a resour…

nmicoud:
…ce with an UoM without symbol

https://idempiere.atlassian.net/browse/IDEMPIERE-6498

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WAssignmentEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WAssignmentEditor.java
@@ -59,7 +59,7 @@ import org.zkoss.zk.ui.event.Events;
  */
 public class WAssignmentEditor extends WEditor implements ContextMenuListener {
 	
-	private static final String RETRIEVE_RESOURCE_ASSIGNMENT_SQL = "SELECT r.Name,ra.AssignDateFrom,ra.Qty,uom.UOMSymbol "
+	private static final String RETRIEVE_RESOURCE_ASSIGNMENT_SQL = "SELECT r.Name,ra.AssignDateFrom,ra.Qty,COALESCE(uom.UOMSymbol,uom.Name) "
 			+ "FROM S_ResourceAssignment ra, S_Resource r, S_ResourceType rt, C_UOM uom "
 			+ "WHERE ra.S_ResourceAssignment_ID=?"
 			+ " AND ra.S_Resource_ID=r.S_Resource_ID"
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss (commented):
Hi @nmicoud 

in CreateFrom, CreateFromInvoice and CreateFromShipment use 
```
COALESCE(uom.UOMSymbol,uom.Name)
```

not sure if is better for consistency to use the same here?

nmicoud:
hi @CarlosRuiz-globalqss 
As the name is already on the listbox, that will be no interest.
In this case (no symbol), better to show "" ? and test if rs.getString(5) is empty before doing the trim?

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss As the name is already on the listbox, that will be no interest. In this case (no symbol), better to show "" ? and test if rs.getString(5) is empty before doing the trim?

Hi @nmicoud - I'm not following you, where is the name in the listbox?

I'm testing in test.idempiere.org and I don't see the UOM name anywhere

CarlosRuiz-globalqss:
Ah, I think I understand what you mean, the UOM field says "Hour", but that's not a problem, the description of the resource assignment is just in memory from what I see.

This sounds like an uncommon case (I think the print formats are printing the symbol)

My suggestion is to keep consistency with the other classes, but this is not expected to happen really.

nmicoud:
Hi @CarlosRuiz-globalqss 
No I read too fast, your suggestion is good
So I've replaced _+ "uom.C_UOM_ID,COALESCE(uom.UOMSymbol, uom.X12DE355) "_
with your suggesion
_+ "uom.C_UOM_ID,COALESCE(uom.UOMSymbol,uom.Name)"_
in order to be consistent

Thanks

hengsin:
It seems perhaps it is not right here that uomsymbol is nullable, for e.g, many print format expect uomsymbol is not null.

nmicoud:
I don't have any opinion :-/
I faced the issue while making test and I didn't fill properly the UoM.
So maybe making the column as mandatory with default value "?" could be enough? (but I don't use UoM frequently so not sure about edge effects)

CarlosRuiz-globalqss:
> It seems perhaps it is not right here that uomsymbol is nullable, for e.g, many print format expect uomsymbol is not null.

Because that's too old defined that way (compiere times) I would prefer to leave it like that, in case somebody wants to make it mandatory, it can be defined mandatory at field level, the print formats already configured must be managing the null case.

CarlosRuiz-globalqss (changes requested):
Hi @nmicoud 

seems the change is still not complete, I tested and it requires also the following:

[pr2749PR0.patch.txt](https://github.com/user-attachments/files/19854432/pr2749PR0.patch.txt)

nmicoud:
hi @CarlosRuiz-globalqss 
Patch integrated

PR #2756, opened 2025-04-23
IDEMPIERE-6506 Translation cache not reset after update by user

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-6506

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -2212,7 +2212,7 @@ public abstract class PO
 	private static final String TRANSLATION_CACHE_TABLE_NAME = "PO_Trl";
 	
 	/**	Cache						*/
-	private static CCache<String,String> trl_cache	= new CCache<String,String>(TRANSLATION_CACHE_TABLE_NAME, 5, CCache.DEFAULT_EXPIRE_MINUTE, true);
+	private static CCache<String,String> trl_cache	= new CCache<String,String>(TRANSLATION_CACHE_TABLE_NAME, 5, CCache.DEFAULT_EXPIRE_MINUTE, false);
 	/** Cache for foreign keys */
 	private static CCache<Integer,List<ValueNamePair>> fks_cache	= new CCache<Integer,List<ValueNamePair>>("FKs", 5);
 
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -2212,7 +2212,7 @@ public abstract class PO
 	private static final String TRANSLATION_CACHE_TABLE_NAME = "PO_Trl";
 	
 	/**	Cache						*/
-	private static CCache<String,String> trl_cache	= new CCache<String,String>(TRANSLATION_CACHE_TABLE_NAME, 5);
+	private static CCache<String,String> trl_cache	= new CCache<String,String>(TRANSLATION_CACHE_TABLE_NAME, 5, CCache.DEFAULT_EXPIRE_MINUTE, true);
 	/** Cache for foreign keys */
 	private static CCache<Integer,List<ValueNamePair>> fks_cache	= new CCache<Integer,List<ValueNamePair>>("FKs", 5);
 
@@ -2739,6 +2739,9 @@ public abstract class PO
 				success = false;
 			}
 		}
+		
+		//collect changed columns for translation cache reset below
+		List<String> updatedColumns = new ArrayList<>();
 		//	OK
 		if (success)
 		{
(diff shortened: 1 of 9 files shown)


org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -2803,6 +2809,57 @@ public void afterClose(Trx trx) {
...
+			for (int i = 0; i < parentInfo.getColumnCount(); i++)
+			{
+				String columnName = parentInfo.getColumnName(i);
+				if (parentInfo.isColumnTranslated(i))
+				{
+					translatedColumns.add(columnName);
+				}
+			}
+			if (translatedColumns.size() > 0) {
+				int id = get_ValueAsInt(table.get_KeyColumns()[0]);
```
CarlosRuiz-globalqss, line 2826: get_KeyColumns is for MTable, it must use getKeyColumns instead

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -2803,6 +2809,57 @@ public void afterClose(Trx trx) {
 				else if (get_ID() > 0)
 					Adempiere.getThreadPoolExecutor().submit(() -> CacheMgt.get().newRecord(p_info.getTableName(), get_ID()));
 			}
+		} else if (p_info.getTableName().endsWith("_Trl") && CacheMgt.get().hasCache(TRANSLATION_CACHE_TABLE_NAME) && !newRecord) {
+			MTable table = MTable.get(getCtx(), p_info.getTableName().substring(0, p_info.getTableName().length() - 4));
+			POInfo parentInfo = POInfo.getPOInfo(getCtx(), table.getAD_Table_ID());
+			// take all translated columns, is_valuechanged(columnName) not working at this point
```
CarlosRuiz-globalqss, line 2815: a list of changed columns could be obtained around line 2763, it's too aggressive to reset all columns as it will take extra time making more dangerous to get an invalid cache

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -2803,6 +2809,57 @@ public void afterClose(Trx trx) {
 				else if (get_ID() > 0)
 					Adempiere.getThreadPoolExecutor().submit(() -> CacheMgt.get().newRecord(p_info.getTableName(), get_ID()));
 			}
+		} else if (p_info.getTableName().endsWith("_Trl") && CacheMgt.get().hasCache(TRANSLATION_CACHE_TABLE_NAME) && !newRecord) {
```
CarlosRuiz-globalqss, line 2812: Checking the logic, I think all block of MRecentItems and reset cache must be enclosed within the if (success) block.  There is no need to reset cache if the update was not successful

org.adempiere.base/src/org/compiere/util/ResetCacheCallable.java
```diff
@@ -28,7 +28,17 @@ public class ResetCacheCallable implements Callable<Integer>, Serializable
...
+	private Integer Record_ID;
+	private String key;
+	
+	/**
+	 * @param tableName
+	 * @param key
+	 */
+	public ResetCacheCallable(String tableName, String key)
+	{
+		this.key = key;
```
CarlosRuiz-globalqss, line 40: need here also the line:
```
this.tableName = tableName;
```

as there is no tablename is resetting the whole cache and that's why the unit test works

CarlosRuiz-globalqss (changes requested):
Hi @hengsin - after the fixes I request here the unit test is not working, so, there is still something missing.

I checked two things:

one idea: what if we make internally all the caches String, also the int caches?   Would that simplify the code?

one discovery: we cannot use this approach on MSysConfig.afterSave, unfortunately it seems we need to keep there resetting the whole cache of MSysConfig because the key is composed by other records too.

org.adempiere.base/src/org/compiere/model/PO.java
```diff
@@ -2209,8 +2209,10 @@ final protected void setUpdatedBy (int AD_User_ID)
 		set_ValueNoCheck ("UpdatedBy", Integer.valueOf(AD_User_ID));
 	}	//	setAD_User_ID
 
+	private static final String TRANSLATION_CACHE_TABLE_NAME = "PO_Trl";
+	
 	/**	Cache						*/
-	private static CCache<String,String> trl_cache	= new CCache<String,String>("PO_Trl", 5);
+	private static CCache<String,String> trl_cache	= new CCache<String,String>(TRANSLATION_CACHE_TABLE_NAME, 5, CCache.DEFAULT_EXPIRE_MINUTE, true);
```
CarlosRuiz-globalqss, line 2215: I think we don't have any distributed cache in iDempiere core, distributed cache are a lot slower than local caches, I tested with non-distributed and it worked fine, is there any need here for distributed?

hengsin: Without distributed cache, changes make in one sever will not auto refresh the translation cache of another server. You can test that out by running 2 instance of idempiere (with hazelcast tcp/ip join enable).

hengsin: The reset call will get transmitted to other server nodes. With that, the distributed cache is probably not necessary, will re-test to confirm.

CarlosRuiz-globalqss (approved):

PR #2768, opened 2025-04-30
IDEMPIERE-6526 Performance improvement exporting to XLSX

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6526

# Pull Request Checklist

### Tests

### DIFF
--- a/migration/iD12/oracle/202505011526_IDEMPIERE-6526.sql
+++ b/migration/iD12/oracle/202505011526_IDEMPIERE-6526.sql
@@ -1,0 +1,10 @@
+-- IDEMPIERE-6526 Performance improvement exporting to XLSX
+SELECT register_migration_script('202505011526_IDEMPIERE-6526.sql') FROM dual;
+
+SET SQLBLANKLINES ON
+SET DEFINE OFF
+
+-- May 1, 2025, 3:26:37 PM CEST
+INSERT INTO AD_SysConfig (AD_SysConfig_ID,AD_Client_ID,AD_Org_ID,Created,Updated,CreatedBy,UpdatedBy,IsActive,Name,Value,Description,EntityType,ConfigurationLevel,AD_SysConfig_UU) VALUES (200271,0,0,TO_TIMESTAMP('2025-05-01 15:26:37','YYYY-MM-DD HH24:MI:SS'),TO_TIMESTAMP('2025-05-01 15:26:37','YYYY-MM-DD HH24:MI:SS'),100,100,'Y','XLSX_EXPORT_USE_FAST_METHOD','Y','Defines if exporting to XLSX uses the streaming method (less memory intensive)','D','S','2d1b3c19-aee1-44d2-8d7e-a3ae7d92086c')
+;
+
--- a/migration/iD12/postgresql/202505011526_IDEMPIERE-6526.sql
+++ b/migration/iD12/postgresql/202505011526_IDEMPIERE-6526.sql
@@ -1,0 +1,7 @@
+-- IDEMPIERE-6526 Performance improvement exporting to XLSX
+SELECT register_migration_script('202505011526_IDEMPIERE-6526.sql') FROM dual;
+
+-- May 1, 2025, 3:26:37 PM CEST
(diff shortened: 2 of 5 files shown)


CarlosRuiz-globalqss:
Hi @uthadehikaru and @hengsin 

I tested the current pull request with good results.

The memory consumption of exporting big XLS files is lower and the speed is a lot better.

The only problem is at AbstractXLSXExporter.fixColumnWidth
because using the streaming SXSSFWorkbook cannot use autoSizeColumn, so we need to do some default width depending on the datatype of the column

Also, I just tested some basic stuff, it requires bigger testings.

hengsin:
hi @CarlosRuiz-globalqss , I don't think this is 100% backward compatible. It would be better if we can keep the existing export as an option for the user.

hengsin:
hi @CarlosRuiz-globalqss , because of column width and the notes in https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/streaming/SXSSFWorkbook.html - "known to produce documents that are incompatible with some clients".

It will also break any existing custom implementation of AbstractXLSXExporter

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , because of column width and the notes in https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/streaming/SXSSFWorkbook.html - "known to produce documents that are incompatible with some clients".

Ah OK, that's something I faced there yesterday, that solved using the constructor:
```
m_workbook = new SXSSFWorkbook(null, SXSSFWorkbook.DEFAULT_WINDOW_SIZE, true, true);
```
[fourth parameter](https://poi.apache.org/apidocs/dev/org/apache/poi/xssf/streaming/SXSSFWorkbook.html#SXSSFWorkbook-org.apache.poi.xssf.usermodel.XSSFWorkbook-int-boolean-boolean-) instructs to use shared strings, I think that was not the big problem at the end but without that it created bigger files, using shared strings create littler files but consumes more memory, however I think is manageable.

The big issue about "incompatible with some clients" I found is that the produced xlsx files shown an error with libreoffice (file is corrupt do you want to recover it?  and clicking yes was recovered correctly), but Excel online opened the file without issues.

After googling I found the culprit was on the zip version used, and that's the reason of the line:
```
m_workbook.setZip64Mode(Zip64Mode.Never);
```


> It will also break any existing custom implementation of AbstractXLSXExporter and existing plugin that uses the AbstractXLSXExporter and PrintDataXLSXExporter class. It is not just a change of internal implementation but also public API.

Well, that's a strong argument to keep both versions, so, I think having both versions and a SysConfig to switch back could be the most friendly, WDYT?

hengsin:
yeah, SysConfig is fine and we can use the new implementation as the default.

CarlosRuiz-globalqss:
> yeah, SysConfig is fine and we can use the new implementation as the default.

done

uthadehikaru (approved):
tested functionality and looks good.

hengsin:
hi @CarlosRuiz-globalqss , if we wants to maintain binary compatibility with existing plugin, we would have to keep existing AbstractXLSXExporter and PrintDataXLSXExporter and create new class for the streaming version (for e.g AbstractSXLSXExporter and PrintDataSXLSXExporter). Anyway, if that's not one of the objective here, this is good to go as it is.

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , if we wants to maintain binary compatibility with existing plugin, we would have to keep existing AbstractXLSXExporter and PrintDataXLSXExporter and create new class for the streaming version (for e.g AbstractSXLSXExporter and PrintDataSXLSXExporter). Anyway, if that's not one of the objective here, this is good to go as it is.

Thanks @hengsin 

I tried firstly that approach, but then the change was too big as I needed to change also ArrayExcelExporter, GridTabXLSXExporter, RModelExcelExporter, XlsxExporter and XLSXSReportRenderer.

The new code is backward compatible because is just using an upper interface instead of the implementation.

So, I think as you mention, because of binary compatibility the jar of a plugin extending AbstractSXLSXExporter potentially would not work directly, but it will compile against the new code without changes.

PR #2781, opened 2025-05-07
IDEMPIERE-6532 Regression: For Tab 0, Env.parseContext is always gett…

hengsin:
…ing value from window context instead of from tab 0 context

https://idempiere.atlassian.net/browse/IDEMPIERE-6532

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
+++ b/org.adempiere.base/src/org/compiere/util/DefaultEvaluatee.java
@@ -195,7 +195,7 @@ public class DefaultEvaluatee implements Evaluatee {
 						m_onlyTab != null ? m_onlyTab.booleanValue() : false);
 				value = Env.getContext(Env.getCtx(), m_windowNo, m_tabNo, keycolumnName, m_onlyTab != null ? m_onlyTab.booleanValue() : false);
 			}
-			else if (m_tabNo <= 0)
+			else if (m_tabNo < 0)
 			{
 				if (!tabOnly)
 					value = Env.getContext (ctx, m_windowNo, variableName, m_onlyWindow);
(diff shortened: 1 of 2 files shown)


nmicoud:
Hi @hengsin 
Tested and is ok when a field is empty
But is not ok is the field is not in the window (I mean there is no AD_Field_ID record at all ; in the test case, you need to delete _Bill_BPartner_ID_ field), user list is still empty

hengsin:
hi @nmicoud , I don't think invalid field (i.e non-exists field) is a valid use case to support.

nmicoud:
Hi @hengsin 
Yes that make sense. I was testing this as it was possible on v11.
But the workaround (or best practice?) is to add the default value, so I guess is ok this way

hengsin:
> Hi @hengsin Yes that make sense. I was testing this as it was possible on v11. But the workaround (or best practice?) is to add the default value, so I guess is ok this way

But what's the use case here to create validation rule with column/field that doesn't exists ?

nmicoud:
> > Hi @hengsin Yes that make sense. I was testing this as it was possible on v11. But the workaround (or best practice?) is to add the default value, so I guess is ok this way
> 
> But what's the use case here to create validation rule with column/field that doesn't exists ?

I have a table which is used on several windows (on per tenant as is easier to maintain customizations).
And in some of them, the 2nd field (XXA_DivAssurance_BPartner_ID) was not in AD_Field.
And the validation rule was set on the column, so it fails

hengsin:
Well, if that's needed, I will need some extra hours to find out the changes that causes the difference from r11 (I've try but not able to yet). If your guys have a fix for it, you can add it to here too.

hengsin:
hi @nmicoud ,

For validation rule, the invalid field expression doesn't works for r11 either.
I change the validation rule to `AD_User.C_BPartner_ID IN (@C_BPartner_ID@, @Bill_BPartner1_ID@)` and tested in r11:

https://github.com/user-attachments/assets/e2659e14-acaf-4c57-aabd-97642fb21e77

The initial value in User/Contact is actually populated by callout and MLookup's getDirect call (after couldn't get the user record using the validation rule). After I've remove the User/Contact value, save and re-open the record, the User/Contact dropdown is now empty.

nmicoud:
Hi @hengsin 
Thanks for looking into it, and sorry if I check wrongly and wasted your time.
As the validation rule was used on column, as you pointed, maybe our callout was/is doing some stuff to help populate fields.

CarlosRuiz-globalqss:
> hi @nmicoud ,
> 
> For validation rule, the invalid field expression doesn't works for r11 either. I change the validation rule to `AD_User.C_BPartner_ID IN (@C_BPartner_ID@, @Bill_BPartner1_ID@)` and tested in r11:

I'm not 100% following the logic here - but wouldn't be this test a non-existing field with a default?
Like `@Bill_BPartner1_ID:0@`

nmicoud (approved):
tested

PR #2790, opened 2025-05-12
IDEMPIERE-6540 Import/Export script doesn't support Oracle running as…

hengsin:
… Docker Container

https://idempiere.atlassian.net/browse/IDEMPIERE-6540

# Pull Request Checklist

### Tests
### Documentation


CarlosRuiz-globalqss:
Hi @hengsin - I have not tested yet, but reading the code I see the scripts are still using local sqlplus.

I think it can be worthy to use the sqlplus within the docker container, that way you don't need an oracle client installed in your machine.

I think the easiest is to run something like:
cat script.sql | $DOCKER_EXEC sqlplus ...

WDYT?

hengsin:
> Hi @hengsin - I have not tested yet, but reading the code I see the scripts are still using local sqlplus.
> 
> I think it can be worthy to use the sqlplus within the docker container, that way you don't need an oracle client installed in your machine.
> 
> I think the easiest is to run something like: cat script.sql | $DOCKER_EXEC sqlplus ...
> 
> WDYT?

I did try that but there are complication like getting the exit code from docker exec (for e.g for testing whether idempiere database user exists). I think it is a fair requirement that if you are using Oracle, you should at least have sqlplus install. It is also very easy this day to get oracle client install (https://www.oracle.com/my/database/technologies/instant-client/downloads.html).

CarlosRuiz-globalqss:
I use oracle docker locally, and no oracle client, but my approach is different, I docker folders on start the seed and the scripts, and then sqlplus can execute them locally.

CarlosRuiz-globalqss:
Checking here:
https://stackoverflow.com/questions/34447818/the-result-of-docker-exec-command

It seems docker exec is OK to get the exit code of the sqlplus - what is the inconvenient that you found with that?

I really would prefer not having to install an oracle client in my laptop (if that's not possible I think I'll pass from testing this, but please just push it if your tests are successful and not breaking the usual install).

hengsin:
> Checking here: https://stackoverflow.com/questions/34447818/the-result-of-docker-exec-command
> 
> It seems docker exec is OK to get the exit code of the sqlplus - what is the inconvenient that you found with that?
> 
> I really would prefer not having to install an oracle client in my laptop (if that's not possible I think I'll pass from testing this, but please just push it if your tests are successful and not breaking the usual install).

when using docker exec, the sqlplus -S -L idempiere/idempiere test doesn't provide the expected result. I will check the link above to see whether that can be make to work.

hengsin:
> I use oracle docker locally, and no oracle client, but my approach is different, I docker folders on start the seed and the scripts, and then sqlplus can execute them locally.

so the existing scripts works for you through the mapping of host folder to docker ?

ocurieles:
Hi,I think i can create a idempiere docker for oracle..

CarlosRuiz-globalqss:
> so the existing scripts works for you through the mapping of host folder to docker ?

Yes, mapping the folders to docker then you have there the scripts and the dmp seed available so you can run the "docker exec impdp" and "sqlplus script..." without needing to copy the files.

About impdp and expdp is needed to setup the permissions of the folder on the host, so the oracle user within docker can read/write from there.

CarlosRuiz-globalqss:
ah - I remember something else

I created some wrapper scripts for impdp, expdp and sqlplus

they are the same script and looks like:
```shell
docker exec -u oracle -e ORACLE_HOME=$ORACLE_DOCKER_HOME -e NLS_LANG=$NLS_LANG -i $ORACLE_DOCKER_CONTAINER $ORACLE_DOCKER_HOME/bin/$(basename $0) "$@"
```

CarlosRuiz-globalqss (approved):
tested with docker

PR #2803, opened 2025-05-30
IDEMPIERE-6563 Attach stacktrace and Unresolved to the 2pack as error_yyyymmddhhmmss.log (FHCA)

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6563

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.pipo/src/org/adempiere/pipo2/PackInNotifier.java
+++ b/org.adempiere.pipo/src/org/adempiere/pipo2/PackInNotifier.java
@@ -25,14 +25,21 @@
 
 package org.adempiere.pipo2;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.StringTokenizer;
 
+import org.compiere.model.MAttachment;
 import org.compiere.model.MClient;
 import org.compiere.model.MSysConfig;
+import org.compiere.model.X_AD_Package_Imp_Proc;
 import org.compiere.util.EMail;
 import org.compiere.util.Env;
 import org.compiere.util.KeyNamePair;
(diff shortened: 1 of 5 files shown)


org.adempiere.pipo/src/org/adempiere/pipo2/PackInNotifier.java
```diff
@@ -144,41 +157,73 @@ public void notifyRecipient() {
...
+			if (email != null)
+			{
+				if (!packIn.isSuccess()) {
+					email.setHeader("X-Priority", "1");
+					email.setHeader("Priority","Urgent");
+					email.setHeader("Importance","high");
+				}
+				while (st.hasMoreTokens())
+					email.addTo(st.nextToken());
+				email.send();
```
hengsin: no handling of email delivery failure ?

CarlosRuiz-globalqss: > no handling of email delivery failure ?

nothing to do there if the email fails.

Or maybe you're suggesting that if the email fails we attach the file independent of the new SysConfig?

hengsin: The thing is we wouldn't know the 2pack failure notification is not sent. I think we hit that before in a project.

CarlosRuiz-globalqss: > The thing is we wouldn't know the 2pack failure notification is not sent. I think we hit that before in a project.

yes - but what would be the suggestion?

I think with this new feature we can attach to the PackIn when email fails.  Would that  be enough?

hengsin: yes, that should help, at lease there's an indicator that the email notification delivery doesn't happens as it should

CarlosRuiz-globalqss: > yes, that should help, at lease there's an indicator that the email notification delivery doesn't happens as it should

Thanks @hengsin - ready, implemented suggestion

hengsin (approved):
looks good

PR #2813, opened 2025-06-12
IDEMPIERE-6575: Add Custom Key column to Doc sequence, allow make separate Series

PeterTakacs300:
https://idempiere.atlassian.net/browse/IDEMPIERE-6575

# Pull Request Checklist

### Tests
### Documentation


norbertbede:
pls help us review this important, critical change for us.

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -542,8 +569,7 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
 
 		//	create DocumentNo
 		StringBuilder doc = new StringBuilder();
-		if (prefix != null && prefix.length() > 0) {
-			String prefixValue = Env.parseVariable(prefix, po, trxName, false);
+		if (prefixValue != null && prefixValue.length() > 0) {
 			if (!Util.isEmpty(prefixValue))
```
hengsin: shouldn't use double if here, just change "!Util.isEmpty(prefixValue)" to "!Util.isEmpty(prefixValue, true)

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -553,8 +579,7 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
 		else
 			doc.append(next);
 
-		if (suffix != null && suffix.length() > 0) {
-			String suffixValue = Env.parseVariable(suffix, po, trxName, false);
+		if (suffixValue != null && suffixValue.length() > 0) {
 			if (!Util.isEmpty(suffixValue))
```
hengsin: same as line 572 above, shouldn't use double if here.

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -564,6 +589,48 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
 				+ " - Sequence=" + AD_Sequence_ID + " [" + trx + "]");
 		return documentNo;
 	}
+	
+	private Boolean isSequenceNoLevel = null;
+	private Boolean isUsePrefixAsKey = null;
+	private Boolean isUseSuffixAsKey = null;
```
hengsin: this is not thread safe, use AtomicBoolean instead, i.e "private AtomicBoolean isSequenceNoLevel = new AtomicBoolean(false);"

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -564,6 +589,48 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
...
+		return isSequenceNoLevel.booleanValue();
+	}
+	
+	/**
+	 * Is the prefix used as key for the sequence no
+	 * @return true if the prefix starts with prefix key
+	 */
+	public boolean isUsePrefixAsKey() {
+		if (isUsePrefixAsKey == null) {
+			isUsePrefixAsKey = Boolean.valueOf(!Util.isEmpty(getPrefix()) && getPrefix().contains(KEY_CONTEXT_VARIABLE+"@"));
```
hengsin: I think it is simpler to just use "/K" suffix as Key indicator for the "Prefix" and "Suffix" string. Probably should raise a warning to user if suffix/prefix is use to pick sequence no line but doesn't contains any variables. Also, if we wants a more intuitive UI, should use a check box that say "Use Suffix to Pick Sequence No Line" instead.

PeterTakacs300: Not sure if I understand correctly, is that what you mean to simply add /K at the end of the whole Prefix? That would not allow us to pick specific parameters to the key.

With the current approach you can add /K for each context variable you want to include, e.g. `GW/@DateOrdered<yyyyMM>@/@C_BPartner_ID/K@` <- here only the C_BPartner_ID will be included in the key

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -1271,5 +1354,164 @@ public String getOrgColumn() {
...
+				// Check for "KEY_CONTEXT_VARIABLE" suffix
+				if (var.substring(0, var.length()-1).endsWith(KEY_CONTEXT_VARIABLE)) {
+					isKey = true;
+					int kIndex = var.lastIndexOf(KEY_CONTEXT_VARIABLE);
+					var = var.substring(0, kIndex) + var.substring(kIndex + KEY_CONTEXT_VARIABLE.length());
+				}
+
+				// Parse the variable
+				String value = Env.parseVariable(var, evaluatee, false, false);;
+				if (value != null && isKey) {
```
hengsin: safer to use "!Util.isEmpty(value, true)" here

org.adempiere.base/src/org/compiere/model/MSequence.java
hengsin: ok, that make sense.

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: done

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: done

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: done

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: done

PeterTakacs300:
Added unit test for basic use cases + improved the MDocType.get method by transaction, so I can create a testing document type in my unit test, and also needed to use the new get method in MOrder.beforeSave, since I'm using an order for testing.

org.adempiere.base/src/org/compiere/model/MOrder.java
```diff
@@ -1371,7 +1371,7 @@ protected boolean beforeSave (boolean newRecord)
 		}
 
 		// IDEMPIERE-4318 Validation - Prepay Order must not allow Cash payment rule
-		MDocType dt = MDocType.get(getCtx(), getC_DocTypeTarget_ID());
+		MDocType dt = MDocType.get(getCtx(), getC_DocTypeTarget_ID(), get_TrxName());
```
hengsin: don't do that here, potential to create performance issue.

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,136 @@
...
+        seq.setCurrentNext(1000);
+        seq.setIncrementNo(1);
+        seq.setIsActive(true);
+        seq.setIsAutoSequence(true);
+        seq.setStartNewYear(false);
+        seq.setStartNewMonth(false);
+        seq.saveEx();
+
+        // Create test document type
+        MDocType docType = new MDocType(ctx, 0, trxName);
```
hengsin, line 55: create your test doctype outside of trx (i.e with trxName == null) and delete it at the end of the test.

PeterTakacs300: ok I'll do that instead

PeterTakacs300: done

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,140 @@
...
+        String docNo3 = order3.getDocumentNo();
+        assertTrue(docNo3.endsWith(newOrgID + TEST_SUFFIX), 
+                "Document number should end with new org ID key and suffix");
+        String seqPart3 = docNo3.substring((TEST_PREFIX + TEST_DATE_yyyyMM).length(), 
+                docNo3.length() - (String.valueOf(newOrgID) + TEST_SUFFIX).length());
+        assertEquals("1000", seqPart3, "Sequence number should restart at 1000 for new org key");
+        
+        // Delete the test records
+        docType.deleteEx(true);
+        seq.deleteEx(true);
```
hengsin, line 138: the 2 delete call should be wrap in a try finally block.

hengsin:
@PeterTakacs300 , please apply the attached patch. It uses Mockito to remove the need to create doctype and sequence outside of transaction.

[2813.patch.txt](https://github.com/user-attachments/files/20744049/2813.patch.txt)

PeterTakacs300:
Thanks @hengsin, patch applied and tested

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,142 @@
...
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+
+/**
+ * Tests for {@link org.compiere.model.MSequence} class.
+ * Specifically testing document number generation using AD_Sequence
+ * with new context variable keys feature (/K flag).
+ */
+@Isolated
+@Execution(ExecutionMode.SAME_THREAD)
```
hengsin: this 2 annotation is not needed.

hengsin (changes requested):

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
PeterTakacs300: removed the annotations

PeterTakacs300:
Pushed a minor optimisation improvement, and a fix for the use case, when the sequence is only "org" level (year, month, prefix and suffix is not key), in this case the current next value was not incremented.

hengsin:
To be documented in wiki

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -542,28 +574,64 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
...
+		
 		String documentNo = doc.toString();
 		if (s_log.isLoggable(Level.FINER)) s_log.finer (documentNo + " (" + incrementNo + ")"
 				+ " - Sequence=" + AD_Sequence_ID + " [" + trx + "]");
 		return documentNo;
 	}
+	
+	private AtomicBoolean isSequenceNoLevel = null;
+	private AtomicBoolean isUsePrefixAsKey = null;
+	private AtomicBoolean isUseSuffixAsKey = null;
```
hengsin: This wouldn't make it thread safe either. If you must do lazy initialization and cache the result, use either:
1. `AtomicReference<Boolean> isSequenceNoLevel = new AtomicReference<>()` OR
2. Use primitive int so that you can have tri-state of the variable. 
`private int isSequenceNoLevel=2; //2=unknown, 1=true, 0=false`

org.adempiere.base/src/org/compiere/model/MSequence.java
```diff
@@ -488,14 +520,14 @@ public static String getDocumentNoFromSeq(MSequence seq, String trxName, PO po)
...
 				{	// create sequence (CurrentNo = StartNo + IncrementNo) for this year/month/org and return first number (=StartNo)
 					next = startNo;
 
 					X_AD_Sequence_No seqno = new X_AD_Sequence_No(Env.getCtx(), 0, trxName);
 					seqno.setAD_Sequence_ID(AD_Sequence_ID);
 					seqno.setAD_Org_ID(docOrg_ID);
-					seqno.setCalendarYearMonth(calendarYearMonth);
+					seqno.setSequenceKey(keyParts.getKey());
 					seqno.setCurrentNext(startNo + incrementNo);
 					seqno.saveEx();
```
hengsin, line 532: This give the possibility of duplicate AD_Sequence_No record. Perhaps we should have a unique constraint for AD_Sequence_No(AD_Sequence_D, AD_Org_ID, SequenceKey) to prevent this from happening.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: Hi, I guess it's not worth the effort, the operation is not that expensive, I would just remove the global variables and call the `getPrefix().contains(KEY_CONTEXT_VARIABLE+"@")` each time, wdyt?

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: created the unique constraint

org.adempiere.base/src/org/compiere/model/MSequence.java
PeterTakacs300: done using `AtomicReference<Boolean>`

migration/iD13/postgresql/202506201409_IDEMPIERE-6575.sql
```diff
@@ -0,0 +1,19 @@
...
+-- Jun 20, 2025, 2:09:52 PM CEST
+INSERT INTO AD_IndexColumn (AD_Client_ID,AD_Org_ID,AD_IndexColumn_ID,AD_IndexColumn_UU,Created,CreatedBy,EntityType,IsActive,Updated,UpdatedBy,AD_Column_ID,AD_TableIndex_ID,SeqNo) VALUES (0,0,201762,'bcc86c03-2b40-4bad-9fb3-c3edc44da760',TO_TIMESTAMP('2025-06-20 14:09:52','YYYY-MM-DD HH24:MI:SS'),100,'D','Y',TO_TIMESTAMP('2025-06-20 14:09:52','YYYY-MM-DD HH24:MI:SS'),100,749,201289,10)
+;
+
+-- Jun 20, 2025, 2:10:21 PM CEST
+INSERT INTO AD_IndexColumn (AD_Client_ID,AD_Org_ID,AD_IndexColumn_ID,AD_IndexColumn_UU,Created,CreatedBy,EntityType,IsActive,Updated,UpdatedBy,AD_Column_ID,AD_TableIndex_ID,SeqNo) VALUES (0,0,201763,'bf557da7-d608-456d-a31a-ea6639c9f8af',TO_TIMESTAMP('2025-06-20 14:10:21','YYYY-MM-DD HH24:MI:SS'),100,'D','Y',TO_TIMESTAMP('2025-06-20 14:10:21','YYYY-MM-DD HH24:MI:SS'),100,262,201289,20)
+;
+
+-- Jun 20, 2025, 2:10:35 PM CEST
+ALTER TABLE AD_Sequence_No ADD CONSTRAINT ad_sequence_no_ad_org_sequencekey UNIQUE (AD_Org_ID,SequenceKey)
```
hengsin: this is not correct, should be AD_Sequence_ID, AD_Org_ID,SequenceKey instead.

PeterTakacs300: ah sorry I missed that, committed the fix

migration/iD13/oracle/202506201409_IDEMPIERE-6575.sql
```diff
@@ -0,0 +1,22 @@
...
+-- Jun 20, 2025, 2:09:52 PM CEST
+INSERT INTO AD_IndexColumn (AD_Client_ID,AD_Org_ID,AD_IndexColumn_ID,AD_IndexColumn_UU,Created,CreatedBy,EntityType,IsActive,Updated,UpdatedBy,AD_Column_ID,AD_TableIndex_ID,SeqNo) VALUES (0,0,201762,'bcc86c03-2b40-4bad-9fb3-c3edc44da760',TO_TIMESTAMP('2025-06-20 14:09:52','YYYY-MM-DD HH24:MI:SS'),100,'D','Y',TO_TIMESTAMP('2025-06-20 14:09:52','YYYY-MM-DD HH24:MI:SS'),100,749,201289,10)
+;
+
+-- Jun 20, 2025, 2:10:21 PM CEST
+INSERT INTO AD_IndexColumn (AD_Client_ID,AD_Org_ID,AD_IndexColumn_ID,AD_IndexColumn_UU,Created,CreatedBy,EntityType,IsActive,Updated,UpdatedBy,AD_Column_ID,AD_TableIndex_ID,SeqNo) VALUES (0,0,201763,'bf557da7-d608-456d-a31a-ea6639c9f8af',TO_TIMESTAMP('2025-06-20 14:10:21','YYYY-MM-DD HH24:MI:SS'),100,'D','Y',TO_TIMESTAMP('2025-06-20 14:10:21','YYYY-MM-DD HH24:MI:SS'),100,262,201289,20)
+;
+
+-- Jun 20, 2025, 2:10:35 PM CEST
+ALTER TABLE AD_Sequence_No ADD CONSTRAINT ad_sequence_no_ad_org_sequencekey UNIQUE (AD_Org_ID,SequenceKey)
```
hengsin: @PeterTakacs300 , need to remove this from migration script ( and the corresponding DROP Constraint statement in the subsequent migration script file ). This can cause unique constraint exception (and stop the execution of migration script) on table with existing data since the unique constraint doesn't include either client id or sequence id.

hengsin (changes requested):

migration/iD13/oracle/202506201409_IDEMPIERE-6575.sql
PeterTakacs300: yes I see thanks, I adjusted the migration scripts

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+    @Test
+    public void testDocumentNoWithContextVariableKey() {
+        Properties ctx = Env.getCtx();
+        String trxName = getTrxName();
+
+        // Create test sequence
+        MSequence seq = new MSequence(ctx, 0, trxName);
+        seq.setName("Test Sequence with Context Key");
+        seq.setDescription("Test sequence for document numbers with context variable key");
+        seq.setPrefix(TEST_PREFIX + "@Updated<yyyyMM>/K@"); // /K flag indicates context variable use
```
hengsin: This will not always match with `TEST_DATE_yyyyMM = "202506"` as Updated is fill with current system date instead of from "#Date". In my local testing, I have to change it to `@DateOrdered<yyyyMM>/K@` to pass the unit test. `@Updated<yyyyMM>/K@` with evaluate to "202507" (current system date) instead .

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+	        order.setDateOrdered(testDate);
+	        order.setDatePromised(testDate);
+	        order.saveEx();
+
+	        String docNo = order.getDocumentNo();
+	        assertNotNull(docNo, "Document number should not be null");
+
+	        // The document number should follow pattern: TestSeq-202506{seq}11-End
+	        assertTrue(docNo.startsWith(TEST_PREFIX + TEST_DATE_yyyyMM),
+	                "Document number should start with prefix and date key");
```
hengsin: suggest to change it to "Document number should start with prefix and date key: "+docNo

org.idempiere.test/src/org/idempiere/test/model/MSequenceTest.java
```diff
@@ -0,0 +1,137 @@
...
+	        assertTrue(docNo.endsWith(testOrgID + TEST_SUFFIX),
+	                "Document number should end with org ID key and suffix");
```
hengsin: Suggest to change it to "Document number should end with org ID key and suffix: "+docNo

hengsin (approved):
pass unit test

PR #2831, opened 2025-06-22
IDEMPIERE-6596 Implement $sysconfig. prefix for access to AD_SysConfi…

hengsin:
…g value in context variable expression

https://idempiere.atlassian.net/browse/IDEMPIERE-6596

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/Env.java
+++ b/org.adempiere.base/src/org/compiere/util/Env.java
@@ -139,6 +139,8 @@ public final class Env
 	public static final String USER_LEVEL = "#User_Level";
 
 	public static final String PREFIX_SYSTEM_VARIABLE = "$env.";
+	
+	public static final String PREFIX_SYSCONFIG_VARIABLE = "$sysconfig.";
 
 	@Deprecated
 	private final static ContextProvider clientContextProvider = new DefaultContextProvider();
@@ -602,6 +604,11 @@ public final class Env
 			String retValue = System.getenv(context.substring(PREFIX_SYSTEM_VARIABLE.length()));
 			if (retValue == null)
 				retValue = System.getProperty(context.substring(PREFIX_SYSTEM_VARIABLE.length()), "");
+			return retValue;
+		} else if (context.startsWith(PREFIX_SYSCONFIG_VARIABLE)) {
+			String retValue = MSysConfig.getValue(context.substring(PREFIX_SYSCONFIG_VARIABLE.length()), Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
+			if (retValue == null)
+				retValue = "";
 			return retValue;
 		}
 		String value = ctx.getProperty(context, "");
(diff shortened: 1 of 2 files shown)


org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -603,6 +605,11 @@ public static String getContext (Properties ctx, String context)
 			if (retValue == null)
 				retValue = System.getProperty(context.substring(PREFIX_SYSTEM_VARIABLE.length()), "");
 			return retValue;
+		} else if (context.startsWith(PREFIX_SYSCONFIG_VARIABLE)) {
+			String retValue = MSysConfig.getValue(context.substring(PREFIX_SYSCONFIG_VARIABLE.length()), Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
```
dpansheriya: is here good to use org from ctx? Org in context are normally used as default org. 
If organization level system configuration needs to support then organization from window context should use.

hengsin: At this stage, the windowNo variable is not available to access window context. So your suggestion is to remove the AD_Org_ID argument ?

dpansheriya: Yes, Just remove AD_Org_ID and support only client level or system level configuration.

dpansheriya: @hengsin  I think also possible to handle Org level if we add logic in 
getContext (Properties ctx, int WindowNo, int TabNo, String context)

org.adempiere.base/src/org/compiere/util/Env.java
```diff
@@ -605,11 +605,8 @@ public static String getContext (Properties ctx, String context)
 			if (retValue == null)
 				retValue = System.getProperty(context.substring(PREFIX_SYSTEM_VARIABLE.length()), "");
 			return retValue;
-		} else if (context.startsWith(PREFIX_SYSCONFIG_VARIABLE)) {
-			String retValue = MSysConfig.getValue(context.substring(PREFIX_SYSCONFIG_VARIABLE.length()), Env.getAD_Client_ID(Env.getCtx()), Env.getAD_Org_ID(Env.getCtx()));
-			if (retValue == null)
-				retValue = "";
-			return retValue;
+		} else if (isSysConfig(context)) {
+			return getSysConfigValue(context, Env.getAD_Org_ID(ctx));
```
dpansheriya, line 609: @hengsin  still here loading org level system configuration using Default org stored in context. The org selected on login screen is not working ord means user can create record in any org. So better only support here client level system configurations.
While other places where window or tab no  available, we are geting org based on window are correct.

hengsin: I think that's fine when user is not working at windows/tab. Most of the time, login org is the user's working org. At worst, it is the org that the user has access to and it is not wrong to use the config value of that org.

dpansheriya: ohk, Then I am merging this.

dpansheriya (approved):

PR #2836, opened 2025-06-26
IDEMPIERE-6600 : Posting Error For No Cost Product On Physical Inventory

uthadehikaru:
https://idempiere.atlassian.net/browse/IDEMPIERE-6600

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/acct/Doc_Inventory.java
+++ b/org.adempiere.base/src/org/compiere/acct/Doc_Inventory.java
@@ -36,7 +36,9 @@ import org.compiere.model.MInventoryLine;
 import org.compiere.model.MInventoryLineMA;
 import org.compiere.model.MProduct;
 import org.compiere.model.ProductCost;
+import org.compiere.util.DB;
 import org.compiere.util.Env;
+import org.compiere.util.Msg;
 import org.compiere.util.Util;
 
 /**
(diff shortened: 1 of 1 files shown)


hengsin:
If possible, please add unit test to help prevent future regression, thanks.

CarlosRuiz-globalqss:
I think the COUNT approach has two issues:

1 - it can be slow, and it will keep getting slower when you have more data

2 - if you need a previous zero cost, how do you manage the first one?   Does orders and invoices accept zero cost?


On the other hand, can this raise some issue about divizion by zero in other operations?   (just thinking about potential implications, would be worthy to have an end to end test about this)

jcarlisleEK:
In my situation, I hit this error during initial tenant setup to import starting inventory.  In which case, I would still hit this issue, as I do not have any orders input yet.  So I'm not sure this will solve all cases.

It looks like the suggested change is a copy of the code in [org.adempiere.base/src/org/compiere/acct/Doc_InOut.java](https://github.com/liuxin-reg/idempiere/blob/0b64948283b7894d0b6881afd2063f03019b0206/org.adempiere.base/src/org/compiere/acct/Doc_InOut.java#L184).
So if we are already doing the same thing for Doc_InOut, it seems like it should be safe to do it here?

Also, there is a flag in the [getProductCosts()](https://github.com/liuxin-reg/idempiere/blob/master/org.adempiere.base/src/org/compiere/model/ProductCost.java#L282) function called "zeroCostsOK", which sounds like a way to ensure zero costs are handles correctly, however it appears to only be used [here](https://github.com/liuxin-reg/idempiere/blob/master/org.adempiere.base/src/org/compiere/model/MCost.java#L212), which doesn't seem to do anything if the flag is false.

hengsin (approved):
looks good

PR #2839, opened 2025-06-28
IDEMPIERE-6314 Issues with attachment on System records

CarlosRuiz-globalqss:
- testing IDEMPIERE-6570 found that the "read-only" change on MAttachment is not allowing the Migrate Storage Provider to migrate attachments from tenants


CarlosRuiz-globalqss:
forced push as is a blocking bug

PR #2856, opened 2025-07-15
IDEMPIERE-6619 - Cannot enter multiple lines in fields in Grid View o…

d-ruiz:
…r Quick Form

https://idempiere.atlassian.net/browse/IDEMPIERE-6619

# Pull Request Checklist

### Tests
- [X] I checked all collaterals that can be affected by my changes, and tested other potential affected scenarios

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
@@ -46,6 +46,7 @@ import org.zkoss.zk.ui.event.EventListener;
 import org.zkoss.zk.ui.event.Events;
 import org.zkoss.zk.ui.event.KeyEvent;
 import org.zkoss.zk.ui.sys.ComponentCtrl;
+import org.zkoss.zk.ui.util.Clients;
 
 /**
  * Default editor for text display type (String, PrinterName, Text, TextLong and Memo).<br/>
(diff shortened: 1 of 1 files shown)


CarlosRuiz-globalqss:
@d-ruiz - testing this, it looks like a workaround but I don't find the result good.

The user cannot enter a newline in the middle of an existing string.

I mean, for example if I have this:

This is a long test.  Better to split it in two lines.

And I want to enter a Shift in the middle to make it

This is a long test.
Better to split it in two lines.

That's not possible, the new line is always added at the end.

The solution for that sounds a lot more tricky involving probably javascript to detect the position of the cursor in the string.

@hengsin - do you think there is maybe an easier way to achieve that?   Or is maybe a zk bug?

d-ruiz:
Thanks @CarlosRuiz-globalqss, 

I tested it mostly using QuickForm and it does not allow to move the cursor back. I updated the solution to solve the issue you found. I did not find a clean way to do it in Java, so I am using javascript instead. 

>  Or is maybe a zk bug?

The bug, IMHO, is from ZK, in GridView it captures Shift+Enter as ON_OK, which is wrong.

CarlosRuiz-globalqss (approved):
Hi @hengsin -

I tested this successfully.

Still I would like to have your peer review as you are more experienced with the zk intricacies

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WStringEditor.java
```diff
@@ -206,6 +208,36 @@ else if (gridField.getDisplayType() == DisplayType.Memo)
...
+		                	    "(function() {" +
+		                	    "  var elem = jq('#%s')[0];" +
+		                	    "  if (!elem) return;" +
+		                	    "  var start = elem.selectionStart;" +
+		                	    "  var end = elem.selectionEnd;" +
+		                	    "  var value = elem.value;" +
+		                	    "  elem.value = value.substring(0, start) + '\\n' + value.substring(end);" +
+		                	    "  elem.selectionStart = elem.selectionEnd = start + 1;" +
+		                	    "  var cmp = zk.Widget.$('$%s');" +
+		                	    "  if (cmp) cmp.fire('onChanging', {value: elem.value});" +
```
hengsin: should get widget at the start and use $n() (see https://www.zkoss.org/javadoc/7.0.3/jsdoc/zk/Widget.html) to get the dom element. Also, usually, we use zk.Widget.$('#uuid') instead. Another thing - it is using selectionStart and selectionEnd, wondering whether this would works correctly if user make real text selection before pressing shift+enter.

d-ruiz: Thanks @hengsin, just updated the script with your suggestions. 

> Another thing - it is using selectionStart and selectionEnd, wondering whether this would works correctly if user make real text selection before pressing shift+enter.

Yes - this is to behave exactly how it does natively on Form View, when the user selects a text and hit Shift+Enter, the selected text is replaced by a newline.

Please let me know if more adjustments to the script should be done

hengsin (approved):
looks good and tested by carlos

PR #2880, opened 2025-08-05
IDEMPIERE-6642 Performance issue with the PO getters on X_ classes - fix deprecate warnings

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6642

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MAsset.java
+++ b/org.adempiere.base/src/org/compiere/model/MAsset.java
@@ -242,7 +242,7 @@ public class MAsset extends X_A_Asset {
 		
 		MProduct product = MProduct.get(getCtx(), invLine.getM_Product_ID());
 		// Defaults from group:
-		MAssetGroup assetGroup = MAssetGroup.get(invLine.getCtx(), invLine.getM_Product().getM_Product_Category().getA_Asset_Group_ID());
+		MAssetGroup assetGroup = MAssetGroup.get(invLine.getCtx(), invLine.getMProduct().getMProductCategory().getA_Asset_Group_ID());
 		if (assetGroup == null)
 			assetGroup = MAssetGroup.get(invLine.getCtx(), product.getA_Asset_Group_ID());
 		setAssetGroup(assetGroup);
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -2014,7 +2014,7 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 				if (product != null
 					&& isSOTrx()
 					&& product.isCreateAsset()
-					&& !product.getM_Product_Category().getA_Asset_Group().isFixedAsset()
+					&& !product.getMProductCategory().getMAssetGroup().isFixedAsset()
 					&& sLine.getMovementQty().signum() > 0
 					&& !isReversal())
 				{
--- a/org.adempiere.base/src/org/compiere/model/MInventory.java
+++ b/org.adempiere.base/src/org/compiere/model/MInventory.java
@@ -804,7 +804,7 @@ public class MInventory extends X_M_Inventory implements DocAction
 				if(qtyDiff.compareTo(Env.ZERO)>0)
 				{
 					//AttributeSetInstance enable
-					I_M_AttributeSet as = line.getM_Product().getM_AttributeSet();
+					I_M_AttributeSet as = line.getMProduct().getM_AttributeSet();
 					if (as != null && as.isInstanceAttribute())
 					{
 						//add quantity to last attributesetinstance
--- a/org.adempiere.base/src/org/compiere/model/MInventoryLine.java
+++ b/org.adempiere.base/src/org/compiere/model/MInventoryLine.java
@@ -482,4 +482,12 @@ public class MInventoryLine extends X_M_InventoryLine
 		return getMovementQty().signum() < 0;
 	}
 	
+	/**
+	 * Get Product object from cache, no transaction 
+	 * @return
+	 */
+	public MProduct getMProduct() {
+		return MProduct.get(getCtx(), getM_Product_ID());
+	}	//	getMProduct
+
 }	//	MInventoryLine
--- a/org.adempiere.base/src/org/compiere/model/MProduct.java
+++ b/org.adempiere.base/src/org/compiere/model/MProduct.java
@@ -1201,4 +1201,13 @@ public class MProduct extends X_M_Product implements ImmutablePOSupport
 		else
 			return false;
 	}
+
+	/**
+	 * Get Product Category object from cache, no transaction 
+	 * @return
+	 */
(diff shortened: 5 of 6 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - what do you think with this approach to solve the warnings?

I created the following methods:

MInventoryLine.getMProduct()
There is already a getProduct method, but the difference is that we need to return a ZERO-ID product to mimic what the method get_M_Product was doing.

same for:
org.compiere.model.MProduct.getMProductCategory()
org.compiere.model.MProductCategory.getMAssetGroup()


So, questions:

1 - is this approach OK?

2 - is the naming convention OK?
I thought about keeping the same name from the X file, like MInventoryLine.get_M_Product() - but as this method behaves different maybe is better to do a clear distinction that is returning a MProduct object.
The X_ class is returning an I_ object which could be another defined in a plugin for example.

WDYT?

hengsin:
> Hi @hengsin - what do you think with this approach to solve the warnings?
> 
> I created the following methods:
> 
> MInventoryLine.getMProduct() There is already a getProduct method, but the difference is that we need to return a ZERO-ID product to mimic what the method get_M_Product was doing.
> 
> same for: org.compiere.model.MProduct.getMProductCategory() org.compiere.model.MProductCategory.getMAssetGroup()
> 
> So, questions:
> 
> 1 - is this approach OK?
> 
> 2 - is the naming convention OK? I thought about keeping the same name from the X file, like MInventoryLine.get_M_Product() - but as this method behaves different maybe is better to do a clear distinction that is returning a MProduct object. The X_ class is returning an I_ object which could be another defined in a plugin for example.
> 
> WDYT?

I think we can use getProduct instead and there's no need to return a ZERO-ID product to mimic what the method getM_Product was doing. The getM_Product method is deprecated so it wasn't a must to maintain 100% compatibility with it. Also, returning a new MProduct instance when M_Product_ID=0 is in fact a common source of hard to trace bug.

CarlosRuiz-globalqss:
> I think we can use getProduct instead and there's no need to return a ZERO-ID product to mimic what the method getM_Product was doing. The getM_Product method is deprecated so it wasn't a must to maintain 100% compatibility with it. Also, returning a new MProduct instance when M_Product_ID=0 is in fact a common source of hard to trace bug.

OK thanks - that makes sense.

I was just looking for an easy way to refactor that without big changes, but better to do it the right way, it will just take a bit more of time to refactor.

CarlosRuiz-globalqss:
OK @hengsin - I changed the approach - can you please check with this little sample to see if that's the correct way to go?

hengsin:
> OK @hengsin - I changed the approach - can you please check with this little sample to see if that's the correct way to go?

Looks good to me. I notice "New and existing unit tests pass locally with my changes" is not check and I think we need to make sure the changes doesn't break any unit test.

CarlosRuiz-globalqss:
> Looks good to me. I notice "New and existing unit tests pass locally with my changes" is not check and I think we need to make sure the changes doesn't break any unit test.

Yes, that's the plan, this is still in draft, I'll run tests every time I have a bigger batch of changes and change the PR to ready.

CarlosRuiz-globalqss:
Hi @hengsin 

This pull request is a first round solving the first 104 from 413 warnings.

org.adempiere.base/src/org/adempiere/base/AbstractProductPricing.java
```diff
@@ -94,7 +98,7 @@ public void setPriceDate(Timestamp priceDate) {
 	public void setOrderLine(I_C_OrderLine orderLine, String trxName) {
 		m_M_Product_ID = orderLine.getM_Product_ID();
 		if (orderLine.getC_Order_ID() > 0) {
-			I_C_Order order = orderLine.getC_Order();
+			I_C_Order order = ((MOrderLine)orderLine).getParent();
```
hengsin: why the `(MOrderLine)` casting is needed here ?

CarlosRuiz-globalqss: > why the `(MOrderLine)` casting is needed here ?

Checking the code and the calling points, it seems is not guaranteed that the orderLine here is a MOrderLine object, so I changed it to validate it.

CarlosRuiz-globalqss:
Solved 28 more warnings in last commit.

Just org.adempiere.base is left with 281 pending warnings.

org.adempiere.base/src/org/adempiere/base/AbstractProductPricing.java
```diff
@@ -94,7 +98,14 @@ public void setPriceDate(Timestamp priceDate) {
...
 		m_M_Product_ID = orderLine.getM_Product_ID();
 		if (orderLine.getC_Order_ID() > 0) {
-			I_C_Order order = orderLine.getC_Order();
+			MOrder order;
+			if (orderLine instanceof MOrderLine oLine) {
+				order = oLine.getParent();
+			} else {
+				// Fallback without cache
+				order = new MOrder(Env.getCtx(), orderLine.getC_Order_ID(), trxName);
+				m_isSOTrx = order.isSOTrx();
```
hengsin: this line is redundant

org.adempiere.base/src/org/adempiere/base/AbstractProductPricing.java
```diff
@@ -150,7 +161,7 @@ public void setRequisitionLine(I_M_RequisitionLine reqLine, String trxName) {
 	public void setRMALine(I_M_RMALine rmaLine, String trxName) {
 		m_M_Product_ID = rmaLine.getM_Product_ID();
 		if (rmaLine.getM_RMA_ID() > 0) {
-			I_M_RMA rma = rmaLine.getM_RMA();
+			I_M_RMA rma = ((MRMALine)rmaLine).getParent();
```
hengsin: the three set methods have the same issue with the I_C_OrderLine above, I guess we should just implement the same pattern here. For me, if we wants to enforce it is using M* class (MOrderLine, MProjectLine, etc), that should be enforce at the method parameter level instead of using I* parameter and will throw class cast exception if it is not an instance of M* class.

org.adempiere.base/src/org/adempiere/base/AbstractProductPricing.java
CarlosRuiz-globalqss: thanks - fixed

org.adempiere.base/src/org/adempiere/base/AbstractProductPricing.java
CarlosRuiz-globalqss: ready - implemented same approach in the three methods

CarlosRuiz-globalqss:
Hi @hengsin 

I pushed more warning fixes in last commit.

It would be good if we can move this ahead and I'll open new pull requests for the pending, that way the number of warnings decrease for everybody, and the peer review maybe becomes also easier for you

Regards,

Carlos Ruiz

PR #2894, opened 2025-08-26
IDEMPIERE-6661: Schedulers : reuse sessions

nmicoud:
https://idempiere.atlassian.net/browse/IDEMPIERE-6661

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
+++ b/org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
@@ -145,7 +145,7 @@ public class Scheduler extends AdempiereServer
 			//Create new Session and set #AD_Session_ID to context
 			
 			// Examples : @#Date@ / Forever / @AD_Scheduler_ID@
-			String sessionIdFormat = "@AD_Scheduler_ID@"; // TODO add column on AD_Scheduler
+			String sessionIdFormat = "@AD_Scheduler_ID@/@#Date@"; // TODO add column on AD_Scheduler
 			String parsedSessionId = "";
 			if (!Util.isEmpty(sessionIdFormat)) {
 
@@ -153,15 +153,18 @@ public class Scheduler extends AdempiereServer
 
 				parsedSessionId = Env.parseVariable(sessionIdFormat, new DefaultEvaluatee(), true, false);
 
-				StringBuilder whereClause = new StringBuilder("CreatedBy = ? AND ServerName = ? AND WebSession = ?");
+				StringBuilder whereClause = new StringBuilder("CreatedBy = ? AND AD_Org_ID = ? AND AD_Role_ID = ? AND ServerName = ? AND WebSession = ?");
 				ArrayList<Object> params = new ArrayList<Object>();
 				params.add(getAD_User_ID());
+				params.add(Env.getAD_Org_ID(getCtx()));
+				params.add(Env.getAD_Role_ID(getCtx()));
 				params.add(WebUtil.getServerName());
-				params.add(parsedSessionId);
+				params.add(getWebSessionLabel(parsedSessionId));
 
 				Query query = new Query(Env.getCtx(), MSession.Table_Name, whereClause.toString(), null)
 						.setParameters(params)
-						.setClient_ID();
+						.setClient_ID()
+						.setOrderBy("Updated DESC");
 
 				int oldSessionID = query.firstId();
 				if (oldSessionID > 0)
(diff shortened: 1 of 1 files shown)


org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
```diff
@@ -140,11 +143,46 @@ protected void doWork ()
...
+			// Examples : @#Date@ / Forever / @AD_Scheduler_ID@
+			String sessionIdFormat = "@AD_Scheduler_ID@"; // TODO add column on AD_Scheduler
+			String parsedSessionId = "";
+			if (!Util.isEmpty(sessionIdFormat)) {
+
+				Env.setContext(getCtx(), "AD_Scheduler_ID", AD_Scheduler_ID); // TODO use ENV.AD_Scheduler_ID ?
+
+				parsedSessionId = Env.parseVariable(sessionIdFormat, new DefaultEvaluatee(), true, false);
+
+				StringBuilder whereClause = new StringBuilder("CreatedBy = ? AND ServerName = ? AND WebSession = ?");
```
CarlosRuiz-globalqss: I just made a test and the sessions created by Schedulers can have Organization.
And also they have a Role assigned (not sure which one when the user have several, maybe the first?)

Anyways, it would be better to search for the session with same org, and role if possible.

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
```diff
@@ -166,7 +204,7 @@ protected void doWork ()
 				if (m_trx != null)
 					m_trx.close();
 				m_trx = null;
-				session.logout();
+				session.logout(); // TODO only if there is no other schedulers running in parallel ?
```
CarlosRuiz-globalqss: not sure about this - maybe we can do logout just if the new column is empty (so, no reuse of session).

If the session is reusable, then we can keep the session open.

Also, please check for example com.trekglobal.idempiere.rest.api.v1.auth.filter.RequestFilter.validate:241
in that code, if the session is closed (because of a server reboot), it just marks it as reopened and updates the Updated time

So, these sessions will be marked as closed just no server reboot (same as it's happening with idempiere-rest sessions).

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
```diff
@@ -140,11 +143,46 @@ protected void doWork ()
...
+				int oldSessionID = query.firstId();
+				if (oldSessionID > 0)
+					Env.setContext(getCtx(), Env.AD_SESSION_ID, oldSessionID);
+			}
+
 			MSession session = MSession.get(Env.getCtx());
 			if(session == null) {
 				session = MSession.create(Env.getCtx());
+				if (!Util.isEmpty(parsedSessionId)) {
+					session.setWebSession(parsedSessionId);
```
CarlosRuiz-globalqss: I think we also consider adding a prefix that makes clear these sessions are created from Schedulers.

For example the WebSession could be something like:
"Sched-[scheduler-ID]-" - and then the parsed session.

I think we need also to increase the length of the WebSession column, at this moment is just 40 chars.

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
nmicoud: done

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
nmicoud: done

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
nmicoud: Hi @CarlosRuiz-globalqss 

Done, but while doing it, I was wondering if it would be better to do the following:
- rename _sessionIdFormat_ column to _WebSessionLabel_ (or something like that)
- if the _WebSessionLabel_ is empty, the we force it to "Scheduler-[" + AD_Scheduler_ID + "]"
- if is filled, AD_Session.WebSession = WebSessionLabel (parsed)
wdyt?

Since I don't know which option to go with, I haven't created any migration scripts.

nb: I also tried to add a format on the date (at the moment, @#Date@ returns '2025-11-01 00:00:00' ; so I tried to use _@#Date<dd/MM/yyyy>@_. But from what I see, is ignored on org.compiere.util.DefaultEvaluatee.get_ValueAsString(Properties, String) as it is a globalVariable, then dataValue is null. I didn't find a nice way to fix that (maybe @hengsin have you an idea?)

Regards,

Nicolas

d-ruiz:
Hi @nmicoud,

**Quick status check:**
- [x] Still working on this and plan to complete soon? 
- [ ] Ready to close for now?

nmicoud:
is still in progress; since this may cause disruptions, it has been agreed to resume after v13

org.adempiere.server/src/main/server/org/compiere/server/Scheduler.java
CarlosRuiz-globalqss: Hi @nmicoud 

> * rename _sessionIdFormat_ column to _WebSessionLabel_ (or something like that)

no problem

>     * if the _WebSessionLabel_ is empty, the we force it to "Scheduler-[" + AD_Scheduler_ID + "]"
>     * if is filled, AD_Session.WebSession = WebSessionLabel (parsed)

not sure, that would make this different than current behavior, what if somebody wants to create a session per each execution?

Maybe you can define the default of _WebSessionLabel_ to be that by default, and people can clean it if they want it different.


> nb: I also tried to add a format on the date (at the moment, @#Date@ returns '2025-11-01 00:00:00' 

I would not worry about that - as the goal is achieved, to have a session per date.

nmicoud:
Closing this one, and doing changes here: https://github.com/idempiere/idempiere/pull/3204

PR #2896, opened 2025-08-28
Idempiere 6663

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-6663

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/adempiere/base/Core.java
+++ b/org.adempiere.base/src/org/adempiere/base/Core.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
 
+import javax.crypto.SecretKey;
 import javax.script.ScriptEngine;
 import javax.script.ScriptEngineFactory;
 import javax.script.ScriptEngineManager;
@@ -283,10 +284,19 @@ public class Core {
 			if (keystoreService != null)
 				return keystoreService;
 		}
-		IServiceReferenceHolder<IKeyStore> serviceReference = Service.locator().locate(IKeyStore.class).getServiceReference();
-		if (serviceReference != null) {
-			keystoreService = serviceReference.getService();
-			s_keystoreServiceReference = serviceReference;
+		
+		IServicesHolder<IKeyStore> list = Service.locator().list(IKeyStore.class);
+		List<IServiceReferenceHolder<IKeyStore>> references = list.getServiceReferences();
+		for (IServiceReferenceHolder<IKeyStore> ref : references) {
+			IKeyStore service = ref.getService();
+			if (service != null) {
+				SecretKey key = service.getKey(0);
+				if (key != null) {
+					keystoreService = service;
+					s_keystoreServiceReference = ref;
+					break;
+				}
+			}
 		}
 		return keystoreService;
 	}
--- a/org.adempiere.base/src/org/adempiere/base/Core.java
+++ b/org.adempiere.base/src/org/adempiere/base/Core.java
@@ -290,6 +290,7 @@ public class Core {
 		for (IServiceReferenceHolder<IKeyStore> ref : references) {
 			IKeyStore service = ref.getService();
 			if (service != null) {
+				//test key store service is working
 				SecretKey key = service.getKey(0);
 				if (key != null) {
 					keystoreService = service;
--- a/org.adempiere.base/src/org/compiere/util/CLogMgt.java
+++ b/org.adempiere.base/src/org/compiere/util/CLogMgt.java
@@ -596,7 +596,9 @@ public class CLogMgt
 		sb.append("Free Heap = "+formatMemoryInfo(runtime.freeMemory())).append(NL);
 		//
 		//thread info
-		sb.append("Active Threads = " + Thread.activeCount());
+		sb.append("Active Threads = " + Thread.activeCount()).append(NL);
+		// Key Store info
+		sb.append("Key Store = ").append(Core.getKeyStore().getClass().getName());
 		//
 		//cluster info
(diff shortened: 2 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin 

I tested this one successfully

Do you think it would be a good idea to add a warning like this:
[pr2896PR0.patch.txt](https://github.com/user-attachments/files/22021602/pr2896PR0.patch.txt)

CarlosRuiz-globalqss:
Hi @hengsin 

One more change suggested:
[pr2896PR1.patch.txt](https://github.com/user-attachments/files/22023703/pr2896PR1.patch.txt)

I think is better to log when a keystore plugin is not working - most probable not properly configured.

hengsin:
> Hi @hengsin
> 
> One more change suggested: [pr2896PR1.patch.txt](https://github.com/user-attachments/files/22023703/pr2896PR1.patch.txt)
> 
> I think is better to log when a keystore plugin is not working - most probable not properly configured.

The assumption for null here is it is not configure rather than there's any error. If there's any configuration error, that should be logged by the key store service. Would add it as info log instead for not configure service.

CarlosRuiz-globalqss:
> The assumption for null here is it is not configure rather than there's any error. If there's any configuration error, that should be logged by the key store service. Would add it as info log instead for not configure service.

Hi @hengsin - I disagree, I don't see a reason why would somebody install a keystore plugin as default to not use it.

And it sounds very dangerous to run the server if the configuration was lost for some reason, or changing the keystore plugin.
They cannot read encrypted data with another plugin, and if they write a new value it would create a mix of encrypted data in their database that can be irrecoverable.

Indeed such scenario makes me think if a simple warning is not enough.

Maybe we can combine this with the ["mandatory plugins"](https://wiki.idempiere.org/en/NF12_Validate_Required_Plugins) to stop the server if the keystore plugin is not properly configured.

hengsin:
> > The assumption for null here is it is not configure rather than there's any error. If there's any configuration error, that should be logged by the key store service. Would add it as info log instead for not configure service.
> 
> Hi @hengsin - I disagree, I don't see a reason why would somebody install a keystore plugin as default to not use it.
> 
> And it sounds very dangerous to run the server if the configuration was lost for some reason, or changing the keystore plugin. They cannot read encrypted data with another plugin, and if they write a new value it would create a mix of encrypted data in their database that can be irrecoverable.
> 
> Indeed such scenario makes me think if a simple warning is not enough.
> 
> Maybe we can combine this with the ["mandatory plugins"](https://wiki.idempiere.org/en/NF12_Validate_Required_Plugins) to stop the server if the keystore plugin is not properly configured.

Regardless of what we do here with the key store service here, you can always use the mandatory plugins to stop the server, that's not really related.

If that's consider really dangerous, we should simply stop the application if the top priority key store service is not working rather than doing any logging.

CarlosRuiz-globalqss:
> Regardless of what we do here with the key store service here, you can always use the mandatory plugins to stop the server, that's not really related.
> 
> If that's consider really dangerous, we should simply stop the application if the top priority key store service is not working rather than doing any logging.

But there are two cases here:

- the mandatory plugins stop the server when the plugin does not exist, but here the plugin exists but is not or mis-configured

- and we have the case of installing two keystores from a p2 repository, and we could want to use the first keystore in one system, and the second keystore in another system

That's why I tend to think we would need to combine both things - 
mandatory+misconfigured -> stop the server
not mandatory -> in this case maybe a log.info is enough as you suggested

hengsin:
> > Regardless of what we do here with the key store service here, you can always use the mandatory plugins to stop the server, that's not really related.
> > If that's consider really dangerous, we should simply stop the application if the top priority key store service is not working rather than doing any logging.
> 
> But there are two cases here:
> 
>     * the mandatory plugins stop the server when the plugin does not exist, but here the plugin exists but is not or mis-configured
> 
>     * and we have the case of installing two keystores from a p2 repository, and we could want to use the first keystore in one system, and the second keystore in another system
> 
> 
> That's why I tend to think we would need to combine both things - mandatory+misconfigured -> stop the server not mandatory -> in this case maybe a log.info is enough as you suggested

Ok, we can add the mandatory checking.

hengsin:
Also, do note that you can also use "components.blacklist" (https://idempiere.atlassian.net/browse/IDEMPIERE-1224) to disable a key store service.

CarlosRuiz-globalqss:
> Also, do note that you can also use "components.blacklist" (https://idempiere.atlassian.net/browse/IDEMPIERE-1224) to disable a key store service.

ah, good idea for our use case

CarlosRuiz-globalqss:
> Also, do note that you can also use "components.blacklist" (https://idempiere.atlassian.net/browse/IDEMPIERE-1224) to disable a key store service.

Hi @hengsin 

If it is difficult to make it work with "mandatory plugin" - because we have the classname and not the plugin name - I think maybe we can do what you suggested:
"If that's consider really dangerous, we should simply stop the application if the top priority key store service is not working rather than doing any logging."

and the solution for the second case I described is simply to blacklist the component.

hengsin:
> > Also, do note that you can also use "components.blacklist" (https://idempiere.atlassian.net/browse/IDEMPIERE-1224) to disable a key store service.
> 
> Hi @hengsin
> 
> If it is difficult to make it work with "mandatory plugin" - because we have the classname and not the plugin name - I think maybe we can do what you suggested: "If that's consider really dangerous, we should simply stop the application if the top priority key store service is not working rather than doing any logging."
> 
> and the solution for the second case I described is simply to blacklist the component.

There's a semantic mismatch here - the mandatory plugin is for a bundle/plugin but the key store service is an OSGI component. A bundle/plugin can have more than one components.

For now, I will just throw exception if the top priority key store service is not working.

PR #2903, opened 2025-09-03
IDEMPIERE-6670 Refactoring: make CLogMgt.reInit public

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-6670

# Pull Request Checklist

### Tests
### Documentation

### DIFF
--- a/org.adempiere.base/src/org/compiere/util/CLogMgt.java
+++ b/org.adempiere.base/src/org/compiere/util/CLogMgt.java
@@ -69,7 +69,7 @@ public class CLogMgt
 	/**
 	 * Re-initialize after log configuration change. 
 	 */
-	private static synchronized void reInit() {
+	public static synchronized void reInit() {
 		CLogMgt.initialize(Ini.isClient());
 		if (!levelMap.isEmpty()) {
 			loggerMap.clear();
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - there is a bug not caused by this pull request - it was present before and it continues there.

Pushing the "Reload Log Props" in test.idempiere.org is throwing error:
not a hexadecimal digit: "S" = 83

I'll try to check later, there is a discrepancy how the properties are read the first time vs the reload.

CarlosRuiz-globalqss (changes requested):
Hi @hengsin 

After tested and analyzed this case I think the following patch is needed:
[pr2903PR0.patch.txt](https://github.com/user-attachments/files/22393373/pr2903PR0.patch.txt)

* it fixes the existing bug about reloading properties not encrypted
* it adds comments in Adempiere.java and AboutWindow.java about the fact that .TraceLevel properties are not encrypted
* it removes the call to CLogMgt.reInit() at the end - it is not necessary as all the levels are being set before, and indeed calling reInit is causing that a log level cannot be removed from the properties

So, making the reInit method public will work fine for our LogEventHandler case, but is not necessary for AboutWindow - this will just keep the refactoring you made and the bug fix.

Do you agree?

PR #2918, opened 2025-09-23
IDEMPIERE-6122 Query class to accept a list of columns to select - fix support order by virtual column not in select

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6122?focusedCommentId=59847

# Pull Request Checklist

### Tests

### DIFF
--- a/org.idempiere.test/src/org/idempiere/test/DictionaryIDs.java
+++ b/org.idempiere.test/src/org/idempiere/test/DictionaryIDs.java
@@ -141,6 +141,17 @@ public final class DictionaryIDs {
 		}
 	}
 	
+	public enum C_BankAccount_Processor {
+		MONEYBANK_1234("f4a64026-bf68-4c8c-b238-8cdf006aae04"),
+		MONEYBANK_5678("f8f892f0-36ab-4b4d-9dd3-c3bbe12cf455");
+
+		public final String uuid;
+
+		private C_BankAccount_Processor(String uuid) {
+			this.uuid = uuid;
+		}
+	}
+
 	public enum C_BPartner {
 		STANDARD(112,"dc4962ad-066f-4ebf-88ac-545d12c9f1f0"),
 		TREE_FARM(114,"5c557ed6-c443-4b99-ae77-546ccd46bc1f"),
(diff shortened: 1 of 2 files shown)


CarlosRuiz-globalqss:
Hi @hengsin - I changed this back to draft because of [the second finding](https://idempiere.atlassian.net/browse/IDEMPIERE-6122?focusedCommentId=59848)

> Same issue found about expand columns, a column cannot be expanded if not included in the $select

Because of that maybe is better to do the fix in the REST plugin instead of core?

WDYT?

CarlosRuiz-globalqss:
OK, after analyzed better both cases:

The issue with the order by a virtual column must be solved in core - because is something that can happen anywhere, not just in REST.
So, this pull request is valid and solves that problem.

the $expand issue must be solved in the REST layer.

org.adempiere.base/src/org/compiere/model/Query.java
```diff
@@ -856,6 +856,24 @@ private final String buildSQL(StringBuilder selectClause, boolean useOrderByClau
...
+						}
+						if (!found) {
+							// add vc to array selectColumns
+							String[] tmp = new String[selectColumns.length + 1];
+							System.arraycopy(selectColumns, 0, tmp, 0, selectColumns.length);
+							tmp[selectColumns.length] = vc;
+							selectColumns = tmp;
+						}
+					}
+				}
```
hengsin, line 876: isn't this should be a check on the order by columns instead of virtual columns ? If a table has column A, B and C, selectColumns A, B and Order By C, would it work ?

CarlosRuiz-globalqss: > isn't this should be a check on the order by columns instead of virtual columns ? If a table has column A, B and C, selectColumns A, B and Order By C, would it work ?

Is the same as in postgres, you can order by columns not in the select clause, but you cannot order by "virtual columns"

So, in postgres (and in REST) this is valid:
```sql
SELECT Name FROM C_Charge ORDER BY C_Charge_ID   /* Valid */
```

But the query of virtual columns must be something like this:
```sql
SELECT Name, UPPER(Description) as UpperDescription FROM C_Charge ORDER BY UpperDescription   /* Valid */
```

But, because the virtual column is not in the select columns the query produced is this:
```sql
SELECT Name FROM C_Charge ORDER BY UpperDescription   /* Error */
```

PR #2919, opened 2025-09-25
IDEMPIERE-6683 Concurrent client requests can cause "key already exists" errors for user preference

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6683

# Pull Request Checklist

### Tests

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MUserPreference.java
+++ b/org.adempiere.base/src/org/compiere/model/MUserPreference.java
@@ -27,6 +27,7 @@ package org.compiere.model;
 import java.sql.ResultSet;
 import java.util.Properties;
 
+import org.adempiere.exceptions.DBException;
 import org.compiere.util.CacheMgt;
 import org.compiere.util.Env;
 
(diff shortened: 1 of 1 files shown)


PeterTakacs300:
Hi @CarlosRuiz-globalqss, I tested the fix with rest. The re-query works fine, but the `org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "ad_userpreference_user_idx"` error is still logged in the console. Could that be avoided? It can be misleading and creates unnecessary logs (in prod server for one tenant we had cca 150 of these during 3 days)

CarlosRuiz-globalqss:
> Hi @CarlosRuiz-globalqss, I tested the fix with rest. The re-query works fine, but the `org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "ad_userpreference_user_idx"` error is still logged in the console. Could that be avoided? It can be misleading and creates unnecessary logs (in prod server for one tenant we had cca 150 of these during 3 days)

In principle it doesn't seem easy to hide that message.

That sounds odd, is your application creating users on the fly?  Maybe when creating the user you can also create the preference record and it will avoid this problem.

CarlosRuiz-globalqss:
Merging as tested successfully by me in zkwebui and @PeterTakacs300 in REST.

PR #3131, opened 2026-02-24
IDEMPIERE-6769 Enabling "Set as Default" option incorrectly sets default for all users when "Share for all users" is enabled

CarlosRuiz-globalqss:
https://idempiere.atlassian.net/browse/IDEMPIERE-6769

# Pull Request Checklist

### Tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced default user query initialization to apply only queries belonging to the current user or global queries created by the current user, preventing incorrect query assignments.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/adwindow/ADWindowToolbar.java
@@ -1569,9 +1569,13 @@ public class ADWindowToolbar extends ToolBar implements EventListener<Event>
 	 */
 	public boolean initDefaultQuery() {
 		if(userQueries != null) {
+    		int currentUserId = Env.getAD_User_ID(Env.getCtx());
 	        for (int i = 0; i < userQueries.length; i++) {
-	        	if(userQueries[i].isDefault()) {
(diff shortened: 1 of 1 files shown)


d-ruiz:
Hi @CarlosRuiz-globalqss, 

Is the desired behavior that an Admin user, for instance, cannot set a default query for everyone? Or that a user cannot set a global query as the default for hiimself?

CarlosRuiz-globalqss:
Hmm - it sounds tricky:

> Is the desired behavior that an Admin user, for instance, cannot set a default query for everyone?

No, that's something that we must be able to offer.

> Or that a user cannot set a global query as the default for hiimself?

In current code (before this patch), a user cannot do that.  To do that it would need to save the query with another name and set it as default.

But it sounds tricky to have both worlds - if a user can change the query to "default" with the actual code is changing that for everyone.
That's something we would like to allow for Admins, but not for any user.


So, let's keep this in draft and design how to solve the Admin case - OK?

hengsin (changes requested):
please address the coderabbitai review comments.

CarlosRuiz-globalqss:
> please address the coderabbitai review comments.

Hi @hengsin  - I didn't fix that suggestion because it was not changed by me and all the tests worked fine as is.  So, I think it needs deeper testing to see if is a valid suggestion

PR #3196, opened 2026-03-16
IDEMPIERE-6903: InvoiceCreditMemo process: should not block if previo…

nmicoud:
…us credit memo is reversed

https://idempiere.atlassian.net/browse/IDEMPIERE-6903

# Pull Request Checklist

### Tests
### Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Credit memo handling improved: when voiding or reversing credit memos, the related-invoice link is now cleared to prevent stale or incorrect associations and downstream reconciliation issues.
* **Chores**
  * A database migration is included to update existing voided/reversed credit memo records so their related-invoice links are cleared consistently.

### DIFF
--- a/migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
+++ b/migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
@@ -1,0 +1,5 @@
+-- IDEMPIERE-6903: InvoiceCreditMemo process: should not block if previous credit memo is reversed or void
+SELECT register_migration_script('202603231415_IDEMPIERE-6903.sql') FROM dual;
+
+UPDATE C_Invoice SET RelatedInvoice_ID = NULL WHERE C_DocTypeTarget_ID IN (SELECT C_DocType_ID FROM C_DocType WHERE docbasetype = 'ARC') AND DocStatus IN ('VO', 'RE')
+;
--- a/migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
+++ b/migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
@@ -1,5 +1,16 @@
 -- IDEMPIERE-6903: InvoiceCreditMemo process: should not block if previous credit memo is reversed or void
 SELECT register_migration_script('202603231415_IDEMPIERE-6903.sql') FROM dual;
 
-UPDATE C_Invoice SET RelatedInvoice_ID = NULL WHERE C_DocTypeTarget_ID IN (SELECT C_DocType_ID FROM C_DocType WHERE docbasetype = 'ARC') AND DocStatus IN ('VO', 'RE')
+UPDATE C_Invoice i
+SET RelatedInvoice_ID = NULL
+WHERE EXISTS (
+    SELECT 1
+    FROM C_DocType dt
+    WHERE dt.DocBaseType IN ('ARC', 'APC')
+    AND (
+        (i.DocStatus = 'VO' AND dt.C_DocType_ID = i.C_DocTypeTarget_ID)
+        OR
+        (i.DocStatus = 'RE' AND dt.C_DocType_ID = i.C_DocType_ID)
+    )
+)
 ;
--- a/org.adempiere.base/src/org/compiere/model/MInvoice.java
+++ b/org.adempiere.base/src/org/compiere/model/MInvoice.java
@@ -2558,6 +2558,9 @@ public class MInvoice extends X_C_Invoice implements DocAction, IDocsPostProcess
 				return reverseCorrectIt();
 		}
 
+		if (isCreditMemo())
+			setRelatedInvoice_ID(0);
+
 		// After Void
 		m_processMsg = ModelValidationEngine.get().fireDocValidate(this,ModelValidator.TIMING_AFTER_VOID);
 		if (m_processMsg != null)
@@ -2642,6 +2645,7 @@ public class MInvoice extends X_C_Invoice implements DocAction, IDocsPostProcess
 		
 		//
 		reverseAllocations(accrual, getC_Invoice_ID());
+
 		//	Reverse/Delete Matching
 		if (!isSOTrx())
 		{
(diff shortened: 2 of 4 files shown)


CarlosRuiz-globalqss (changes requested):
checking DocStatus=DR is incomplete - there are other statuses that represent a Draft

Another approach for this (maybe better) could be to remove the RelatedInvoice_ID relationship when reversing the credit memo

nmicoud:
> checking DocStatus=DR is incomplete - there are other statuses that represent a Draft
> 
> Another approach for this (maybe better) could be to remove the RelatedInvoice_ID relationship when reversing the credit memo

Hi @CarlosRuiz-globalqss 

You're right, that would be more consistent.
But I think that should also be added in the voidIt method, right? (as the credit memo can be created in draft).
I don't think reactivateIt method should be changed as once the CM is InProgress, it can be deleted or void.

wdyt?

CarlosRuiz-globalqss:
> You're right, that would be more consistent. But I think that should also be added in the voidIt method, right? (as the credit memo can be created in draft). I don't think reactivateIt method should be changed as once the CM is InProgress, it can be deleted or void.
> wdyt?

Yes, voidIt also, BTW, can it be deleted?  Maybe that's blocked because of foreign key?

nmicoud:
> > You're right, that would be more consistent. But I think that should also be added in the voidIt method, right? (as the credit memo can be created in draft). I don't think reactivateIt method should be changed as once the CM is InProgress, it can be deleted or void.
> > wdyt?
> 
> Yes, voidIt also, BTW, can it be deleted? Maybe that's blocked because of foreign key?

noted for voidIt.

The CM can be deleted, but the invoice cannot (as is completed/closed) - or am I missing something

CarlosRuiz-globalqss:
> The CM can be deleted, but the invoice cannot (as is completed/closed) - or am I missing something

I mean, have you tried to delete it?  Is it possible?  Or the database foreign key restricts it?

nmicoud:
> > The CM can be deleted, but the invoice cannot (as is completed/closed) - or am I missing something
> 
> I mean, have you tried to delete it? Is it possible? Or the database foreign key restricts it?



> > The CM can be deleted, but the invoice cannot (as is completed/closed) - or am I missing something
> 
> I mean, have you tried to delete it? Is it possible? Or the database foreign key restricts it?

The Delete button is inactive, so is not possible. Or do you mean in another way?
image

CarlosRuiz-globalqss:
> The Delete button is inactive, so is not possible. Or do you mean in another way? 

OK - this is what I tested in test.idempiere.org

- create sales invoice and complete
- create a credit memo using the process and complete it
- configure AR Credit Memo to allow reactivation
- reactivate the credit memo
- delete the credit memo

it's fine then, let's focus on the reverse/void stuff

nmicoud:
Hi @CarlosRuiz-globalqss 
I've restored the query from InvoiceCreateCreditMemo process and add some code on reverse/voidIt methods.

org.adempiere.base/src/org/compiere/model/MInvoice.java
```diff
@@ -3658,5 +3665,5 @@ private String updateProjectInvoiceAmt(boolean negateAmt) {
 
 		return "";
 	}
-	
+
```
hengsin, line 3668: I think we should have a migration script that clean up existing data,

nmicoud: Added

nmicoud:
Unit tests passed

hengsin:
> Unit tests passed

hi @nmicoud , the protocol is to update the pull request check list for unit test (New and existing unit tests pass locally with my changes).

nmicoud:
> > Unit tests passed
> 
> hi @nmicoud , the protocol is to update the pull request check list for unit test (New and existing unit tests pass locally with my changes).

hi @hengsin 
done

migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
```diff
@@ -0,0 +1,5 @@
+-- IDEMPIERE-6903: InvoiceCreditMemo process: should not block if previous credit memo is reversed or void
+SELECT register_migration_script('202603231415_IDEMPIERE-6903.sql') FROM dual;
+
+UPDATE C_Invoice SET RelatedInvoice_ID = NULL WHERE C_DocTypeTarget_ID IN (SELECT C_DocType_ID FROM C_DocType WHERE docbasetype = 'ARC') AND DocStatus IN ('VO', 'RE')
```
hengsin: although it should be the same but is is safer to use C_DocType_ID instead.

hengsin (changes requested):

migration/iD13/postgresql/202603231415_IDEMPIERE-6903.sql
nmicoud: not sure as if the credit memo is still in draft, C_DocType_ID will be 0, no?

hengsin: > not sure as if the credit memo is still in draft, C_DocType_ID will be 0, no?

ah ok so should be C_DocType_ID for RE and C_DocTypeTarget_ID for VO

nmicoud: wouldn't it be easier to read the query to use C_DocTypeTarget_ID for all records?
I mean, no need to have a 'complex' where clause here or do you think is necessary?

hengsin: semantic wise, C_DoctypeTarget_ID is only valid for draft document.

nmicoud: hi @hengsin 
Done

hengsin (approved):
looks good

hengsin:
hi @CarlosRuiz-globalqss , this looks good for integration.

PR #3250, opened 2026-04-27
IDEMPIERE-6981 Multi-select columns are not interpreted in cube reports.

VitorHenri:
[ IDEMPIERE-6981 ](https://idempiere.atlassian.net/browse/IDEMPIERE-6981)

# Pull Request Checklist

### Tests
### Documentation

## Summary by CodeRabbit

* **New Features**
  * Printed reports now aggregate and display multiple-selected lookup values, including translated names when not using the base language.
* **Bug Fixes**
  * Improved consistency of grouping, ordering and display for multi-selection fields across database types so all selections appear correctly in print output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/idempiere/idempiere/pull/3250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/org.adempiere.base/src/org/compiere/print/DataEngine.java
+++ b/org.adempiere.base/src/org/compiere/print/DataEngine.java
@@ -520,59 +520,139 @@ public class DataEngine
 						lookupSQL = ColumnSQL;
 
 					String display = ColumnName;
+					boolean isOracle = DB.isOracle();
 
+					String lookupValue;
+					String csvValuesSql;
+
+					if (isOracle) {
+						lookupValue = "NVL(TO_CHAR(" + lookupSQL + "), '')";
+
+						csvValuesSql =
+							"(select trim(regexp_substr(replace(" + lookupValue + ", ' ', ''), '[^,]+', 1, level)) " +
+							"   from dual " +
+							" connect by regexp_substr(replace(" + lookupValue + ", ' ', ''), '[^,]+', 1, level) is not null)";
+					} else {
+						lookupValue = "NVL(" + lookupSQL + "::text,'')";
+						csvValuesSql = "string_to_array(replace(" + lookupValue + ", ' ', ''), ',')";
+					}
+
+					// Multi seleção baseada em lista
 					if (DisplayType.isList(AD_Reference_ID) || AD_Reference_Value_ID <= 0) {
 						String eSql;
 
 						if (Env.isBaseLanguage(m_language, "AD_Ref_List")) {
-							eSql = "(select string_agg(rl.Name, ', ' order by rl.Name) " + " from AD_Ref_List rl "
-									+ " where rl.AD_Reference_ID=" + AD_Reference_Value_ID
-									+ "   and rl.Value = any(string_to_array(replace(NVL(" + lookupSQL
-									+ "::text,''),' ',''), ',')))";
+							if (isOracle) {
+								eSql =
+									"(select LISTAGG(rl.Name, ', ') WITHIN GROUP (ORDER BY rl.Name) " +
+									"   from AD_Ref_List rl " +
+									"  where rl.AD_Reference_ID = " + AD_Reference_Value_ID +
+									"    and rl.Value in " + csvValuesSql +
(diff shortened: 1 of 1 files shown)


org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -514,6 +514,148 @@ else if (AD_Reference_ID == DisplayType.Table
...
+						csvValuesSql =
+							"(select trim(regexp_substr(replace(" + lookupValue + ", ' ', ''), '[^,]+', 1, level)) " +
+							"   from dual " +
+							" connect by regexp_substr(replace(" + lookupValue + ", ' ', ''), '[^,]+', 1, level) is not null)";
+					} else {
+						lookupValue = "NVL(" + lookupSQL + "::text,'')";
+						csvValuesSql = "string_to_array(replace(" + lookupValue + ", ' ', ''), ',')";
+					}
+
+					// Multi seleção baseada em lista
```
hengsin: please use english only for in code comments.

org.adempiere.base/src/org/compiere/print/DataEngine.java
```diff
@@ -514,6 +514,148 @@ else if (AD_Reference_ID == DisplayType.Table
...
+									"(select string_agg(" + translatedName + ", ', ' order by " + translatedName + ") " +
+									"   from AD_Ref_List rl " +
+									"   left join AD_Ref_List_Trl trl " +
+									"     on trl.AD_Ref_List_ID = rl.AD_Ref_List_ID " +
+									"    and trl.AD_Language = '" + m_language.getAD_Language() + "' " +
+									"  where rl.AD_Reference_ID = " + AD_Reference_Value_ID +
+									"    and rl.Value = any(" + csvValuesSql + ")" +
+									")";
+							}
+						}
```
hengsin, line 583: please use DB.inFilterForCSV instead to prevent potential SQL injection attack.

hengsin (changes requested):
please address coderabbitai review comments.

org.adempiere.base/src/org/compiere/print/DataEngine.java
VitorHenri: Hi @hengsin, I think DB.inFilterForCSV() may not fit directly in this case because the CSV values are generated dynamically by the SQL expression itself (string_to_array(...) in PostgreSQL / regexp_substr(...) in Oracle), rather than being available as Java-side parameters during query construction.

My interpretation is that the values extracted from csvValuesSql are treated by the database as data elements of the array/subquery, not as executable SQL, since we are not concatenating the CSV contents directly into the SQL structure itself.

However, I may be missing something here. If you believe there is still a possible injection scenario in this approach, could you please clarify the expected usage of DB.inFilterForCSV() in this context?

hengsin: hmm .. if you use `DB.inFilterForCSV()`, you will pass `lookupValue` to the method and drop the use of `string_to_array(...)`

VitorHenri: Hi @hengsin, I think there may be a misunderstanding about what lookupValue represents in this implementation.

lookupValue is not the actual CSV value available on the Java side (e.g. "A,B,C"). It is currently part of a SQL expression that is evaluated by the database at runtime, and only then produces the individual values.

Because of that, I don't currently have the CSV content available when the SQL is being assembled, so I cannot directly replace the string_to_array(...) / regexp_substr(...) logic with DB.inFilterForCSV().

If I'm misunderstanding your suggestion, could you clarify how you would apply DB.inFilterForCSV() in a case where the CSV values are only available after evaluating a database expression?

hengsin: ok, my mistake.

hengsin (changes requested):
please address the coderabbitai review comments.

hengsin (approved):
looks good

PR #3277, opened 2026-06-01
IDEMPIERE-7022: AllocationReset : error message not explicit when com…

nmicoud:
…mit fails

https://idempiere.atlassian.net/browse/IDEMPIERE-7022

# Pull Request Checklist

### Tests
### Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of allocation reset by strengthening transaction commit/rollback flow to avoid partial commits.
  * Immediate abort and clearer error reporting when a delete operation fails, preventing inconsistent state.
  * Accurate reporting of successful deletions so administrators see correct counts after a reset.

### DIFF
--- a/org.adempiere.base.process/src/org/compiere/process/AllocationReset.java
+++ b/org.adempiere.base.process/src/org/compiere/process/AllocationReset.java
@@ -195,8 +195,15 @@ public class AllocationReset extends SvrProcess
 			if (log.isLoggable(Level.FINE)) log.fine(hdr.toString());
 			success = true;
 		}
-		if (success)
-			success = m_trx.commit();
+		if (success) {
+
+			try {
+				success = m_trx.commit(true);
+			}
+			catch (Exception e) {
+				throw new AdempiereException (e.getMessage());
+			}
+		}
 		else
 			m_trx.rollback();
 		return success;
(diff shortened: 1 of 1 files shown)


org.adempiere.base.process/src/org/compiere/process/AllocationReset.java
```diff
@@ -195,8 +195,18 @@ protected boolean delete(MAllocationHdr hdr)
...
+		if (success) {
+
+			try {
+				success = m_trx.commit(true);
+				if (!success)
+					m_trx.rollback();
+
+			} catch (Exception e) {
+				m_trx.rollback();
+				throw new AdempiereException(e.getMessage(), e);
```
hengsin: the improvement is fine but it seems inconsistent with the delete failure above. I think you either throw exception for both or log the error and return "delete error" and "commit error" respectively.

hengsin (changes requested):

org.adempiere.base.process/src/org/compiere/process/AllocationReset.java
nmicoud: Hi @hengsin 
I tried to handle the error in a better way while preserving the existing delete(hdr) method, with the addition of m_lastError and deleteAndGetError.

To be consistent, I think a new AD_Message is needed here : CommitError

wdyt?

hengsin (approved):
looks good

PR #3279, opened 2026-06-03
IDEMPIERE-7023 Add ISQLStatementRewriter extension point in Convert / DB providers

scolettads:
## Linked JIRA

[IDEMPIERE-7023](https://idempiere.atlassian.net/browse/IDEMPIERE-7023)

## Companion PR

IDEMPIERE-7024 (`IUIBehaviour` extension point) — independent OSGi
extension point sharing the same driver use case.

## Summary

Adds a new OSGi service interface `org.compiere.dbPort.ISQLStatementRewriter`
that lets a plugin pre-rewrite SQL statements before `Convert.convert()` and
the dialect `convertStatement()` execute them.

**Hook points:**
- `Convert.java`: static `rewriteStatements(String)` and `isConvertCacheable()`,
  with a `ThreadLocal` re-entrancy guard to prevent `StackOverflowError` when
  a rewriter implementation transitively executes SQL during its own
  evaluation
- `DB_Oracle.convertStatement()`: calls `Convert.rewriteStatements()` first
- `DB_PostgreSQL.convertStatement()`: same, plus skips the in-memory
  `convertCache` when at least one registered rewriter is not cacheable
  (i.e. depends on a dynamic context such as ThreadLocal)

## Use cases unlocked

- Transparent time-travel queries (e.g. read historicized records as-of a
  given date)
- Multi-tenant row-level SQL injection
- Audit logging at the SQL layer
- PII masking at the SQL layer
- Custom row-level security on top of OOTB ACL

## Backward compatibility

Without any provider registered (vanilla install),
`Service.locator().list(ISQLStatementRewriter.class).getServices()` returns
an empty list and `rewriteStatements()` returns the input unchanged.
Bit-for-bit identical to vanilla behavior.

Same OSGi pattern already used in iDempiere for `IModelValidator`,
`IColumnCallout`, `IFactRegister`, `ITabActivate`, etc.

## Tested on

iDempiere "Orion" with PostgreSQL 17. The hook is exercised by an
out-of-tree plugin (`it.finmatica.history-records`, proposed for community
publication under GPL v2 in a separate repository) which provides a
concrete `ISQLStatementRewriter` implementation for transparent time-travel
queries.

## Diff

~50 LOC across 4 files:
- `org.adempiere.base/src/org/compiere/dbPort/ISQLStatementRewriter.java` (new)
- `org.adempiere.base/src/org/compiere/dbPort/Convert.java`
- `org.compiere.db.oracle.provider/src/org/compiere/db/DB_Oracle.java`
- `org.compiere.db.postgresql.provider/src/org/compiere/db/DB_PostgreSQL.java`

[IDEMPIERE-7023]: https://idempiere.atlassian.net/browse/IDEMPIERE-7023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added an extensible SQL statement rewriting framework through a new extension point for pre-conversion SQL transformations.
  * Oracle and PostgreSQL conversion now runs registered rewrite steps sequentially before proceeding.

* **Bug Fixes**
  * Improved conversion result caching by considering whether each registered rewriter marks its output as safe to cache; caching is now enabled only when appropriate.

### DIFF
--- a/org.adempiere.base/src/org/compiere/dbPort/Convert.java
+++ b/org.adempiere.base/src/org/compiere/dbPort/Convert.java
@@ -243,9 +243,6 @@ public abstract class Convert
 			log.info(m_conversionError);
 			return null;
 		}
-		// IDEMPIERE-7023 hook: apply ISQLStatementRewriter providers before the
-		// Oracle->PostgreSQL conversion takes place.
-		sqlStatements = rewriteStatements(sqlStatements);
 		return convertIt (sqlStatements);
 	}   //  convert
 
--- a/org.adempiere.base/src/org/compiere/dbPort/Convert.java
+++ b/org.adempiere.base/src/org/compiere/dbPort/Convert.java
@@ -243,9 +243,74 @@ public abstract class Convert
 			log.info(m_conversionError);
 			return null;
 		}
-		//
+		// IDEMPIERE-7023 hook: apply ISQLStatementRewriter providers before the
+		// Oracle->PostgreSQL conversion takes place.
+		sqlStatements = rewriteStatements(sqlStatements);
 		return convertIt (sqlStatements);
 	}   //  convert
+
+	/**
+	 * IDEMPIERE-7023
+	 * ThreadLocal re-entrancy guard: when a rewriter (or its evaluation logic)
+	 * transitively executes a SQL statement that flows through convertStatement(),
+	 * the rewrite / cacheable check must NOT recurse into itself (StackOverflowError).
+	 * While the guard is ON, rewriteStatements() returns the input unchanged and
+	 * isConvertCacheable() returns true (vanilla behavior).
+	 */
+	private static final ThreadLocal<Boolean> rewriteGuard = ThreadLocal.withInitial(() -> Boolean.FALSE);
+
+	/**
+	 * IDEMPIERE-7023
+	 * Iterates all OSGi-registered ISQLStatementRewriter providers and chains
+	 * their rewrites. Returns the input unchanged when no provider is registered
+	 * or when invoked re-entrantly.
+	 * @param sqlStatements input SQL
+	 * @return possibly rewritten SQL
+	 */
+	public static String rewriteStatements(String sqlStatements)
+	{
+		if (sqlStatements == null) return null;
+		if (Boolean.TRUE.equals(rewriteGuard.get())) return sqlStatements;  // re-entrancy guard
+		java.util.List<org.compiere.dbPort.ISQLStatementRewriter> rewriters =
+				org.adempiere.base.Service.locator().list(org.compiere.dbPort.ISQLStatementRewriter.class).getServices();
+		if (rewriters == null || rewriters.isEmpty()) return sqlStatements;
+		rewriteGuard.set(Boolean.TRUE);
+		try {
+			String result = sqlStatements;
+			for (org.compiere.dbPort.ISQLStatementRewriter r : rewriters) {
+				result = r.rewriteStatements(result);
+			}
+			return result;
+		} finally {
+			rewriteGuard.set(Boolean.FALSE);
+		}
+	}
+
+	/**
+	 * IDEMPIERE-7023
+	 * @return true if EVERY registered rewriter declares its output cacheable.
+	 *         Used by the Oracle->PostgreSQL conversion cache in DB_PostgreSQL
+	 *         to avoid stale entries when a rewriter is dynamic (e.g. depends on
+	 *         a ThreadLocal context). Returns true under re-entrancy (a rewriter
+	 *         running SQL during its own evaluation) to break the loop and let
+	 *         the inner conversion proceed.
+	 */
+	public static boolean isConvertCacheable()
+	{
+		if (Boolean.TRUE.equals(rewriteGuard.get())) return true;  // re-entrancy guard
+		java.util.List<org.compiere.dbPort.ISQLStatementRewriter> rewriters =
+				org.adempiere.base.Service.locator().list(org.compiere.dbPort.ISQLStatementRewriter.class).getServices();
+		if (rewriters == null || rewriters.isEmpty()) return true;
+		rewriteGuard.set(Boolean.TRUE);
+		try {
+			for (org.compiere.dbPort.ISQLStatementRewriter r : rewriters) {
+				if (!r.rewriteIsCacheable()) return false;
+			}
+			return true;
+		} finally {
+			rewriteGuard.set(Boolean.FALSE);
+		}
+	}
 
 	/**
 	 *  Return last conversion error or null.
--- a/org.adempiere.base/src/org/compiere/dbPort/Convert.java
+++ b/org.adempiere.base/src/org/compiere/dbPort/Convert.java
@@ -247,28 +247,6 @@ public abstract class Convert
 	}   //  convert
 
 	/**
-	 * IDEMPIERE-7023
-	 * Iterates all OSGi-registered ISQLStatementRewriter providers and chains
-	 * their rewrites. Delegates to {@link SQLStatementRewriterProvider}.
-	 * @param sqlStatements input SQL
-	 * @return possibly rewritten SQL
-	 */
-	public static String rewriteStatements(String sqlStatements)
-	{
-		return SQLStatementRewriterProvider.rewriteStatements(sqlStatements);
-	}
-
-	/**
-	 * IDEMPIERE-7023
-	 * @return true if EVERY registered rewriter declares its output cacheable.
-	 *         Delegates to {@link SQLStatementRewriterProvider}.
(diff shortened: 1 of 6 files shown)


scolettads:
> Thanks for the catch. Confirmed: the `ThreadLocal rewriteGuard` only protects against re-entrant calls within the same stack frame (a rewriter that internally issues SQL going back through `convertStatement()`), not against the sequential dual call where `DB_PostgreSQL.convertStatement()` invokes `Convert.rewriteStatements()` on line 363 and then `m_convert.convert()` on line 384 would invoke it again.
>
> Fixed in 4b570123d2185f6a531a73eeaaa1c6c39f7a9b7d by removing the rewrite hook from `Convert.convert()`. The rewriter now runs exactly once per statement, in the DB-specific provider layer (`DB_Oracle.convertStatement` and `DB_PostgreSQL.convertStatement`). `Convert.convert(String)` has a single caller in the codebase (`DB_PostgreSQL.java:384`), so coverage is preserved.

scolettads:
@coderabbitai review

scolettads:
@coderabbitai review

org.adempiere.base/src/org/compiere/dbPort/Convert.java
```diff
@@ -243,10 +243,72 @@ public String[] convert (String sqlStatements)
...
+	 * or when invoked re-entrantly.
+	 * @param sqlStatements input SQL
+	 * @return possibly rewritten SQL
+	 */
+	public static String rewriteStatements(String sqlStatements)
+	{
+		if (sqlStatements == null) return null;
+		if (Boolean.TRUE.equals(rewriteGuard.get())) return sqlStatements;  // re-entrancy guard
+		java.util.List<org.compiere.dbPort.ISQLStatementRewriter> rewriters =
+				org.adempiere.base.Service.locator().list(org.compiere.dbPort.ISQLStatementRewriter.class).getServices();
```
hengsin: Instead of using locator, we now prefer the use of Component and `@Reference` annotation instead.

org.adempiere.base/src/org/compiere/dbPort/Convert.java
```diff
@@ -243,10 +243,72 @@ public String[] convert (String sqlStatements)
...
 
+	/**
+	 * IDEMPIERE-7023
+	 * ThreadLocal re-entrancy guard: when a rewriter (or its evaluation logic)
+	 * transitively executes a SQL statement that flows through convertStatement(),
+	 * the rewrite / cacheable check must NOT recurse into itself (StackOverflowError).
+	 * While the guard is ON, rewriteStatements() returns the input unchanged and
+	 * isConvertCacheable() returns true (vanilla behavior).
+	 */
+	private static final ThreadLocal<Boolean> rewriteGuard = ThreadLocal.withInitial(() -> Boolean.FALSE);
```
hengsin: Please investigate a solution that doesn't use ThreadLocal. This add overhead to everyone that's not using the rewriter extension point (which as it is, is everyone else!).

hengsin (changes requested):
see comments inline

org.adempiere.base/src/org/compiere/dbPort/Convert.java
scolettads: Thank you for the feedback. The ThreadLocal re-entrancy guard has been removed.

Re-entrancy is now prevented by contract on the ISQLStatementRewriter interface rather than by a runtime guard: the Javadoc on rewriteStatements() now explicitly states that implementations must not execute SQL through the iDempiere DB layer inside that method. If an implementation needs DB-derived metadata (e.g. to decide whether a table has history enabled), that data must be loaded into a local cache before rewrite time and served from the cache at rewrite time.

The History Records plugin already honours this contract: ADHistoryReplacer.hasHistory() reads from an in-memory map populated at bundle activation, so no SQL is executed during rewrite.

With no ISQLStatementRewriter provider registered (the default for everyone not using this extension point), rewriteStatements() now returns immediately after the inst == null || inst.rewriters.isEmpty() check — zero ThreadLocal access.

org.adempiere.base/src/org/compiere/dbPort/Convert.java
scolettads: Done

org.adempiere.base/src/org/compiere/dbPort/SQLStatementRewriterProvider.java
```diff
@@ -0,0 +1,116 @@
...
+	public static boolean isConvertCacheable()
+	{
+		SQLStatementRewriter inst = instance;
+		if (inst == null || inst.rewriters.isEmpty()) return true;
+		for (ISQLStatementRewriter r : inst.rewriters)
+		{
+			if (!r.rewriteIsCacheable()) return false;
+		}
+		return true;
+	}
```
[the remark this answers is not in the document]
scolettads: This is already addressed. Convert.rewriteStatements() and Convert.isConvertCacheable() are now thin delegators with no logic of their own:

public static String rewriteStatements(String sqlStatements) {
    return SQLStatementRewriter.rewriteStatements(sqlStatements);
}
public static boolean isConvertCacheable() {
    return SQLStatementRewriter.isConvertCacheable();
}
All orchestration logic lives exclusively in SQLStatementRewriter. The duplication referenced in the review was present in an earlier version of the PR (before the @Component/@Reference refactoring) and has since been removed.

org.adempiere.base/src/org/compiere/dbPort/Convert.java
hengsin: what do you means by "has been removed" - the thread local variable is still sitting here.

scolettads: Sorry, I missed yesterday's commit. I'm committing it now

org.adempiere.base/src/org/compiere/dbPort/Convert.java
scolettads: I've renamed provider as the same convention already done in other PR 3280

org.adempiere.base/src/org/compiere/dbPort/Convert.java
```diff
@@ -243,10 +243,31 @@ public String[] convert (String sqlStatements)
...
+	/**
+	 * IDEMPIERE-7023
+	 * @return true if EVERY registered rewriter declares its output cacheable.
+	 *         Delegates to {@link SQLStatementRewriterProvider}.
+	 */
+	public static boolean isConvertCacheable()
+	{
+		return SQLStatementRewriterProvider.isConvertCacheable();
+	}
+
```
hengsin: This two additions seems unnecessary. I don't see any gains here over just making direct calls to SQLStatementRewriterProvider.

hengsin (changes requested):

org.adempiere.base/src/org/compiere/dbPort/Convert.java
scolettads: You're right. In the new commit DB_Oracle/DB_Postgress calls directily SqlStatementRewriterProvider.

org.adempiere.base/src/org/compiere/dbPort/Convert.java
```diff
@@ -243,7 +243,6 @@ public String[] convert (String sqlStatements)
 			log.info(m_conversionError);
 			return null;
 		}
-		//
```
hengsin, line 246: please remove the unnecessary changes above.

hengsin (changes requested):

hengsin:
hi @CarlosRuiz-globalqss , this looks fine and non-disruptive to me. Ok to proceed ?

org.adempiere.base/src/org/compiere/dbPort/Convert.java
scolettads: I don't understand why Convert class at line 246 doens't have this  empty comment //

If you check https://github.com/scolettads/idempiere/blame/IDEMPIERE-7023-sqlrewriter/org.adempiere.base/src/org/compiere/dbPort/Convert.java#L278 the actual version in remote IDEMPIERE-7023-sqlrewriter branch has only the comment at } end method bracket, that's usefull to evidence which method is ending. 
What Am I wrong ?

d-ruiz:
@CarlosRuiz-globalqss this pull request is waiting for your confirmation

CarlosRuiz-globalqss:
> hi @CarlosRuiz-globalqss , this looks fine and non-disruptive to me. Ok to proceed ?

yes - go ahead

hengsin (approved):
looks good.

PR #3280, opened 2026-06-03
IDEMPIERE-7024 Add IUIBehaviour extension point for lookup cache and field editability

scolettads:
[IDEMPIERE-7024](https://idempiere.atlassian.net/browse/IDEMPIERE-7024)

## Companion PR

#3279 — [IDEMPIERE-7023](https://idempiere.atlassian.net/browse/IDEMPIERE-7023)
(`ISQLStatementRewriter`) — independent OSGi extension point sharing the
same driver use case.

## Summary

Adds a new OSGi service interface `org.adempiere.base.IUIBehaviour` that
lets a plugin participate in three UI decisions taken inside `MLookup` and
`GridField`:

1. `isLookupCacheable(Lookup, MLookupInfo)` — vetoes `MLookup` cache
   population when the plugin's effective context is dynamic (e.g. a
   ThreadLocal that changes per request)
2. `isEditable(Properties, GridTab)` — additional check on tab editability
3. `isEditable(Properties, GridField, boolean, boolean)` — additional
   check on field editability

A static helper `org.adempiere.base.UIBehaviour` aggregates all registered
providers with AND logic (one veto = denied).

**Hook points:**
- `MLookup.java`: the `m_lookup.put(...)` cache fill is gated by
  `UIBehaviour.isLookupCacheable(this, m_info)`
- `GridField.java`: `isEditable(Properties, boolean, boolean)` consults
  `UIBehaviour.isEditable(...)` and forces read-only if any provider denies

## Driver use case — historicized record navigation

The same out-of-tree History Records plugin that drives PR #3279 also
relies on these UI hooks.

**Scenario A — disable lookup cache during time-machine navigation.**
When the user activates a "Time Machine" date in the session context
(ThreadLocal), every `MLookup` dropdown must read fresh values: the SQL
rewriter (PR #3279) needs to see the lookup query at every access, so it
can rewrite it against the historicized `_HST` table for the correct
date. Without the hook, `MLookup` keeps serving cached values populated
when the context was "present time" — silent data inconsistency that
application code cannot fix because the cache is private to `MLookup`.

With the hook, the plugin registers an `IUIBehaviour` that returns
`false` from `isLookupCacheable` when the Time Machine context is
active, and the cache is transparently bypassed for that request.

**Scenario B — read-only on sealed past snapshots.**
When the user browses a row in the History sub-tab with
`HSTActualRecord = 'N'` (a sealed past snapshot), every field must
become read-only — otherwise the historical audit trail can be tampered
with on save. Today the only ways are: patching `GridField` (fork the
core) or throwing from a `ModelValidator` on save (bad UX: the user
sees the error only when clicking Save). With `IUIBehaviour.isEditable`
returning `false`, the field shows greyed out as soon as the row is
selected.

## Backward compatibility

Without any provider registered (vanilla install), all helper methods
return `true` and behavior is bit-for-bit identical to current core.
Same OSGi pattern already used for `IModelValidator`, `IColumnCallout`,
`IFactRegister`, etc.

## Tested on

iDempiere "Orion" with PostgreSQL 17.

## diff

4 files:
- `org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java` (new)
- `org.adempiere.base/src/org/adempiere/base/UIBehaviour.java` (new helper)
- `org.adempiere.base/src/org/compiere/model/MLookup.java`
- `org.adempiere.base/src/org/compiere/model/GridField.java`

# Pull Request Checklist

### Tests
### Documentation
- - [x ] Breaking change (fix or feature that would cause existing functionality to not work as expected): I have documented the change in a clear way that everyone in the community can understand the impact of the change.
- - [x] Improvement (improves and existing functionality): This documentation is needed if the improvement changes the way the user interacts with the system or the outcome of a process/task changes. If it is just, for instance, a performance improvement, documentation might not be needed. 

> Note on documentation checkboxes: I will create the wiki page once the
> two PRs (IDEMPIERE-7023 + IDEMPIERE-7024) are approved and the driver
> plugin repository is publicly released. Will reference both JIRA tickets
> and PRs from the wiki page.

> Note on unit tests: I did not add unit tests for the two hooks because
> they are pure pass-through plumbing (delegating to OSGi service
> registry) and there is no existing unit-test suite for `MLookup` /
> `GridField` to extend. Happy to add tests if the reviewers prefer.

[IDEMPIERE-7024]: https://idempiere.atlassian.net/browse/IDEMPIERE-7024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IDEMPIERE-7023]: https://idempiere.atlassian.net/browse/IDEMPIERE-7023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a new UI behavior extensibility point that can control (a) whether lookup results may be stored in shared caches and (b) whether tabs and fields are editable.
  * Extensions can return “not relevant” to keep defaults, or explicitly allow/deny; decisions are aggregated across all registered extensions using veto/AND behavior.

* **Behavior Changes**
  * Lookup direct retrieval now bypasses shared-cache read/write when extensions disallow caching, while still returning resolved results for the current request.


org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
```diff
@@ -0,0 +1,99 @@
...
+	{
+		List<IUIBehaviour> services = Service.locator().list(IUIBehaviour.class).getServices();
+		if (services == null || services.isEmpty())
+			return true;
+		for (IUIBehaviour svc : services) {
+			Boolean res = svc.isLookupCacheable(lookup, lookupInfo);
+			if (Boolean.FALSE.equals(res))
+				return false;
+		}
+		return true;
```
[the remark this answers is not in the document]
scolettads: > _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> 
> **Isolate provider failures inside the aggregator.**
> 
> A single `RuntimeException` from any `IUIBehaviour` implementation currently bubbles out of core UI code and can abort lookup resolution or field editability checks for the whole window. Since this is the extension boundary, catch/log provider failures here and apply an explicit fallback instead of letting one bad plugin take down the caller.
> 
> Also applies to: 75-79, 92-96
> 
> 🤖 Prompt for AI Agents

Fixed in b175398a75. Wrapped every svc.xxx(...) call in UIBehaviour with try { ... } catch (Throwable t) { log.log(WARNING, ..., t); }, treating a failing provider as neutral (res = null) so it does not alter the AND-aggregated decision. This is the same pattern already used by AbstractEventHandler in this codebase for service-handler boundaries. The choice of Throwable (rather than Exception) is intentional: in OSGi a misconfigured bundle can raise LinkageError / NoClassDefFoundError, which are Error subclasses; Exception alone would leave that surface uncovered.

scolettads:
> [Review Change Stack](https://app.coderabbit.ai/change-stack/idempiere/idempiere/pull/3280?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)
> 
> No actionable comments were generated in the recent review. 🎉
> 
> ℹ️ Recent review info
> ⚙️ Run configuration
> **Configuration used**: Path: .coderabbit.yaml
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> **Run ID**: `c4573281-6bc7-4b0d-9851-98caba8dca69`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [aa193b4](https://github.com/idempiere/idempiere/commit/aa193b4989430ce661402557c0eae7386d546ce3) and [bc513f3](https://github.com/idempiere/idempiere/commit/bc513f32fad7416d40e7562c1468a4db3dbd3d53).
> 
> 📒 Files selected for processing (1)
> * `org.adempiere.base/src/org/compiere/model/MLookup.java`
> 
> ## Walkthrough
> This PR adds an OSGi extension point interface `IUIBehaviour`, a `UIBehaviour` aggregation utility that AND-combines provider answers, and wires those aggregated decisions into lookup cache writes and GridField editability checks.
> 
> ## Changes
> **UI Behaviour Extension and Framework Integration**
> 
> Layer / File(s)	Summary
> **Extension Point Contract**
> `org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java`	Three nullable-`Boolean` methods define the extension contract: `isLookupCacheable`, `isEditable(Properties, GridTab)`, and `isEditable(Properties, GridField, boolean, boolean)` with `null` meaning "not relevant" and `true/false` as explicit decisions.
> **Decision Aggregation Service**
> `org.adempiere.base/src/org/adempiere/base/UIBehaviour.java`	Static utility class queries all registered `IUIBehaviour` services via `Service.locator().list(IUIBehaviour.class).getServices()`, applies AND logic (any `false` blocks, `null` non-blocking), logs provider exceptions as neutral, and returns `true` when no providers exist.
> **Lookup Caching Integration**
> `org.adempiere.base/src/org/compiere/model/MLookup.java`	`getDirect()` calls `UIBehaviour.isLookupCacheable()` and gates the cache read/write paths (KeyNamePair/ValueNamePair and local direct cache) with the aggregated result (`saveInCache && uiCacheable`); resolved values are still returned immediately even if caching is blocked. Also tightens MLoader cacheKey parameter appending.
> **Field Editability Integration**
> `org.adempiere.base/src/org/compiere/model/GridField.java`	`isEditable()` calls `UIBehaviour.isEditable()` on entry; a `false` aggregated result logs at `FINEST` and immediately returns `false`, short-circuiting subsequent read-only and tab-level checks.
> 🎯 3 (Moderate) | ⏱️ ~20 minutes
> 
> 🚥 Pre-merge checks | ✅ 5
> ✅ Passed checks (5 passed)
> Check name	Status	Explanation
> Title check	✅ Passed	The title accurately describes the main change—adding a new IUIBehaviour OSGi extension point for controlling lookup cache and field editability decisions.
> Description check	✅ Passed	The PR description is comprehensive, including JIRA link, summary, implementation details, use cases, backward compatibility notes, and a completed checklist; one minor test note is explained.
> Docstring Coverage	✅ Passed	Docstring coverage is 90.91% which is sufficient. The required threshold is 80.00%.
> Linked Issues check	✅ Passed	Check skipped because no linked issues were found for this pull request.
> Out of Scope Changes check	✅ Passed	Check skipped because no linked issues were found for this pull request.
> ✏️ Tip: You can configure your own custom pre-merge checks in the settings.
> 
> ✨ Finishing Touches
> 🧪 Generate unit tests (beta)
> * [ ]     Create PR with unit tests
> 
> Thanks for using [CodeRabbit](https://coderabbit.ai?utm_source=oss&utm_medium=github&utm_campaign=idempiere/idempiere&utm_content=3280)! It's free for OSS, and your support helps us grow. If you like it, consider giving us a shout-out.
> 
> ❤️ Share
> * [X](https://twitter.com/intent/tweet?text=I%20just%20used%20%40coderabbitai%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20the%20proprietary%20code.%20Check%20it%20out%3A&url=https%3A//coderabbit.ai)
> * [Mastodon](https://mastodon.social/share?text=I%20just%20used%20%40coderabbitai%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20the%20proprietary%20code.%20Check%20it%20out%3A%20https%3A%2F%2Fcoderabbit.ai)
> * [Reddit](https://www.reddit.com/submit?title=Great%20tool%20for%20code%20review%20-%20CodeRabbit&text=I%20just%20used%20CodeRabbit%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20proprietary%20code.%20Check%20it%20out%3A%20https%3A//coderabbit.ai)
> * [LinkedIn](https://www.linkedin.com/sharing/share-offsite/?url=https%3A%2F%2Fcoderabbit.ai&mini=true&title=Great%20tool%20for%20code%20review%20-%20CodeRabbit&summary=I%20just%20used%20CodeRabbit%20for%20my%20code%20review%2C%20and%20it%27s%20fantastic%21%20It%27s%20free%20for%20OSS%20and%20offers%20a%20free%20trial%20for%20proprietary%20code)
> 
> Comment `@coderabbitai help` to get the list of available commands and usage tips.

MLookup.java:537-545 + batched path — Apply cache veto to the batched getDirect(Object[]) too

scolettads:
@coderabbitai review

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
```diff
@@ -0,0 +1,81 @@
...
+	 * Additional check on field editability. The field is editable only if
+	 * EVERY registered provider allows it.
+	 *
+	 * @param ctx          current context
+	 * @param field        the field being evaluated
+	 * @param checkContext pass-through flag from GridField.isEditable
+	 * @param isGrid       pass-through flag from GridField.isEditable
+	 * @return null when not relevant, true to allow edit, false to deny
+	 */
+	public Boolean isEditable(Properties ctx, GridField field,
```
hengsin: why not just a simple boolean ?

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
```diff
@@ -0,0 +1,81 @@
...
+{
+	/**
+	 * Allow caching of the lookup. The lookup cache is enabled only if EVERY
+	 * registered provider allows it.
+	 *
+	 * @param lookup     the lookup being evaluated
+	 * @param lookupInfo the lookup info (may be null)
+	 * @return null when not relevant, true to allow cache, false to disable
+	 */
+	public Boolean isLookupCacheable(Lookup lookup, MLookupInfo lookupInfo);
```
hengsin: turning off cache has pretty significant impact on UI performance, how is that being addressed in your implementation ?

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
```diff
@@ -0,0 +1,133 @@
...
+		// utility class - no instances
+	}
+
+	/**
+	 * @return true if EVERY registered IUIBehaviour service allows caching
+	 *         (or if no service is registered).
+	 */
+	public static boolean isLookupCacheable(Lookup lookup, MLookupInfo lookupInfo)
+	{
+		List<IUIBehaviour> services = Service.locator().list(IUIBehaviour.class).getServices();
```
hengsin: Instead of using locator, we now prefer the Component and `@Reference` annotation approach.

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
```diff
@@ -0,0 +1,81 @@
...
+	 *
+	 * @param ctx          current context
+	 * @param field        the field being evaluated
+	 * @param checkContext pass-through flag from GridField.isEditable
+	 * @param isGrid       pass-through flag from GridField.isEditable
+	 * @return null when not relevant, true to allow edit, false to deny
+	 */
+	public Boolean isEditable(Properties ctx, GridField field,
+	                          boolean checkContext, boolean isGrid);
+}
```
hengsin, line 95: Cleaner to use `isFieldEditable` and `isTabEditable` instead. Also, the isEditable for GridTab is not call by GridTab.

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
```diff
@@ -0,0 +1,133 @@
...
+			return true;
+		for (IUIBehaviour svc : services) {
+			Boolean res = null;
+			try {
+				res = svc.isLookupCacheable(lookup, lookupInfo);
+			} catch (Throwable t) {
+				// IDEMPIERE-7024: isolate provider failures - log and treat as neutral
+				log.log(Level.WARNING,
+					"IUIBehaviour provider " + svc.getClass().getName()
+					+ " threw in isLookupCacheable; treating as neutral", t);
```
hengsin: Dangerous to catch and logged warning for Throwable since Throwable can include JVM issues like OOM.

org.adempiere.base/src/org/compiere/model/GridField.java
```diff
@@ -446,6 +446,16 @@ public boolean isEditable (Properties ctx, boolean checkContext,boolean isGrid)
...
+		// IDEMPIERE-7024 hook: ask registered IUIBehaviour providers whether
+		// this field is editable in the current context; any provider returning
+		// false forces read-only. Placed before the Posted /
+		// Record_ID escape hatch so providers can veto even those normally-
+		// always-enabled fields (e.g. time-travel read-only mode).
+		if (!org.adempiere.base.UIBehaviour.isEditable(ctx, this, checkContext, isGrid)) {
+			if (log.isLoggable(Level.FINEST)) log.finest(m_vo.ColumnName + " NO - UIBehaviour denied");
+			return false;
+		}
+
```
hengsin: Perhaps better to move this after the always updateable check.

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
scolettads: In iDempiere core, the lookup cache continues to work as usual — the IUIBehaviour interface is neutral by design: when no provider is registered, or when a provider returns null, caching behaviour is identical to vanilla iDempiere.

When the History Records plug-in is installed, lookups pointing to a table that is explicitly marked as historicised will always read from the history (_HST) table, bypassing the cache. This is an intentional trade-off limited to that specific edge case: a table must be actively flagged for historicisation, and only lookups targeting that table are affected. All other lookups preserve their cache as normal.

This is a known limitation of the first community release. Future versions of the History Records plug-in may introduce optimisations — for example, bypassing the cache only when the Time Machine is actually active on the current thread.

hengsin: I know it doesn't effect current core. My question is referring to the performance impact when using the time machine.

scolettads: Good point on the performance impact. We actually see an opportunity to leverage the existing cache infrastructure rather than bypassing it entirely.

The idea is to add a getLookupCacheKeySuffix() method to IUIBehaviour. The existing cache key in MLookup.getDirect() is a plain string:

String cacheKey = tableName + "|" + keyColumn + "|" + refId + "|" + language;
A provider could append a suffix to it — for example the History Records plug-in would append the active history date:

"HST@" + historyDate.getDate()   // e.g. "C_BPartner|C_BPartner_ID|...|HST@2026/22/06"
This creates a separate cache bucket per history date using the existing CCache machinery — no new infrastructure needed. In normal mode (no time-travel active) the suffix is null and the cache key is identical to vanilla, so there is zero impact on standard usage.

isLookupCacheable would no longer need to return false in most cases — caching stays on, just partitioned by date.

We plan to evolve IUIBehaviour in this direction. Would you be open to including getLookupCacheKeySuffix() as a default method in this PR, or would you prefer to keep this interface minimal for now and address it in a follow-up?

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
scolettads: Done. UIBehaviour is now a @Component(immediate = true, service = {}) with a @Reference(cardinality = MULTIPLE, policy = DYNAMIC) field backed by a CopyOnWriteArrayList, and a volatile static instance for the existing static method API.

See the last commit

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
scolettads: Done. Both methods renamed in IUIBehaviour and UIBehaviour. isTabEditable is now called from GridTab.isReadOnly()

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
scolettads: Done. JVM-level errors (OutOfMemoryError, StackOverflowError, etc.) now propagate normally. Only Exception is caught, logged at WARNING and treated as neutral.

org.adempiere.base/src/org/compiere/model/GridField.java
scolettads: Done. The hook now sits after both IsAlwaysUpdateable and AlwaysUpdatableLogic, so providers cannot veto fields that core logic forces editable.

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
scolettads: Done. I change IUIBehaviour methods to return boolean (veto semantics)

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * For each decision the aggregator applies AND/veto logic:
+ *   - all providers return true (or null = neutral) -> true
+ *   - at least one provider returns false           -> false
+ *
+ * With no registered provider returns true (vanilla behaviour).
+ *
+ * Provider failures are isolated: any Throwable raised by a provider is
+ * logged at WARNING and treated as neutral, so a single bad plug-in
+ * cannot abort lookup resolution or field-editability for the whole window.
```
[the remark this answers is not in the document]
scolettads: Done

org.adempiere.base/src/org/compiere/model/GridTab.java
```diff
@@ -3725,11 +3729,11 @@ public void reset() {
...
-			return dse.dataStatusEvent();
-		}
-		return null;
+	public DataStatusEvent getLastDataStatusEvent() {
+		DataStatusEventRecord dse = m_lastDataStatusEventReference.get();
+		if (dse != null) {
+			return dse.dataStatusEvent();
+		}
+		return null;
 	}
```
hengsin, line 3738: there are some white space only changes above, please remove it.

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * With no registered provider returns true (vanilla behaviour).
+ *
+ * Provider failures are isolated: any Exception raised by a provider is
+ * logged at WARNING and treated as neutral (allowing), so a single bad plug-in
+ * cannot abort lookup resolution or field-editability for the whole window.
+ * JVM errors (OutOfMemoryError, etc.) are intentionally not caught and propagate normally.
+ */
+@Component(name = "org.adempiere.base.UIBehaviour", immediate = true, service = {})
+public class UIBehaviour
```
hengsin: Better rename this as UIBehaviour provider to remove the confusion that this is an implementation of IUIBehaviour.

hengsin (changes requested):

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
hengsin: If possible, it is better to create an API that's more practical in a production environment.

org.adempiere.base/src/org/compiere/model/GridTab.java
scolettads: Fixed. The two remaining whitespace-only hunks (lineNoCurrentRow and getLastDataStatusEvent) are incidental trailing-whitespace differences introduced when Eclipse auto-saved the file during editing. They are confirmed whitespace-only via git diff -w. The net functional change to GridTab.java is only the four lines of the isTabEditable hook.

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
scolettads: Done

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
scolettads: Thank you for the direction.

We've added getLookupCacheKeySuffix(Lookup, MLookupInfo) as a method on IUIBehaviour (returns null by default) and the corresponding aggregation logic in UIBehaviourProvider.

The idea: instead of disabling the cache, a provider returns, such as history-records plugin, return a key cache based on suffix (e.g. "HST@2026-06-23"). MLookup would append it to the existing cache key, creating an isolated bucket per history date:

example:
C_BPartner|C_BPartner_ID|0|it_IT|HST@2026-06-23
In normal mode (no time-travel active) the suffix is null — zero overhead, identical to vanilla behaviour.

MLookup doens't not call getLookupCacheKeySuffix in this PR to limit the scope to the extension point API.

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
hengsin: hi, you forgot to remove this file.

org.adempiere.base/src/org/compiere/model/GridTab.java
hengsin: It will be good if you can remove the trailing-whitespace differences too. I guess you are using the AnyEdit plugin, if yes, the auto removal of trailing whitespaces can be turned off in preference.

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
```diff
@@ -0,0 +1,111 @@
...
+	 * overhead, identical to vanilla iDempiere.
+	 * <p>
+	 * When a non-null suffix is returned by any provider, {@link #isLookupCacheable}
+	 * is overridden to {@code true} for that lookup (partitioned cache is on).
+	 *
+	 * @param lookup     the lookup being evaluated
+	 * @param lookupInfo the lookup info (may be null)
+	 * @return null to use the standard cache key, or a non-empty string to partition
+	 */
+	default String getLookupCacheKeySuffix(Lookup lookup, MLookupInfo lookupInfo) {
```
hengsin, line 69: Perhaps you have not pushed some commits ? The isLookupCacheable method is still there and MLookup is still calling isLookupCacheable  instead of getLookupCacheKeySuffix

hengsin (changes requested):

org.adempiere.base/src/org/adempiere/base/IUIBehaviour.java
scolettads: Done. I was intending to add only the method for future purpose. BUt you are right. I pushed getLookupCacheKeySuffix, changed MLookup and removed isLookupCacheable

org.adempiere.base/src/org/adempiere/base/UIBehaviour.java
scolettads: Done

org.adempiere.base/src/org/compiere/model/GridTab.java
scolettads: I'm following your suggestion on using AnyEdit plugin to auto removal trailing whitespaces. IN order to clean already committed white spaces, I've restore GridTab from master version and then add only the hook portion of code. In this way, after merging the pull request, doesn't should appear trailing white space commit with the risk to have useless conflict

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -1173,9 +1185,9 @@ protected void doRun()
 			boolean isNumber = m_info.KeyColumn.endsWith("_ID");
 			
 			String cacheKey = sql.toString();
-			if (m_info.Parameters != null && !m_info.Parameters.isEmpty())
-			{
-				cacheKey += "|" + m_info.Parameters.toString();
+			if (m_info.Parameters != null && !m_info.Parameters.isEmpty())
+			{
+				cacheKey += "|" + m_info.Parameters.toString();
```
hengsin: this looks like another block of white space only changes.

org.adempiere.base/src/org/compiere/model/MLookup.java
```diff
@@ -682,8 +692,8 @@ public NamePair[] getDirect(Object[] keys)
 					vnp = new ValueNamePair(key.toString(), null);
 					notInCaches.put(key.toString(), i);
 				}
-				list.add(vnp);				
-			}			
+				list.add(vnp);
```
hengsin: 668 - 686 : white space only changes

org.adempiere.base/src/org/adempiere/base/UIBehaviourProvider.java
```diff
@@ -0,0 +1,171 @@
...
+	public static String getLookupCacheKeySuffix(Lookup lookup, MLookupInfo lookupInfo)
+	{
+		UIBehaviourProvider inst = instance;
+		if (inst == null || inst.behaviours.isEmpty())
+			return null;
+		for (IUIBehaviour svc : inst.behaviours) {
+			try {
+				String suffix = svc.getLookupCacheKeySuffix(lookup, lookupInfo);
+				if (suffix != null)
+					return suffix;
```
hengsin: this multiple provider and return first non-null suffix flow looks random and confusing (think if you have > 1 provider return non-null suffix, there's no well define rules here).

I guess your use case probably has just 1 provider ...

scolettads: Thank you for the observation. You are right that "first non-null wins" is undefined when multiple providers are registered.

I changed the aggregation to concatenate all non-null suffixes with | as separator. The semantics are well-defined: each provider contributes an independent dimension to the composite cache key.

Example with two providers active simultaneously:

History Records provider → "HST@2026-06-23" (partitions by history date)
A hypothetical role-based provider → "ROLE@admin" (partitions by role)
Resulting cache key:

C_BPartner|C_BPartner_ID|0|it_IT|HST@2026-06-23|ROLE@admin
If the role changes to user, the key becomes ...|HST@2026-06-23|ROLE@user — a separate bucket, which is correct because the value for that lookup may differ per role. Each combination of active dimensions gets its own isolated cache entry.

With a single provider (our use case) the result is identical to returning the suffix directly. With zero providers the method returns null and MLookup behaves exactly as vanilla iDempiere.

org.adempiere.base/src/org/compiere/model/MLookup.java
scolettads: Done. 
It was and old bommit did before AnyEdit installation

org.adempiere.base/src/org/compiere/model/MLookup.java
scolettads: Fix whitespace-only diffs in MLookup: reset to master and re-apply hook lines only, to preserve original trailing whitespace on untouched lines

scolettads:
I believe I've addressed all the review comments from @hengsin. @hengsin, could you please take another look and let me know if everything is okay?

org.adempiere.base/src/org/adempiere/base/UIBehaviourProvider.java
```diff
@@ -0,0 +1,174 @@
...
+		behaviours.remove(behaviour);
+	}
+
+	/**
+	 * Return the first non-null suffix from any registered provider, or null.
+	 * A non-null suffix appended to the cache key partitions the cache (e.g. by
+	 * history date) without disabling it.
+	 *
+	 * @return first non-null suffix, or null if no provider supplies one
+	 */
```
hengsin, line 109: The javadoc is not in sync with the latest change

hengsin (changes requested):

org.adempiere.base/src/org/adempiere/base/UIBehaviourProvider.java
scolettads: Done. I've written the Multi-provider concatenation javadoc mechanism

CarlosRuiz-globalqss:
@scolettads - please check @hengsin 's comment at https://idempiere.atlassian.net/browse/IDEMPIERE-7023?focusedCommentId=73345

scolettads:
> @scolettads - please check @hengsin 's comment at https://idempiere.atlassian.net/browse/IDEMPIERE-7023?focusedCommentId=73345

Done. Added OSGI-INF/org.adempiere.base.UIBehaviourProvider.xml

hengsin (approved):
looks good

PR #3311, opened 2026-07-18
IDEMPIERE-6552 SSO with OIDC enforces a login form

mbozem:
## Description

OIDC authentication currently adds `prompt=login` to every authorization request. This forces the identity provider to display a login form even when an existing session or remember-me cookie is available.

This change makes that behavior configurable per SSO principal configuration through the new **Force Login** flag. Existing configurations default to enabled, preserving the current logout behavior. Administrators can disable it for providers where an existing authenticated session or remember-me cookie should be reused.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-6552

## Root cause

`OIDCPrincipalService` unconditionally set `Prompt.Type.LOGIN` when building the authorization request. Removing it globally would fix remember-me behavior, but can cause problems in the logout flow for providers that require forced reauthentication.

## Changes

- Add the `IsForceLogin` setting to `SSO_PrincipalConfig`.
- Default the setting to enabled for backward compatibility.
- Show **Force Login** only for OIDC provider configurations.
- Add `prompt=login` only when **Force Login** is enabled.

## Impact

Each OIDC provider configuration can now choose the appropriate behavior:

- Enabled: preserve the existing forced-login behavior.
- Disabled: allow the provider to reuse an existing authenticated session or remember-me cookie.

## Validation

- Verified the generated model and interface changes.
- Verified Oracle and PostgreSQL migration scripts.
- `git diff --check` passes.

## How to test

1. Apply the migration and open an OIDC SSO principal configuration.
2. Verify that **Force Login** is visible for OIDC and enabled by default.
3. With **Force Login** enabled, start OIDC login and verify that the provider requires authentication.
4. Disable **Force Login**, establish a persistent provider session, and start OIDC login again.
5. Verify that the provider can reuse the existing session instead of forcing the login form.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable **Force Login** setting for OIDC single sign-on providers.
  * The setting is available in the provider configuration and defaults to enabled.
  * The option is displayed only for OIDC providers.

* **Bug Fixes**
  * OIDC authentication no longer always forces the login prompt, allowing the identity provider’s existing session to be used.

### DIFF
--- a/org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
+++ b/org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
@@ -288,7 +288,6 @@ public class OIDCPrincipalService implements ISSOPrincipalService {
 					new URI(getRedirectURL(principalConfig, redirectMode))) 
 			    .state(new State())
 			    .nonce(new Nonce())
-			    .prompt(new Prompt(Prompt.Type.LOGIN)) 
 			    .endpointURI(getMetaData().getAuthorizationEndpointURI())
 			    .build();
 		} catch (URISyntaxException e) {
--- a/org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
+++ b/org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
@@ -281,15 +281,17 @@ public class OIDCPrincipalService implements ISSOPrincipalService {
 			throws IOException {		
 		AuthenticationRequest authRequest = null;
 		try {
-			authRequest = new AuthenticationRequest.Builder(
+			AuthenticationRequest.Builder builder = new AuthenticationRequest.Builder(
 					new ResponseType(AUTHENTICATION_CODE_PARAMETER),		      
 					new Scope("openid", "profile", "email"),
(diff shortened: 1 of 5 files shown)


org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
```diff
@@ -288,7 +288,6 @@ public void redirectForAuthentication(HttpServletRequest request, HttpServletRes
 					new URI(getRedirectURL(principalConfig, redirectMode))) 
 			    .state(new State())
 			    .nonce(new Nonce())
-			    .prompt(new Prompt(Prompt.Type.LOGIN))
```
hengsin, line 291: Please make this configurable by provider. This is problematic for the logout flow for some provider.

hengsin (changes requested):

org.idempiere.ui.sso.oidc/src/org/idempiere/ui/sso/oidc/service/OIDCPrincipalService.java
mbozem: Thanks, Heng Sin. I made `prompt=login` configurable per SSO principal configuration through a new **Force Login** flag. It defaults to enabled to preserve the existing behavior and avoid regressions in provider logout flows. For providers where an existing session or remember-me cookie should be reused, the flag can be disabled.

hengsin (approved):
looks good

PR #3318, opened 2026-07-29
IDEMPIERE-7070 Make report content generation extensible through OSGi

mbozem:
## Status

The PR ist now ready for review.

## Description

Introduce UI-independent extension points for report content generation and post-processing. Plug-ins can select a report renderer through a ranked OSGi service and can then augment the generated content through an ordered processor chain, for example with PDF attachments, PDF/A, signatures, watermarks, or ZUGFeRD/Factur-X content.

The extension API is not limited to Jasper Reports. The default renderer is provided by the Jasper bundle for Jasper-based reports; other report types continue to use the standard `ReportEngine` fallback. Post-processors operate on either result through the same API.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7070

## Problem

The previous extension point was part of the ZK report viewer and exposed ZK- and Jasper-specific types. It was therefore unavailable to headless and server-side consumers. Several report-generation paths called `ReportEngine` directly and bypassed enhanced output supplied by plug-ins.

A single winning renderer also cannot cleanly combine independent enhancements. For example, ZUGFeRD generation, PDF/A conversion, and signing may all need to operate on the same report in a defined order.

Jasper export files are generated as temporary files. The viewer previously exposed the physical temporary filename, including the random suffix generated by `File.createTempFile`, as the visible preview and download filename.

## Changes

- Add UI- and report-engine-independent report content contracts to the Base bundle.
- Add ranked OSGi renderer lookup and a central report-content API to `Core`.
- Add `IReportContentProcessor` and apply all matching processors in descending OSGi service-ranking order after rendering or fallback generation.
- Cache processed viewer content per MIME type and file extension so preview and export reuse the same final result.
- Preserve the original `ProcessInfo`, parameters, record selections, language, user/client, and transaction context when creating report content.
- Support optional target files and copy only the final processed content to the target.
- Add the default Jasper renderer to `org.adempiere.report.jasper` with service ranking `0`.
- Route the ZK report viewer, workflow reports, server-side exports, document PDF generation, and batch-print processes through the common API.
- Keep the existing direct-printer path unchanged.
- Retain the legacy `ZkJRViewer` adapters and the report viewer's existing export formats.
- Use the logical report name for Jasper preview and download media while retaining unique temporary filenames internally.
- Add integration coverage for renderer ranking, unsupported-request fallback, default Jasper registration, processor ordering/applicability, invalid processor output, and processed-content caching.

## Compatibility

- Existing Jasper reports are handled by the default renderer from the Jasper bundle.
- Non-Jasper reports and installations without a matching renderer fall back to `ReportEngine`.
- Installations without an `IReportContentProcessor` retain the generated content unchanged.
- Headless installations can use the Jasper renderer and processor chain without the ZK bundle.
- Direct printer output retains its existing behavior.
- Viewer downloads no longer expose random temporary-file suffixes; generated content and internal collision avoidance are unchanged.
- No database migration is required.

## Testing still required

- Test preview and export through the ZK report viewer.
- Test workflow reports, server-side exports, document PDFs, and batch-print processes.
- Test the standard `ReportEngine` fallback without a matching renderer.
- Test installations with no content processors.
- Test multiple applicable processors and verify their configured service-ranking order.
- Test processors that return a new file as well as processors that modify the input file.
- Test representative combinations such as ZUGFeRD/Factur-X, PDF/A conversion, attachments, and signing.

## Validation

- `git diff --check` passes.
- `mvn -DskipTests package` passes for the complete 61-module reactor.
- The focused `ReportViewerContentRendererFactoryTest` build compiles and starts the OSGi test runtime; execution in the local checkout is blocked before the test methods by the missing `idempiereEnv.properties` runtime configuration.
- A subsequent focused `org.adempiere.ui.zk` build attempt is blocked during target-platform resolution because `org.idempiere.p2.targetplatform:14.0.0-SNAPSHOT` is unavailable from the configured repositories.
- Direct `ReportEngine` content-export calls in the affected production paths have been replaced by the common API.
- No references to the removed ZK-specific renderer contract or ZK Jasper factory remain.

## Additional fixes from production validation

- Route the legacy direct Jasper viewer through the common content-processor chain while preserving the original `ProcessInfo`.
- Clear stale Jasper print data when regenerating or switching reports so the generic report path does not fail with `No print data`.
- Limit the viewer format selector to formats supported by the active renderer or report engine.
- Preserve record UUIDs when forwarding report context through the viewer and content-rendering pipeline.
- Multi-record Jasper handling is intentionally tracked separately in IDEMPIERE-7073.

## Summary by CodeRabbit

* **New Features**
  * Added unified report generation for PDF, HTML, CSV, XLS, XLSX, and SSV formats.
  * Added ranked renderers, post-processing, output caching, and row-count support.
  * Expanded report preview, export, email attachment, and archive handling across workflows.

* **Bug Fixes**
  * Report-generation failures now provide explicit errors instead of silent failures.
  * Improved consistency, temporary-file cleanup, and processed report handling.

* **Tests**
  * Added coverage for renderer selection, fallback behavior, processor ordering, validation, and caching.


hengsin:
Is this for Jasper Report only ? If yes, the class/interface naming and jira ticket title is misleading.

mbozem:
I apologize for the many changes. The scope became larger than I originally expected, so I need to thoroughly test everything again. I’ll follow up once I’m confident that the PR is ready in its current form.

mbozem:
Implemented the remaining applicable CodeRabbit findings in commit 6ec10e0ab2:\n\n- DelimitedReportRenderer now propagates a failed createdDelimitedOutput result.\n- ReportEngine HTML, CSV, XLS, and XLSX export failures now propagate consistently.\n- ZkJRViewer email and archive paths now use the processed PDF pipeline.\n- DunningPrint and PayPrint temporary files are cleaned up after use; PayPrint uses an explicit .pdf suffix.\n- JasperPrintRenderer now uses AdempiereException consistently.\n- The renderer test now asserts that the processed renderer exists before use.\n\nThe null-content handling in WGenForm and ProcessInfoDialog reports an explicit affected record instead of silently dropping a document.\n\nI did not add the proposed public Core overload. The existing request-based API already carries the same context, and the remaining duplication is handled by the private ServerReportCtl helper without expanding the public API.

hengsin (dismissed):
`mvn integration-test -DskipTests=false` fail with the following errors:

```
Failures:
  ReportTest.testPDFFileName:110 expected: <Invoice_Header.pdf> but was: <60000.pdf>
Errors:
  ServerReportCtlTest.testOrderPrint:66 » Adempiere org.adempiere.exceptions.AdempiereException: java.lang.IllegalStateException: No print data (Delete Print Format and restart)
```

org.adempiere.base/src/org/adempiere/base/Core.java
```diff
@@ -1231,6 +1245,211 @@ public static IReportRenderer<IReportRendererConfiguration> getReportRenderer(St
...
+		for (IServiceReferenceHolder<IReportContentRendererFactory> reference : references) {
+			IReportContentRendererFactory factory = reference.getService();
+			if (factory != null) {
+				IReportContentRenderer renderer = factory.createRenderer(request);
+				if (renderer != null)
+					return renderer;
+			}
+		}
+		return null;
+	}
```
hengsin: Not a must but I now prefer OSGi provider component+`@Reference` over the `Service.locator()` api.
Example in core: 
* https://github.com/idempiere/idempiere/blob/master/org.adempiere.base/src/org/adempiere/base/PayScheduleManager.java
* https://github.com/idempiere/idempiere/blob/master/org.adempiere.base/src/org/adempiere/base/StorageValidatorProvider.java
* https://github.com/idempiere/idempiere/blob/master/org.adempiere.base/src/org/compiere/dbPort/SQLStatementRewriterProvider.java.

org.adempiere.base/src/org/adempiere/base/Core.java
```diff
@@ -1231,6 +1245,211 @@ public static IReportRenderer<IReportRendererConfiguration> getReportRenderer(St
...
+	 * @return processing renderer or {@code null}
+	 */
+	public static IReportContentRenderer getProcessedReportContentRenderer(ReportContentRequest request) {
+		if (request == null)
+			return null;
+		IReportContentRenderer renderer = getReportContentRenderer(request);
+		if (renderer == null && request.reportEngine() != null)
+			renderer = new ReportEngineContentRenderer(request.reportEngine());
+		return renderer != null ? new ProcessedReportContentRenderer(request, renderer) : null;
+	}
```
hengsin, line 1272: This getProcessedReportContentRenderer and getReportContentRenderer API is confusing. Why do we need the 2 instead of 1 API here ?

hengsin: I think the cleaner approach is to have just the getReportContentRenderer API and have a flag in ReportContentRequest to turn on/off the post processing of report content.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/Extensions.java
```diff
@@ -420,8 +420,7 @@ public static List<IReportViewerRenderer> getReportViewerRenderers() {
 		List<IServiceReferenceHolder<IReportViewerRenderer>> references = Service.locator().list(IReportViewerRenderer.class, null, null).getServiceReferences();
 		return references.stream().filter(e -> e.getService() != null).map(e -> e.getService()).collect(Collectors.toCollection(ArrayList::new));
 	}
-	
-	
+
```
hengsin: please remove this white space only changes.

mbozem:
Addressed in a800f15fc6.

- PDF content generation now honors ProcessInfo.getPDFFileName(), instead of always using the generated report name.
- ReportEngine.setPrintFormat() no longer clears existing PrintData. This preserves the server-side document print flow that caused ServerReportCtlTest.testOrderPrint to fail.
- For the ZK viewer, non-Jasper print data is refreshed explicitly after a print-format change by reapplying the current query.

Validation after the final change:
- ServerReportCtlTest: 1 test, 0 failures, 0 errors.
- ReportViewerContentRendererFactoryTest: 9 tests, 0 failures, 0 errors.
- The complete 63-module Maven verify reactor finished with BUILD SUCCESS.

mbozem:
No, this is not limited to Jasper reports.

The report-content API is intentionally renderer-independent. Jasper is one provider implementation, while other OSGi services can provide renderers for the same request contract. If no registered factory handles the request, Core falls back to the standard ReportEngine for PDF, HTML, CSV, XLS, and XLSX output.

The current PR title and the generic interface names reflect that broader scope.

hengsin:
@CarlosRuiz-globalqss , this looks fine to me. Routing to you for further review and testing as you have posted in the ticket that you would like to test this when it is ready for review.

CarlosRuiz-globalqss (approved):
tested successfully

PR #1005, opened 2021-11-25
IDEMPIERE-5040 Stall M_InOutLine.QtyOverReceipt

hengsin:
https://idempiere.atlassian.net/browse/IDEMPIERE-5040

PR #3327, opened 2026-08-11
IDEMPIERE-7075 Fix reservations for over-receipt reversals

mbozem:
## Description

Fix two related reservation corruptions triggered by reversing an over-received material receipt and subsequently closing its order.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7075

## Problem 1: large negative storage reservation

For an over-receipt reversal, the previous code could change `storageReservationToUpdate` from the negative reversal movement quantity to a positive capped quantity. The value was negated again when passed to `MStorageReservation.add()`.

Instead of restoring the reservation, this deducted it a second time and could create a large negative Ordered Qty in `M_StorageReservation` (for example `-1041.55`). These invalid values affected inventory and reservation reports.

## Problem 2: incorrect order-line reservation

Correcting only the sign exposed a second issue. Reversing an over-receipt could restore the complete receipt quantity to `C_OrderLine.QtyReserved`, even when it exceeded the remaining open order quantity.

For example:

- Purchase order quantity: 300
- Over-receipt quantity: 303
- Reservation after receipt: 0
- Incorrect reservation after reversal: 303 instead of 300

For purchase receipts, the `M_MatchPO` reversal updates `QtyDelivered` before the relevant `MInOut.completeIt()` logic runs. The previous safeguard therefore saw `QtyDelivered = 0` and did not detect the over-reservation. When the order was closed, `MOrder.reserveStock()` cleared 303 from a storage reservation containing only 300, leaving `M_StorageReservation` at `-3`.

The order line was therefore already incorrect before closing the order, and closing propagated that inconsistency into storage.

## Root cause

The order-line and storage reservations were calculated through separate sign and capping logic. The calculation did not consistently limit the restored quantity to the actual open order quantity, and it did not account for the different `QtyDelivered` update timing of purchase receipts and sales shipments.

## Changes

- Calculate a single `reservationDelta` for both `C_OrderLine.QtyReserved` and `M_StorageReservation`.
- Calculate the target reservation from the current reservation plus that delta.
- Bound the target reservation to the range from zero to the actual open quantity: `max(0, QtyOrdered - anticipated QtyDelivered)`.
- Account for purchase receipt matching updating `QtyDelivered` before this code, while sales shipment processing updates it later.
- Use the same bounded delta for the order line and storage so both remain consistent.

### Complete over-receipt reversal

- Complete a purchase order for 300.
- Receive 303.
- Void the receipt.
- Close the order.
- Verify the closed order line has `QtyOrdered = 0`, `QtyDelivered = 0`, `QtyLostSales = 300`, and `QtyReserved = 0`.
- Verify the storage reservation is 0.

### Partial receipts and partial reversal

- Complete a purchase order for 200.
- Receive 150.
- Receive another 60, producing a total receipt of 210.
- Void the receipt for 60.
- Verify `QtyDelivered = 150` and `QtyReserved = 50`, rather than incorrectly restoring 60.
- Close the order.
- Verify the closed order line has `QtyOrdered = 150`, `QtyDelivered = 150`, `QtyLostSales = 50`, and `QtyReserved = 0`.
- Verify the storage reservation is 0.

### Reservation isolation between two orders

- Complete two orders for the same product and warehouse.
- Over-receive, reverse, and close only the first order.
- Verify the second order line remains reserved for its complete quantity.
- Verify the shared `M_StorageReservation` retains exactly the second order's reservation.
- Close the second order and verify the combined storage reservation returns to its initial value.

## Impact

Receipt reversals no longer create large negative storage reservations or inflate the related order-line reservation. Closing an affected order clears only its valid remaining reservation and does not corrupt reservations belonging to another order for the same product and warehouse.

## Validation

- Full Maven reactor: `BUILD SUCCESS`
- `org.idempiere.test`: `SUCCESS`
- `git diff --check upstream/master...HEAD`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reservation handling when purchase receipts exceed ordered quantities.
  * Corrected reservation restoration when over-receipts or over-shipments are reversed or orders are closed.
  * Prevented reservation changes from affecting unrelated orders.
  * Improved reservation tracking for service products during shipment completion and reversal.

* **Tests**
  * Added coverage for over-receipts, reversals, closures, partial receipts, service products, and over-shipments.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -1693,6 +1693,7 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 	
 				//	Update Order Line
 				MOrderLine oLine = null;
+				BigDecimal reservationDelta = Env.ZERO;
 				if (sLine.getC_OrderLine_ID() != 0)
 				{
 					oLine = new MOrderLine (getCtx(), sLine.getC_OrderLine_ID(), get_TrxName());
@@ -1744,21 +1745,27 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 								toDelivered = Env.ZERO;
 						}
 					} 
+
+					reservationDelta = sLine.getMovementQty().negate();
 					
-					BigDecimal storageReservationToUpdate = sLine.getMovementQty();
-					if (oLine != null)
+					if (oLine != null && oLine.getQtyOrdered().signum() >= 0)
 					{
-						if (!isReversal()) 
-						{
-							if (storageReservationToUpdate.compareTo(oLine.getQtyReserved()) > 0) 
-								storageReservationToUpdate = oLine.getQtyReserved();
-						}
-						else
-						{
-							BigDecimal tmp = storageReservationToUpdate.negate().add(oLine.getQtyReserved());
-							if (tmp.compareTo(oLine.getQtyOrdered()) > 0)
-								storageReservationToUpdate = oLine.getQtyOrdered().subtract(oLine.getQtyReserved());
-						}
+						BigDecimal currentReserved = oLine.getQtyReserved();
+						BigDecimal targetReserved = currentReserved.add(reservationDelta);
+						BigDecimal anticipatedDelivered = oLine.getQtyDelivered();
+						if (isSOTrx())
+							anticipatedDelivered = anticipatedDelivered.add(sLine.getMovementQty());
+						BigDecimal maxReserved = oLine.getQtyOrdered().subtract(anticipatedDelivered);
+
+						if (targetReserved.signum() < 0)
+							targetReserved = Env.ZERO;
+						if (maxReserved.signum() < 0)
+							maxReserved = Env.ZERO;
+
+						if (targetReserved.compareTo(maxReserved) > 0)
+							targetReserved = maxReserved;
+
+						reservationDelta = targetReserved.subtract(currentReserved);
 					}
 					
 					//
@@ -1833,7 +1840,7 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 								if (!MStorageReservation.add(getCtx(), oLine.getM_Warehouse_ID(),
 										oLine.getM_Product_ID(),
 										oLine.getM_AttributeSetInstance_ID(),
-										storageReservationToUpdate.negate(),
+										reservationDelta,
 										isSOTrx(),
 										get_TrxName(), tracer))
 								{
@@ -1929,7 +1936,7 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 							if (!MStorageReservation.add(getCtx(), oLine.getM_Warehouse_ID(),
 									oLine.getM_Product_ID(),
 									oLine.getM_AttributeSetInstance_ID(),
-									storageReservationToUpdate.negate(), isSOTrx(), get_TrxName(), tracer))
+									reservationDelta, isSOTrx(), get_TrxName(), tracer))
 							{
 								m_processMsg = "Cannot correct Inventory Reserved " + (isSOTrx()? "Reserved [" :"Ordered [") + product.getValue() + "]";
 								return DocAction.STATUS_Invalid;
@@ -1963,12 +1970,7 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 				{
 					if (oLine.getQtyOrdered().signum() >= 0)
 					{
-						oLine.setQtyReserved(oLine.getQtyReserved().subtract(sLine.getMovementQty()));
-
-						if (oLine.getQtyReserved().signum() == -1)
-							oLine.setQtyReserved(Env.ZERO);
-						else if (oLine.getQtyDelivered().compareTo(oLine.getQtyOrdered()) > 0)
-							oLine.setQtyReserved(Env.ZERO);
+						oLine.setQtyReserved(oLine.getQtyReserved().add(reservationDelta));
 					}
 				}
 	
--- a/org.adempiere.base/src/org/compiere/model/MInOut.java
+++ b/org.adempiere.base/src/org/compiere/model/MInOut.java
@@ -1700,6 +1700,25 @@ public class MInOut extends X_M_InOut implements DocAction, IDocsPostProcess
 					if (log.isLoggable(Level.FINE)) log.fine("OrderLine - Reserved=" + oLine.getQtyReserved()
 						+ ", Delivered=" + oLine.getQtyDelivered());
 				}
+				if (oLine != null && oLine.getQtyOrdered().signum() >= 0)
+				{
+					BigDecimal currentReserved = oLine.getQtyReserved();
+					BigDecimal targetReserved = currentReserved.subtract(sLine.getMovementQty());
+					BigDecimal anticipatedDelivered = oLine.getQtyDelivered();
+					if (isSOTrx())
+						anticipatedDelivered = anticipatedDelivered.add(sLine.getMovementQty());
+					BigDecimal maxReserved = oLine.getQtyOrdered().subtract(anticipatedDelivered);
+
+					if (targetReserved.signum() < 0)
+						targetReserved = Env.ZERO;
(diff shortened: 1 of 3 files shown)


hengsin:
We have implemented a solution by adding QtyOverReceipt column to M_InOutLine. QtyOverReceipt is computed in the completeIt of MInOut. When reverse, that's use to compute the correct qty reserved to reverse.

mbozem:
Thanks for the suggestion. However, `M_InOutLine.QtyOverReceipt` was stalled and deactivated with IDEMPIERE-5040 (#1005). The stored value can become stale when an order is reactivated and its quantity is changed after the receipt.

Re-enabling and relying on this column could therefore reintroduce the consistency problem addressed by IDEMPIERE-5040. The current patch calculates the reservation correction from the current order-line state instead.

Do you still prefer reactivating `QtyOverReceipt` for IDEMPIERE-7075? If so, how should the stale-value scenario from IDEMPIERE-5040 be handled?

hengsin:
@mbozem , thanks for reminding me on that. Yeah, sounds better if we can calculate the over receipt on the fly instead.

org.adempiere.base/src/org/compiere/model/MInOut.java
```diff
@@ -1693,12 +1693,32 @@ public String completeIt()
...
+					if (targetReserved.signum() < 0)
+						targetReserved = Env.ZERO;
+					if (maxReserved.signum() < 0)
+						maxReserved = Env.ZERO;
+
+					if (targetReserved.compareTo(maxReserved) > 0)
+						targetReserved = maxReserved;
+
+					reservationDelta = targetReserved.subtract(currentReserved);
+				}
```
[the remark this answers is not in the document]
mbozem: Addressed in a08a8dac7c. Both storage-reservation paths are now gated by `reservationDelta.signum() != 0`, so a positive delta is applied even when the current order-line reservation is zero and `M_StorageReservation` remains synchronized with `C_OrderLine.QtyReserved`.

The regression test now covers an order for 1, an over-shipment of 2, and two subsequent shipments of -1. It verifies both the order-line reservation and storage reservation after each step, including restoration to 1 after the second negative shipment.

The full local Maven test run passed. `SalesOrderTest` completed 21 tests with 0 failures and 0 errors, including `testQtyReservedForOverAndNegativeShipment`.

org.idempiere.test/src/org/idempiere/test/model/SalesOrderTest.java
```diff
@@ -213,14 +261,18 @@ public void testQtyReservedForOverAndNegativeShipment() {
...
+		line1.saveEx();
+		BigDecimal initialQtyReserved = MStorageReservation.getQty(line1.getM_Product_ID(),
+				line1.getM_Warehouse_ID(), 0, true, getTrxName());
 		
 		ProcessInfo info = MWorkflow.runDocumentActionWorkflow(order, DocAction.ACTION_Complete);
 		assertFalse(info.isError(), info.getSummary());
 		order.load(getTrxName());		
 		assertEquals(DocAction.STATUS_Completed, order.getDocStatus());
 		line1.load(getTrxName());
 		assertEquals(1, line1.getQtyReserved().intValue());
```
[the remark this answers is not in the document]
mbozem: Addressed in 60b67a4e77. All `QtyReserved` and `QtyDelivered` checks in this scenario now compare the complete `BigDecimal` value with `compareTo()` instead of truncating through `intValue()`. I also added specific failure messages to the order-line and storage-reservation assertions.

Validation: `mvn -DskipTests=false -Dtest=SalesOrderTest verify` completed with `BUILD SUCCESS`; `SalesOrderTest` ran 21 tests with 0 failures, 0 errors, and 0 skipped tests, including `testQtyReservedForOverAndNegativeShipment`.

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
```diff
@@ -586,6 +586,368 @@ public void testQtyOverReceipt() {
...
+		Properties ctx = Env.getCtx();
+		String trxName = getTrxName();
+		int productId = DictionaryIDs.M_Product.ROSE_BUSH.id;
+		BigDecimal qtyOrdered = new BigDecimal("300");
+		BigDecimal qtyReceived = new BigDecimal("303");
+		BigDecimal initialQtyOrdered = getQtyOrdered(ctx, productId, trxName);
+
+		DB.executeUpdateEx("UPDATE AD_SysConfig SET Value='N' WHERE AD_Client_ID=0 AND Name=?",
+				new Object[] {MSysConfig.VALIDATE_MATCHING_TO_ORDERED_QTY}, null);
+		CacheMgt.get().reset();
```
hengsin: Should use mock instead of update here.

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
```diff
@@ -586,6 +586,368 @@ public void testQtyOverReceipt() {
...
+			info = MWorkflow.runDocumentActionWorkflow(receipt, DocAction.ACTION_Complete);
+			assertFalse(info.isError(), info.getSummary());
+			receipt.load(trxName);
+			assertEquals(DocAction.STATUS_Completed, receipt.getDocStatus(), "Material receipt not completed");
+			orderLine.load(trxName);
+			assertEquals(0, Env.ZERO.compareTo(orderLine.getQtyReserved()), "Order line reservation not cleared by over-receipt");
+			assertEquals(0, initialQtyOrdered.compareTo(getQtyOrdered(ctx, productId, trxName)),
+					"Storage reservation not cleared by over-receipt");
+
+			info = MWorkflow.runDocumentActionWorkflow(receipt, DocAction.ACTION_Void);
```
hengsin: Clearer to just use reversal action instead.

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
```diff
@@ -586,6 +586,368 @@ public void testQtyOverReceipt() {
...
+		int productId = DictionaryIDs.M_Product.ROSE_BUSH.id;
+		BigDecimal qtyOrdered = new BigDecimal("200");
+		BigDecimal firstReceiptQty = new BigDecimal("150");
+		BigDecimal secondReceiptQty = new BigDecimal("60");
+		BigDecimal openQtyAfterFirstReceipt = new BigDecimal("50");
+		BigDecimal initialQtyOrdered = getQtyOrdered(ctx, productId, trxName);
+
+		DB.executeUpdateEx("UPDATE AD_SysConfig SET Value='N' WHERE AD_Client_ID=0 AND Name=?",
+				new Object[] {MSysConfig.VALIDATE_MATCHING_TO_ORDERED_QTY}, null);
+		CacheMgt.get().reset();
```
hengsin: As above, use mockito instead of update.

hengsin (changes requested):

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
mbozem: Addressed in `8765c3472b`. The affected tests now mock `MSysConfig.getBooleanValue(...)` with Mockito and no longer update `AD_SysConfig` or reset the global cache.

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
mbozem: Addressed in `8765c3472b`. The tests now invoke `DocAction.ACTION_Reverse_Correct` directly, and the test names and assertion messages have been updated accordingly.

org.idempiere.test/src/org/idempiere/test/model/PurchaseOrderTest.java
mbozem: Addressed in `8765c3472b`. All four affected purchase-order tests now use a scoped Mockito static mock instead of changing `AD_SysConfig`. Validation: all 16 `PurchaseOrderTest` tests passed with 0 failures and errors; the full Maven reactor completed with `BUILD SUCCESS`.

hengsin (approved):
looks good and pass all unit tests.

PR #3329, opened 2026-08-14
IDEMPIERE-7073 Support multiple record IDs in Jasper reports

mbozem:
## Description

Add explicit and backward-compatible support for running Jasper reports for multiple selected records.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7073

## Root cause

The Jasper report starter treated multiple numeric record IDs through query rewriting only. This did not support UUID-key tables, reports that need to run once per record, or reports that consume `T_Selection`. Multiple `JasperPrint` results were also not preserved consistently for the web report viewer.

## Changes

- Add `AUTO`, `ITERATE`, and `T_SELECTION` record-selection behavior through the `idempiere.recordSelectionMode` Jasper report property.
- Pass sanitized, de-duplicated selections as `RECORD_IDS` or `RECORD_UUS` collection parameters.
- Support scalar `RECORD_ID` and `RECORD_UU` parameters when iterating over selected records.
- Populate the appropriate `T_Selection_ID` or `T_Selection_UU` column for `T_SELECTION` reports.
- Preserve the existing query-filtering behavior as the compatibility fallback.
- Preserve one or multiple `JasperPrint` results as serializable process output and render both forms in the ZK report viewer.
- Add integration coverage for numeric IDs, UUIDs, selection sanitization, `T_Selection`, legacy query filtering, and invalid configuration.

## Impact

Jasper reports can opt into per-record execution or `T_Selection` while existing reports continue to use automatic compatibility behavior. Both numeric and UUID-key records are supported.

## Validation

- `mvn verify -DskipTests`
- `mvn verify -DskipTests=false -Dtest=org.idempiere.test.jasper.JasperMultipleRecordTest`
- 6 tests run, 0 failures, 0 errors
- `git diff --check upstream/master...HEAD`

### DIFF
--- a/org.adempiere.report.jasper/src/org/adempiere/report/jasper/ReportStarter.java
+++ b/org.adempiere.report.jasper/src/org/adempiere/report/jasper/ReportStarter.java
@@ -30,7 +30,9 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


mbozem:
Withdrawn by the author. The existing T_Selection mechanism already covers the intended use case, so this patch is no longer needed.

PR #3331, opened 2026-08-16
IDEMPIERE-7081 Stop SSO callback processing after redirect

mbozem:
## Description

Stop processing an SSO authentication callback after its response has been redirected or committed.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7081

## Root cause

`SSOWebUIFilter` continued to `chain.doFilter(...)` after handling an authentication code. ZK could then try to render the callback request through a response that had already been redirected, causing Jetty to log `EofException: Closed`.

The same continuation occurred when `getAuthenticationToken()` committed a redirect itself, such as for an invalid OIDC state.

## Changes

- Return unconditionally after the authentication-code branch completes.
- Preserve the existing authentication and redirect behavior.

## Impact

SSO callback requests are handled only once. ZK no longer attempts to write a second response after the callback redirect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SSO authentication redirects to prevent unintended additional processing after the redirect.
  * Improved reliability of the sign-in flow by stopping request handling at the correct point.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/sso/filter/SSOWebUIFilter.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/sso/filter/SSOWebUIFilter.java
@@ -228,6 +228,7 @@ public class SSOWebUIFilter implements Filter
 							}
 							httpResponse.sendRedirect(currentUri);
 						}
+						return;
 					}
 					else if (!m_SSOPrincipal.isAuthenticated(httpRequest, httpResponse))
 					{


org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/sso/filter/SSOWebUIFilter.java
```diff
@@ -228,6 +228,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 							}
 							httpResponse.sendRedirect(currentUri);
 						}
+						return;
```
hengsin, line 231: Probably safer to add the return right after the `httpResponse.sendRedirect(currentUri);` call above.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/sso/filter/SSOWebUIFilter.java
mbozem: Thanks for pointing this out. I agree that having the `return` immediately after `sendRedirect(currentUri)` makes the control flow easier to understand.

The unconditional `return` after the `if (!httpResponse.isCommitted())` block is intentional, though. `getAuthenticationToken(...)` may already have handled the response itself. For example, the OIDC implementation redirects when the returned `state` parameter is invalid or does not match the value stored in the session. In that case, the response can already be committed and the outer redirect must not be executed again.

The filter must also return in that case to prevent `chain.doFilter(...)` from rendering the original SSO callback request.

To make both cases explicit, the code could be structured as follows:

```java
if (httpResponse.isCommitted())
    return;

// Redirect to the original request URL after authentication
// and preserve the query string.
Object queryString = httpRequest.getSession()
        .getAttribute(ISSOPrincipalService.SSO_QUERY_STRING);
if (queryString != null && queryString instanceof String
        && !Util.isEmpty((String) queryString))
    currentUri += "?" + (String) queryString;

httpRequest.getSession().removeAttribute(ISSOPrincipalService.SSO_QUERY_STRING);

if (request.getParameter(TENANT_PREFIX_PARAMETER) != null
        && (queryString == null
                || !queryString.toString().contains(TENANT_PREFIX_PARAMETER + "="))) {
    currentUri += (currentUri.contains("?") ? "&" : "?")
            + TENANT_PREFIX_PARAMETER + "="
            + request.getParameter(TENANT_PREFIX_PARAMETER);
}

httpResponse.sendRedirect(currentUri);
return;
```

This keeps the immediate `return` after the redirect while also preserving the early exit when the SSO provider has already committed the response. Would this structure address your concern?

hengsin: You are right, should return too if httpResponse.isCommitted(). No changes needed.

hengsin (approved):
looks good.

PR #3332, opened 2026-08-16
IDEMPIERE-7082 Fix Unicode case-insensitive searches

mbozem:
## Description

Case-insensitive searches can fail for Unicode characters whose Java and database uppercase mappings differ. For example, Java converts `Großformat` to `GROSSFORMAT`, while the database can retain `ß`, so a search for products beginning with `Großformat` returns no results.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7082

## Root cause

Search values were converted to uppercase in Java and then compared with `UPPER(column) LIKE ?`. Java and the configured database do not necessarily use identical Unicode case mappings.

## Changes

- Preserve the original search value in shared model helpers and search dialogs.
- Apply the database case conversion to both operands with `UPPER(column) LIKE UPPER(?)`.
- Cover the common model lookup paths and the affected web and desktop search panels, editors, and dialogs.
- Add regression coverage for the Unicode search value `Großformat` and the shared `PO.getFindParameter` behavior.

## Impact

Prefix and wildcard searches containing Unicode characters now use the database's case mapping consistently. Existing ASCII case-insensitive search behavior and wildcard handling remain unchanged. No database migration is required.

## Validation

- Full 61-module reactor build with focused test execution — BUILD SUCCESS.
- `MQueryTest` and `POFindParameterTest`: 38 tests run, 0 failures, 0 errors.
- Database-backed cases cover `Großformat`, `Straße`, `Änderung`, `Österreich`, `İstanbul`, and an ASCII-only search.
- `git diff --check` passes.
- Static audit found no remaining Java occurrences of `UPPER(column) LIKE ?` in the affected pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved case-insensitive searching across inventory, product, business partner, accounting, asset, document, payment, and other lookup screens.
  * Search results now remain consistent regardless of whether entered text uses uppercase, lowercase, or mixed case.
  * Search fields preserve the text entered by users while continuing to support trailing wildcard matching.
* **Tests**
  * Added coverage for case-preserving search parameters, wildcard handling, and legacy search behavior.

### DIFF
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -5746,10 +5746,12 @@ public abstract class PO
 
 	/**
 	 * 	Convert query value.<br/>
-	 * 	Add % at the end.
+	 * 	Convert to upper case and add % at the end.
 	 *	@param query in string
 	 *	@return converted query value
+	 *	@deprecated use {@link #getFindParameterPreserveCase(String)} instead
 	 */
+	@Deprecated (since="14")
 	protected static String getFindParameter (String query)
 	{
 		if (query == null)
@@ -5758,8 +5760,25 @@ public abstract class PO
 			return null;
 		if (!query.endsWith("%"))
 			query += "%";
-		return query;
+		return query.toUpperCase();
 	}	//	getFindParameter
+
+	/**
+	 * 	Convert query value.<br/>
+	 * 	Preserve case and add % at the end.
+	 *	@param query in string
+	 *	@return converted query value
+	 */
+	protected static String getFindParameterPreserveCase (String query)
+	{
+		if (query == null)
+			return null;
+		if (query.length() == 0 || query.equals("%"))
+			return null;
+		if (!query.endsWith("%"))
+			query += "%";
+		return query;
+	}	//	getFindParameterPreserveCase
 
 	/**
 	 * 	Load LOB
--- a/org.adempiere.base/src/org/compiere/model/PO.java
+++ b/org.adempiere.base/src/org/compiere/model/PO.java
@@ -5746,7 +5746,7 @@ public abstract class PO
 
 	/**
 	 * 	Convert query value.<br/>
-	 * 	Convert to upper case and add % at the end.
+	 * 	Add % at the end.
 	 *	@param query in string
 	 *	@return converted query value
 	 */
@@ -5758,7 +5758,7 @@ public abstract class PO
 			return null;
 		if (!query.endsWith("%"))
 			query += "%";
-		return query.toUpperCase();
+		return query;
 	}	//	getFindParameter
 
 	/**
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WLocatorEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WLocatorEditor.java
@@ -383,8 +383,7 @@ public class WLocatorEditor extends WEditor implements EventListener<Event>, Pro
 		
 		//	Look up - see MLocatorLookup.run		
 		StringBuffer sql = new StringBuffer("SELECT M_Locator_ID FROM M_Locator ")
-			.append(" WHERE IsActive='Y' AND UPPER(Value) LIKE ")
-			.append("UPPER(").append(DB.TO_STRING(text)).append(")");
+			.append(" WHERE IsActive='Y' AND UPPER(Value) LIKE UPPER(?)");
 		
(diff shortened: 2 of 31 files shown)


org.adempiere.ui/src/org/compiere/apps/form/StatementCreateFromBatch.java
```diff
@@ -160,7 +160,7 @@ protected void setParameters(PreparedStatement pstmt, Integer BankAccount, Integ
 			pstmt.setString(index++, getSQLText(DocumentNo));
 		
 		if(AuthCode.length() > 0)
-			pstmt.setString(index++, getSQLText(AuthCode));
+			pstmt.setString(index++, getSQLText(AuthCode).toUpperCase());
```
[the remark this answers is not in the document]
mbozem: Thanks for the review. This behavior is intentionally preserved and is outside the scope of IDEMPIERE-7082.

In the upstream implementation, `StatementCreateFromBatch` uses `p.R_AuthCode LIKE ?` and uppercases the parameter through the shared `getSQLText()` helper. Unlike `DocumentNo`, this comparison did not combine Java uppercasing with `UPPER(column)`, which is the specific mismatch addressed by this patch.

The Java uppercasing was moved to the `AuthCode` binding only because `getSQLText()` must now preserve the original value for `DocumentNo`. This keeps the existing `R_AuthCode` behavior unchanged.

Changing the predicate to `UPPER(p.R_AuthCode) LIKE UPPER(?)` would introduce new case-insensitive semantics and could affect index usage. I suggest handling that independently if case-insensitive authorization-code searches are desired. The separate `CreateFromBatch` path was changed because its original predicate already used `UPPER(p.R_AuthCode)`.

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WLocatorEditor.java
```diff
@@ -378,15 +378,13 @@ private boolean actionText(int only_Warehouse_ID, int only_Product_ID)
...
-		else
-			text = text.toUpperCase() + "%";
+		if (!text.endsWith("%"))
+			text += "%";
 		
 		//	Look up - see MLocatorLookup.run		
 		StringBuffer sql = new StringBuffer("SELECT M_Locator_ID FROM M_Locator ")
 			.append(" WHERE IsActive='Y' AND UPPER(Value) LIKE ")
-			.append(DB.TO_STRING(text));
+			.append("UPPER(").append(DB.TO_STRING(text)).append(")");
```
CarlosRuiz-globalqss: it would be good to change this to use "LIKE UPPER(?)" with binding variable

org.adempiere.base/src/org/compiere/model/PO.java
CarlosRuiz-globalqss, line 5755: Because this is a protected class, there is a possibility this method has been used in plugins.

So, for backward compatibility I think is better to mark this method as deprecated with a note about how to replace it, and create a new method.

CarlosRuiz-globalqss (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/editor/WLocatorEditor.java
mbozem: Thanks, addressed in commit `cae980e45d`. The query now uses `UPPER(Value) LIKE UPPER(?)`, and the search text is passed as a prepared-statement parameter instead of being embedded in the SQL.

org.adempiere.base/src/org/compiere/model/PO.java
mbozem: Thanks, addressed in commit `cae980e45d`. I restored the original behavior of `getFindParameter(String)` for backward compatibility and marked it as deprecated with `getFindParameterPreserveCase(String)` as its replacement. The affected core callers now use the new method, and tests cover both the legacy uppercase behavior and the new case-preserving behavior.

mbozem:
Thanks for the review. Both requested changes have been addressed in `cae980e45d`. The full Maven build passes, including all 9 `POFindParameterTest` tests.

mbozem:
The unintended changes from commit 1500354fd6 have been fully reverted in 4d0958822f. The PR content is now identical to cae980e45d, where both getSQLText Javadocs are already corrected. The unrelated database-search abstraction and SSO filter change are no longer part of this PR. @coderabbitai review

mbozem:
Hi Carlos, both requested changes have been addressed, and the corresponding review threads are now resolved. The current PR content is ready for review. Could you please take another look and update your review decision? Thanks!

PR #3339, opened 2026-08-22
IDEMPIERE-7084 Use node-local connector for WebSocket AU requests

mbozem:
## Description

The WebSocket server-push endpoint forwards ZK AU requests to `/zkau` through an internal HTTP client. That client used `localhost`, which fails when the application server connector is bound only to a non-loopback address. The initial login echo request then never completes and the login fields remain disabled. Switching server push to Atmosphere hides the problem because that implementation does not use this WebSocket forwarding path.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7084

## Root cause

`ServerPushEndPoint` derived its internal AU URL from the WebSocket request but replaced the host with `localhost`. This assumes that the servlet connector always listens on the loopback interface.

## Changes

- Capture the actual local servlet connector address, port, and transport when WebSocket server push starts.
- Store the node-local base URL per desktop in the HTTP session.
- Use that URL for internal `/zkau` requests and cookie handling.
- Keep a `localhost` compatibility fallback for desktops created before deployment or when connector metadata is unavailable.
- Remove the per-desktop session value when server push stops.
- Build the URL through `URI` so IPv4 and IPv6 addresses are handled correctly.

## Cluster and reverse-proxy behavior

Internal AU requests remain on the node that owns the ZK desktop and do not pass through a load balancer, public hostname, or cluster VIP. The connector transport is determined independently of proxy-rewritten browser-facing request metadata.

## Validation

- Full Maven build including tests passed locally.
- WebSocket login was tested successfully on a connector bound to a non-loopback address.
- `git diff --check upstream/master...HEAD` passes.
- The PR contains only the three WebSocket server-push files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket server-push connectivity across local and remote deployments.
  * Correctly detects the active server host, port, and secure connection settings.
  * Fixed session and response cookie domains to use the active server host instead of `localhost`.
  * Added compatibility handling for invalid or unavailable connection details.
  * Improved connection handling when applications run behind different network configurations or server nodes.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
@@ -40,6 +40,13 @@ import org.apache.hc.client5.http.impl.cookie.BasicClientCookie;
 
 public class EndpointConfigurator extends ServerEndpointConfig.Configurator {
 
+	/**
+	 * Resolve the node-local HTTP base URI used to forward WebSocket events.
+	 *
+	 * @param httpSession HTTP session containing the per-desktop connector URI
+	 * @param request WebSocket handshake request
+	 * @return node-local base URI, or a localhost compatibility URI
+	 */
 	static URI getInternalBaseUri(HttpSession httpSession, HandshakeRequest request) {
 		URI requestUri = request.getRequestURI();
 		if (httpSession != null && requestUri != null) {
@@ -72,6 +79,7 @@ public class EndpointConfigurator extends ServerEndpointConfig.Configurator {
 		}
 	}
 
+	/** {@inheritDoc} */
 	@Override
     public void modifyHandshake(ServerEndpointConfig sec, HandshakeRequest request, HandshakeResponse response) {
         // Store the ServletContext, HttpSession and remote address in user properties for the ServerPushEndPoint to access
--- a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
@@ -58,8 +58,9 @@ public class EndpointConfigurator extends ServerEndpointConfig.Configurator {
 				if (value instanceof String localBaseUrl) {
 					try {
 						URI uri = URI.create(localBaseUrl);
+						int port = uri.getPort();
 						if (("http".equals(uri.getScheme()) || "https".equals(uri.getScheme()))
-								&& uri.getHost() != null && uri.getPort() > 0)
+								&& uri.getHost() != null && (port == -1 || port > 0))
 							return uri;
 					} catch (IllegalArgumentException e) {
 						// Ignore invalid session data and use the compatibility fallback below.
--- a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
@@ -24,6 +24,8 @@
  **********************************************************************/
 package org.idempiere.ui.zk.websocket;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


mbozem:
@coderabbitai review

CarlosRuiz-globalqss (approved):
I did some tests in the server test-feature.idempiere.org

There are three issues mentioned in the migration notes:
https://wiki.idempiere.org/en/Migration_Notes#WebSocket_as_default_Server_Push_in_zkwebui

According to my tests, the last two are solved with this pull request:

**Solves** -> Port redirection via iptables: Redirecting port 8443 to 443 using iptables rules can break WebSocket handshake negotiation
**Solves** -> Connecting using a ssh tunnel for port 8443 doesn't work

**This one I CANNOT TEST** -> LXC containers: WebSocket connections may not work correctly depending on the host kernel and network configuration of the LXC environment



So, for me this is OK, but it would be better to have a code peer review from @hengsin before moving this ahead

mbozem:
## LXC regression test report

I tested the previously unverified LXC scenario mentioned in the migration notes. The HTTPS/WSS regression is reproducible without this pull request and is fixed by commit `82f0eb1a6647adbae69b857cbf64dc9fb4180b7d`.

### Environment

- Classic LXC container on Ubuntu 24.04 LTS
- Host/container kernel: Linux `6.8.0-138-generic`, x86_64
- Container address: `10.0.3.221/24` on a veth interface attached to `lxcbr0`
- OpenJDK `17.0.19`
- PostgreSQL `16.15`
- iDempiere Dev14 daily server from 2026-08-18 and its matching PostgreSQL seed
- Google Chrome `151.0.7922.173` in headless/CDP mode
- Firefox used for an additional interactive login test

### Test setup

The application server was configured with:

- `ADEMPIERE_APPS_SERVER=10.0.3.221`
- `ADEMPIERE_WEB_ALIAS=10.0.3.221`
- HTTP connector: `10.0.3.221:8080`
- HTTPS connector: `10.0.3.221:8443`
- Default WebSocket server push; no Atmosphere override

`ss` confirmed that Jetty listened only on the container address for ports 8080 and 8443, not on `127.0.0.1` or `::1`. This is the relevant LXC/non-loopback condition: a browser can reach the container connector, while an internal request hardcoded to `localhost` cannot.

To keep every other variable constant, I used the same runtime, database, connector configuration, and browser instrumentation for both runs:

1. **Baseline:** the unmodified WebSocket classes from the 2026-08-18 Dev14 daily build.
2. **Pull request:** the `org.idempiere.ui.zk` WebSocket package compiled from PR commit `82f0eb1a6647adbae69b857cbf64dc9fb4180b7d` and installed into the same runtime.

The required `org.adempiere.ui.zk-14.0.0-SNAPSHOT.jar` was built successfully from the PR checkout. A later, unrelated Core feature packaging step failed because the test environment could not resolve the `javax.xml` IU; this occurred after the Web UI bundle used for the regression test had been produced.

### Automated HTTPS/WSS test

The browser opened:

`https://10.0.3.221:8443/webui/`

Chrome DevTools Protocol instrumentation recorded the WebSocket lifecycle, messages, and final DOM input state after 15 seconds.

#### Baseline result

- WebSocket opened successfully at `wss://10.0.3.221:8443/webui/serverpush/<desktop-id>`.
- Ping/pong succeeded.
- The initial forwarded AU request returned:

  ```json
  {"status":500,"statusText":"Error processing /zkau request"}
  ```

- Username, password, and language controls remained disabled.
- The login page was unusable.

#### Pull request result

- WebSocket opened successfully at `wss://10.0.3.221:8443/webui/serverpush/<desktop-id>`.
- Ping/pong succeeded and the WebSocket remained open.
- The internal `/webui/zkau` request returned HTTP `200`.
- The AU response contained the commands that set the username, password, and language controls to `disabled=false`.
- All three controls were enabled in the final DOM.
- The returned session cookie was processed successfully.

### Interactive Firefox test

An additional manual test in Firefox against the same PR runtime and URL succeeded. The login controls were enabled and logging into iDempiere completed successfully.

### Comparison

| Scenario | Baseline WebSocket classes | PR `82f0eb1a66` |
|---|---|---|
| LXC HTTPS/WSS connection | WebSocket opens | WebSocket opens |
| Initial forwarded `/zkau` request | HTTP 500 / processing error | HTTP 200 |
| Login controls | Remain disabled | Enabled |
| Interactive Firefox login | Not usable in the reproduced baseline condition | Successful |

### Additional HTTP observation

I also tested unencrypted HTTP on `http://10.0.3.221:8080/webui/`. The login controls remained disabled in both the baseline and PR runs. In the PR run Chrome observed WebSocket close code `1003` with reason `Endpoint notification error`. HTTPS/WSS is therefore verified as fixed in this LXC environment; the HTTP-only result is a separate observation and was not investigated to root cause as part of this test.

### Conclusion

The LXC failure documented in the migration notes was reproduced under a connector bound only to the container's non-loopback address. With PR commit `82f0eb1a66`, the same HTTPS/WSS test forwards the AU request through the node-local connector, receives HTTP 200, enables the login controls, and permits a successful Firefox login.

**Result: LXC HTTPS/WSS regression test passed.**

org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
```diff
@@ -38,6 +40,46 @@
...
+			String path = requestUri.getPath();
+			int separator = path != null ? path.lastIndexOf('/') : -1;
+			if (separator >= 0 && separator + 1 < path.length()) {
+				String desktopId = path.substring(separator + 1);
+				Object value = httpSession.getAttribute(WebSocketServerPush.getLocalBaseUrlAttribute(desktopId));
+				if (value instanceof String localBaseUrl) {
+					try {
+						URI uri = URI.create(localBaseUrl);
+						if (("http".equals(uri.getScheme()) || "https".equals(uri.getScheme()))
+								&& uri.getHost() != null && uri.getPort() > 0)
```
hengsin: `uri.getPort()` return -1 if default is use, for e.g `http://127.0.0.1/webui`.

hengsin (changes requested):

org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/websocket/EndpointConfigurator.java
mbozem: Good catch. Although the value produced by getLocalBaseUrl() includes request.getLocalPort(), a URI without an explicit default port is valid and URI.getPort() returns -1 for it. Updated the validation in 36b8cc4198 to accept -1 while continuing to reject port 0.

hengsin (approved):
looks good.

PR #3346, opened 2026-08-31
IDEMPIERE-5431 Extend workflow editor canvas dynamically

mbozem:
## Description

Extend the workflow editor canvas dynamically for workflows that require more than four columns.

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-5431

## Changes

- Keep four columns as the minimum workflow canvas width.
- Size the editor from the highest stored node X position and provide one empty trailing column for further expansion.
- Preserve and render node positions beyond column four instead of wrapping them to the next row.
- Size the read-only workflow panel to existing node positions without adding the editor-only trailing column.
- Reset cached row and column bounds when reloading a workflow.
- Calculate the canvas height from the actual maximum occupied row.

## Impact

Complex workflows can be expanded progressively by dragging a node into the empty rightmost column. Reloading then adds another empty column automatically. No database migration or SysConfig key is required.

## Validation

- Compiled `WFNodeContainer`, `WFEditor`, and `WFPanel` successfully.
- Verified minimum and dynamic container widths, minimum height, and reset behavior with a focused runtime test.
- Verified that the editor uses the dynamic column count and reserves one trailing column.
- Verified that the workflow panel honors stored X positions without reserving an extra column.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow diagrams now automatically adjust their grid width to fit the widest node layout.
  * Improved node placement for workflows with varying numbers of columns.
  * Prevented invalid node positions from causing unexpected layout wrapping.
  * Added safeguards for workflows that exceed supported row or column limits.
  * Resetting or reloading a workflow now correctly clears previous layout dimensions.
  * Drop targets now appear only when an additional workflow row can be added.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/wf/WFEditor.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/adempiere/webui/apps/wf/WFEditor.java
@@ -300,6 +300,11 @@ public class WFEditor extends ADForm {
 
 		//	Add Nodes for Paint
 		MWFNode[] nodes = m_wf.getNodes(true, Env.getAD_Client_ID(Env.getCtx()));
+		int maxColumn = 0;
+		for (MWFNode node : nodes) {
+			maxColumn = Math.max(maxColumn, node.getXPosition());
+		}
+		nodeContainer.setColumnCount(maxColumn + 1);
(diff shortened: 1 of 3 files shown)


mbozem:
@CarlosRuiz-globalqss, could you please review this PR?

CarlosRuiz-globalqss (approved):
tested successfully

PR #3349, opened 2026-09-01
IDEMPIERE-7070 Add interactive HTML report content type

mbozem:
Fixes the regression reported after #3318 (IDEMPIERE-7070 was reopened).

JIRA: https://idempiere.atlassian.net/browse/IDEMPIERE-7070

## Problem

After merging #3318:

- Native HTML previews lost their drill-down links and viewer formatting. The native report engine fallback rendered HTML directly via `ReportEngine.createHTML(...)`, bypassing `HTMLReportViewerRenderer` and therefore `HTMLExtension`, the web context path and viewer-specific CSS/JavaScript.
- The HTML download link failed for localized report names containing non-ASCII characters (e.g. `Offene_Posten__Nur_Überweisung_.html`) with `HTTP ERROR 400 Ambiguous URI path encoding`, because `FileUtil.makePrefix()` preserves Unicode and the raw name was embedded in the dynamic media URI passed to the anchor href.

## Solution

- New internal content type `text/html; mode=interactive` (`ReportContentType.HTML_INTERACTIVE_CONTENT_TYPE`):
  - The native report engine fallback advertises it but returns no content for it.
  - The ZK viewer's `HTMLReportViewerRenderer` renders the interactive preview (with `HTMLExtension`, drill-down links, viewer CSS/JS) and passes it through the report content processor pipeline using the interactive content type.
  - Plain `text/html` export rendering stays separate, so preview and export use the configuration each of them requires.
- Interactive HTML is delivered to the browser as `text/html` only.
- `FileUtil.makeASCIIPrefix()`: media names delivered through viewer URIs are now ASCII-safe (root-cause fix for the Jetty 400; also protects the pdf.js viewer URL). Download file names for CSV/XLS/XLSX keep localized names.
- Report viewer link (`DynamicMediaLink`): HTML and PDF open in a new browser tab via href/target (previous behavior), other formats download the media; the label shows the localized report name.
- Row count and Summary checkbox are visible again for native reports (`IReportContentRenderer.getRowCount()`), hidden for renderers without row data support.
- Defensive fixes: null-safe `rendererMap` lookups and null guard in `DynamicMediaLink`.

## Tests

`ReportViewerContentRendererFactoryTest` additions:

- Native report engine fallback advertises interactive HTML and returns no content for it.
- Interactive HTML uses its own pipeline cache key; processors run separately for preview and export content types.

Plus full reactor build and manual testing of viewer preview/export/download paths with localized (de_DE) report names.

## Summary by CodeRabbit

- **New Features**
  - Added support for interactive HTML report previews when available.
  - Interactive HTML reports can open directly in the browser.
  - Improved report downloads with safer, more compatible file name prefixes.
  - Export options now distinguish interactive HTML from standard HTML output.

- **Bug Fixes**
  - Improved handling when report content or renderers are unavailable.
  - Prevented unnecessary download actions when opening media in the browser.

- **Style**
  - Limited the preview format selector width on smaller screens for improved usability.

### DIFF
--- a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/HTMLReportViewerRenderer.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/HTMLReportViewerRenderer.java
@@ -32,6 +32,8 @@ import org.compiere.tools.FileUtil;
 import org.compiere.util.CLogger;
 import org.compiere.util.Env;
 import org.compiere.util.Msg;
+import org.idempiere.print.ReportContentRequest;
+import org.idempiere.print.ReportContentType;
 import org.idempiere.print.renderer.HTMLReportRendererConfiguration;
 import org.idempiere.print.renderer.IReportRenderer;
 import org.idempiere.print.renderer.IReportRendererConfiguration;
@@ -99,6 +101,9 @@ public class HTMLReportViewerRenderer implements IReportViewerRenderer {
 					.setExtension(new HTMLExtension(contextPath, "rp", viewer.getUuid()))
 					.setContextPath(contextPath);
 			renderer.renderReport(reportEngine, config);
+			String pipelineContentType = export ? getContentType() : ReportContentType.HTML_INTERACTIVE_CONTENT_TYPE;
+			file = Core.processReportContent(new ReportContentRequest(reportEngine,
+					reportEngine.getProcessInfo(), viewer.getTitle()), pipelineContentType, getFileExtension(), file);
 			return new AMedia(file.getName(), getFileExtension(), getContentType(), file, false);
 		} catch (Exception e) {
 			if (e instanceof RuntimeException)
--- a/org.adempiere.base/src/org/adempiere/base/Core.java
+++ b/org.adempiere.base/src/org/adempiere/base/Core.java
@@ -1366,6 +1366,8 @@ public class Core {
 		private static final ReportContentType[] SUPPORTED_CONTENT_TYPES = {
 				new ReportContentType(Msg.getMsg(Env.getCtx(), "FilePDF"), "pdf", "application/pdf"),
 				new ReportContentType(Msg.getMsg(Env.getCtx(), "FileHTML"), "html", "text/html"),
+				new ReportContentType(Msg.getMsg(Env.getCtx(), "FileHTML"), "html",
+						ReportContentType.HTML_INTERACTIVE_CONTENT_TYPE),
 				new ReportContentType(Msg.getMsg(Env.getCtx(), "FileCSV"), "csv", "text/csv"),
 				new ReportContentType(Msg.getMsg(Env.getCtx(), "FileXLS"), "xls", "application/vnd.ms-excel"),
 				new ReportContentType(Msg.getMsg(Env.getCtx(), "FileXLSX"), "xlsx",
@@ -1379,6 +1381,8 @@ public class Core {
 
 		@Override
 		public File getContent(String contentType, String fileExtension) {
+			if (ReportContentType.isInteractiveHTML(contentType))
+				return null;
 			String extension = fileExtension != null ? fileExtension.toLowerCase() : "";
 			if (!List.of("pdf", "html", "csv", "xls", "xlsx").contains(extension))
 				return null;
--- a/org.adempiere.base/src/org/compiere/tools/FileUtil.java
+++ b/org.adempiere.base/src/org/compiere/tools/FileUtil.java
@@ -535,6 +535,29 @@ public class FileUtil
 	}
 	
 	/**
+	 * Creates a valid file name prefix from "name", replacing every
+	 * character outside the US-ASCII range with an underscore.
+	 * Accented characters are transliterated first via
+	 * {@link Util#setFilenameCorrect(String)}, so e.g. "Ü" becomes "U".
+	 * The result is safe to embed in URLs and URIs.
+	 * @param name
+	 * @return ASCII-safe file name prefix
+	 */
+    public static String makeASCIIPrefix(String name) {
+		String output = Util.setFilenameCorrect(name);
+		StringBuilder prefix = new StringBuilder();
+		char[] nameArray = output.toCharArray();
+		for (char ch : nameArray) {
+			if (ch < 128 && Character.isLetterOrDigit(ch)) {
+				prefix.append(ch);
+			} else {
+				prefix.append("_");
+			}
+		}
+		return prefix.toString();
+	}
+	
+	/**
 	 * 
 	 * @param path
 	 * @return true if deleted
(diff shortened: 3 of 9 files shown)


org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/HTMLReportViewerRenderer.java
```diff
@@ -87,7 +89,7 @@ public AMedia renderMedia(ZkReportViewer viewer, boolean export) {
 		ReportEngine reportEngine = viewer.getReportEngine();
 		try {
 			String path = System.getProperty("java.io.tmpdir");
-			String prefix = makePrefix(reportEngine.getName());
+			String prefix = FileUtil.makeASCIIPrefix(reportEngine.getName());
```
CarlosRuiz-globalqss: The makePrefix in the interface is still used in 5 other points, I think it would be better to change the interface too, like this:

```diff
diff --git a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/IReportViewerRenderer.java b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/IReportViewerRenderer.java
index 3e378d72bd..57b8507359 100644
--- a/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/IReportViewerRenderer.java
+++ b/org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/IReportViewerRenderer.java
@@ -23,6 +23,7 @@ package org.idempiere.ui.zk.report;
 
 import org.adempiere.webui.window.ZkReportViewer;
 import org.compiere.print.ReportEngine;
+import org.compiere.tools.FileUtil;
 import org.zkoss.util.media.AMedia;
 
 /**
@@ -109,15 +110,6 @@ public interface IReportViewerRenderer {
 	 * @return file name prefix from name
 	 */
 	default String makePrefix(String name) {
-		StringBuilder prefix = new StringBuilder();
-		char[] nameArray = name.toCharArray();
-		for (char ch : nameArray) {
-			if (Character.isLetterOrDigit(ch)) {
-				prefix.append(ch);
-			} else {
-				prefix.append("_");
-			}
-		}
-		return prefix.toString();
+		return FileUtil.makeASCIIPrefix(name);
 	}
 }
```

CarlosRuiz-globalqss (changes requested):
I just noticed there is also something else broken in mobile mode:

This is the screenshot from r13:

Image

and this is from r14:

Image

Because the drop down now is too wide, the arrow button at the end is not visible anymore, it must still be there because if I put the mouse and click it, it works.

Also, not caused by this, so maybe for another UX ticket, the dialog opened there cannot be closed without taking one of the actions.

org.adempiere.ui.zk/WEB-INF/src/org/idempiere/ui/zk/report/HTMLReportViewerRenderer.java
mbozem: Adopted - `IReportViewerRenderer.makePrefix()` now delegates to `FileUtil.makeASCIIPrefix(name)`, and the explicit calls in the HTML and PDF renderers were reverted to `makePrefix(...)` so there is a single point where the ASCII-safe prefix is produced. Thanks.

mbozem:
Thanks for the review and the detailed testing, Carlos.

**Mobile dropdown (arrow not visible):** Root cause is the unbounded width of the preview type select: native `<select>` elements do not shrink below their intrinsic (longest entry) width, so on small screens the select pushed the trailing arrow - and in the Jasper process viewer the following toolbar buttons - out of the visible area. Fixed by clamping the select with `max-width: 40%` when the viewport width is below `ClientInfo.SMALL_WIDTH` (the same threshold the toolbar popup already uses), applied to both `ZkReportViewer` (window print flow) and `ZkJRViewer` (process preview flow) so both behave consistently.

**Dialog cannot be closed without taking an action:** Agreed, this is pre-existing and unrelated to this change. Root cause: `WMediaOptions` is created with `ZKUpdateUtil.setWindowWidthX(this, 400)`, which only clamps the width to 100% when `ClientInfo.get().desktopWidth` is detected as smaller than 400px; otherwise the dialog keeps a fixed 400px width. On devices where the desktop width is not correctly reported (or unknown, desktopWidth == 0), the dialog ends up wider than the visible viewport, so the closable icon in the title bar is out of view. There is no Cancel button and highlighted windows do not close on ESC, so the dialog can only be dismissed by executing one of the actions.

Suggestion for a separate UX ticket: clamp the dialog with `max-width: calc(100% - 10px)` (one line in `WMediaOptions`; other `popup-dialog` windows could follow via a theme CSS rule), and optionally add an explicit Cancel button.

Both fixes are pushed to this PR - please re-review.

CarlosRuiz-globalqss (changes requested):
Thanks @mbozem 

I found another issue probably caused by these changes (because is not reproducible in r13):

- Run Open Items (Aging) report
- Choose CSV in the dropdown
- Click on _Preview_ button
- the system shows an error:
  The server is temporarily out of service.
  Would you like to try again?
  (414)
- the console shows this error:
  `-----------> HttpParser.parseLine: URI is too large >8192 [591]`

mbozem:
Thanks for the detailed reproduction, @CarlosRuiz-globalqss.

The root cause of the 414 can be traced. It is **not** caused by the changes in this PR - the CSV/XLS preview flow (Extensions.getMediaView -> WMediaOptions -> Keikai renderMediaView) is identical in r13 and r14. The difference is the server push transport:

- r13 default: **Atmosphere** server push (AU updates travel over the long-lived COMET body, no request-line size issue)
- r14 default: **WebSocket** server push, changed by the separately merged PRs for **IDEMPIERE-3840** ("restore back websocket as default for r14", #3239 / #3224) - not part of this PR

With the WebSocket server push, `ServerPushEndPoint.onMessage()` rebuilds the internal AU request by putting the whole AU content into the **request line** as a query string:

```java
// ServerPushEndPoint.java:204-207
StringBuilder fullUrl = new StringBuilder(this.baseUrl)
        .append(uri)
        .append("?")
        .append(content);   // <-- the entire AU payload goes into the URL
...
HttpPost httpPost = new HttpPost(fullUrl.toString());
```

For a large CSV preview (Open Items / Aging report -> many cells pushed through Keikai into the spreadsheet) the AU content is easily larger than 8192 bytes, so Jetty's `HttpParser` rejects the request line with `HttpParser.parseLine: URI is too large >8192` (HTTP 414) - exactly the error you saw.

I reproduced this before/after with the real Jetty 12.0.24 `HttpParser` from the target platform:

```
AU content size            = 14768 bytes
BEFORE request-line size   = 14804 bytes   <- buggy: content in URL
AFTER  request-line size   = 35 bytes      <- fix:  content as POST body

BEFORE: content in URL  -> REJECTED (Jetty: URI is too large >8192)
AFTER:  content in body -> parsed OK
```

I filed **IDEMPIERE-7095** and prepared a separate PR that sends the AU content as the POST body (`new StringEntity(content, APPLICATION_FORM_URLENCODED)`) instead of the query string. Workaround until then: start the server with `-Dorg.idempiere.ui.zk.serverpush=atmosphere`.

Please let me know if you would like the CSV-preview 414 handled differently, and re-review this PR (the makePrefix and mobile dropdown items from your previous review are addressed).

CarlosRuiz-globalqss (approved):

PR #3364, opened 2026-09-16
IDEMPIERE-6968 NPE in FavoriteSimpleTreeModel.render() when role lacks window access

shawon9324:
https://idempiere.atlassian.net/browse/IDEMPIERE-6968

## Problem

`MRole.getWindowAccess(int)` is documented as:

> `@return null if can't access, TRUE if r/w and FALSE if r/o`

`MToolBarButtonRestrict.isNewButtonRestricted(int, int)` assigned that `Boolean` to a primitive `boolean`, so when a role has no `AD_Window_Access` record for the window the `null` was auto-unboxed and threw:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()"
    because the return value of "org.compiere.model.MRole.getWindowAccess(int)" is null
    at org.compiere.model.MToolBarButtonRestrict.isNewButtonRestricted(MToolBarButtonRestrict.java:220)
    at org.adempiere.webui.component.FavoriteSimpleTreeModel.render(FavoriteSimpleTreeModel.java:210)
    ...
```

This was **the only call site in core that unboxed the result**. Every other caller (~20 of them, e.g. `GenericZoomProvider`, `MDocumentStatus`, `GridWindowVO`, `WEditorPopupMenu`, `FindWindow`, `MTree`, `MTreeFavorite`) holds it in a `Boolean` and null-checks it.

## Reachable from at least two paths

1. **`FavoriteSimpleTreeModel.render()`** — a favourite added under one role is rendered under another role that cannot access the window. The ZK desktop error handler fires and the user is left with a blank menu tree.

2. **`WEditorPopupMenu`** (not mentioned in the ticket) — at lines 186-187, the `else` branch is taken when the role has no access to the SO zoom window (`canAccess == null`) but *does* have access to the PO one. `isNewButtonRestricted(winID)` is then called with a window whose access is `null`, and throws.

## Fix

```java
//	getWindowAccess returns null when the role has no access to the window at all
Boolean isReadWrite = MRole.getDefault().getWindowAccess(windowID);
if (!Boolean.TRUE.equals(isReadWrite))
    return true;
```

"No access record" is treated as "no access", so the New button is reported as restricted rather than throwing. This matches how the rest of the codebase interprets a `null` from `getWindowAccess`. The read/write and read-only branches are unchanged.

## Notes on the ticket's other suggestions

Two points in the ticket did not hold up when checked against the source, so they are deliberately **not** acted on here:

- *"There are other call sites in `MToolBarButtonRestrict` that use the same pattern"* — there are none. Line 220 was the only occurrence in the class.
- *"`FavoriteSimpleTreeModel` could filter out menu nodes whose window is not accessible"* — `MTreeFavorite` already does this at load time via `getAccessForMenuItem()` (`MTreeFavorite.java:135`), skipping nodes whose access is `null`. Adding a second filter in the renderer would be redundant, so the change is kept to the null-safety fix.

## Testing

- Added `org.idempiere.test/src/org/idempiere/test/model/MToolBarButtonRestrictTest.java`, covering both the `null`-access branch (the regression) and the normal accessible-window branch. It uses window id `-9999` for the no-access case, the same idiom as the existing `MRoleTest.testGetWindowAccess()`.
- Full reactor build passes on JDK 17: `mvnw clean verify -DskipTests` → **BUILD SUCCESS**, 61/61 modules, 15:57.
- Behaviour of the old vs. new logic was verified in isolation: the old form throws the exact NPE above on `null`, the new form returns `true`, and the read/write (`false`) and read-only (`true`) results are identical between the two.

**Disclosure:** the new unit test compiles but has **not been executed** — running `org.idempiere.test` requires a PostgreSQL instance seeded with GardenWorld, which was not available in this environment. The corresponding checklist item below is left unticked for that reason, and I would appreciate a reviewer running it.

# Pull Request Checklist

### Tests
- [ ] New and existing unit tests pass locally with my changes — *not verified, no seeded database available*
### Documentation
- - [ ] New feature — n/a
- - [ ] Breaking change — n/a
- - [ ] Improvement — n/a, bug fix with no change to user interaction
- [ ] The changed/added documentation is in the project wiki — n/a

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected toolbar behavior when a role has no access to a window. New buttons are now correctly treated as restricted instead of causing an error.
  * Preserved normal behavior for windows the role can access.

* **Tests**
  * Added coverage for both accessible and inaccessible windows.


hengsin (approved):
looks good.

PR #3366, opened 2026-09-22
IDEMPIERE-7106 Fix broken Capture Process Parameter popup in Scheduler

shawon9324:
## Summary
Fixes https://idempiere.atlassian.net/browse/IDEMPIERE-7106

Opening Scheduler window → Gear button → Capture process parameter shows a
broken, unusable popup: only the title bar and OK/Cancel buttons render: the
actual process parameter fields collapse to zero height.

## Root cause
In `WProcessParameterForm`, the parameter panel container (`dialogContent`)
was set to `vflex="1"` (flex-grow), while the popup `Window` itself uses
`vflex="min"` (shrink-to-fit). A flex item asking to grow inside an ancestor
that is itself sizing to its own content is an unsatisfiable flex constraint,
so the browser collapses the child to zero height. The parameter fields never
get any visible space, leaving just the dialog chrome.

This matches the screenshot attached to the Jira ticket (pulled via Jira's
public REST API): the "Create/Update Table Partition" dialog opens with a
clipped sliver of the first parameter label and the OK/Cancel buttons
floating in otherwise blank space.

## Only window with this conflict
Checked every class in `org.adempiere.ui.zk` that shrinks its own top-level
`Window` to fit its content (`ZKUpdateUtil.setVflex(this, "min")`, the same
idiom used here). Only two hits: `WProcessParameterForm` (this fix) and
`ConfirmPanel` — which isn't a `Window` at all, just an embedded `Div`
button-bar reused inside many dialogs, and doesn't have an inner `vflex="1"`
content child fighting it. So this bug, and the fix, are correctly scoped to
the one popup that has it — no sibling dialog needs the same change.

## Fix
Remove the conflicting `vflex="1"` on `dialogContent` so it sizes to its
natural content height instead, consistent with how the other `HIGHLIGHTED`
`ADForm` popups in the codebase (`AddAuthorizationForm`, `MFARegisterForm`)
are laid out — none of them fight the window's shrink-to-fit sizing with a
flex-grow child.

## Testing
- Verified via code/CSS analysis cross-checked against the actual screenshot
  attached to the Jira ticket.
- Added `org.idempiere.test/src/org/idempiere/test/ui/WProcessParameterFormTest.java`.
  It constructs the real `WProcessParameterForm` (mocking `SessionManager`
  the same way `InfoGeneralPanelTest` already does), walks the actual
  component tree, and asserts `dialogContent`'s `vflex` is never `"1"` again —
  the exact configuration that collapsed the parameter panel to zero height.
  This can't verify rendered pixels in a real browser (nothing in this
  project's test suite does that), but it does lock in the fix at the
  component-tree level, so this specific regression can't silently come back.
- Full reactor build: `mvnw clean verify -DskipTests -Djacoco.skip=true` →
  **BUILD SUCCESS**, 61/61 modules, 1:43 min. `org.adempiere.ui.zk` (the fix)
  and `org.idempiere.test` (the new test) both compile cleanly.
- **Disclosure:** I was not able to visually confirm the rendered fix in a
  live browser session — this environment's Docker setup was too slow to spin
  up a test iDempiere instance in reasonable time, and the new test above was
  not executed against a live seeded database either (same limitation noted
  on IDEMPIERE-7094). Would appreciate a manual check of Scheduler → Gear →
  Capture process parameter on a process with parameters before merge.

# Pull Request Checklist

### Tests
- [x] I have tested the direct scenario that my code is solving — via code/CSS analysis cross-checked against the reported screenshot, not a live browser (see Testing/Disclosure above)
- [x] I checked all collaterals that can be affected by my changes, and tested other potential affected scenarios — audited every other `Window` using the same shrink-to-fit idiom for the same conflict
- [ ] New and existing unit tests pass locally with my changes — *not verified, no seeded database available*
### Documentation
- - [ ] New feature — n/a
- - [ ] Breaking change — n/a
- - [ ] Improvement — n/a, bug fix with no change to user interaction
- [ ] The changed/added documentation is in the project wiki — n/a

## Summary by CodeRabbit

* **Style**
  * Standardized formatting in the process parameter form.
  * Adjusted dialog content layout behavior by removing automatic vertical expansion.

* **Tests**
  * Added coverage to verify the dialog content does not use vertical flex expansion.


CarlosRuiz-globalqss (approved):
tested successfully

CarlosRuiz-globalqss:
Thanks @shawon9324 - happy to have a new committer!

